PR #15, opened 2022-10-30
Get from JSON array

TheShovel:
A reporter block that lets you get a item from a JSON array.


TheShovel:
Sorry for adding TextBox.js by mistake. Removed it now.

GarboMuffin:
I'm not sure if this is needed because you can already do

image

TheShovel:
It wold be more convenient and easier to understand, but I get your point, this is not really needed.

GarboMuffin:
Closing for now per previous comment that this can already be done using other blocks

PR #16, opened 2022-10-31
Add extension: Cloudlink 4

MikeDev101:
https://github.com/mikedev101/cloudlink

### DIFF
--- a/cloudlink.js
+++ b/cloudlink.js
@@ -1,5 +1,7 @@
 // Copy of S4-0_nosuite.js as of 10/31/2022
 
+(function(Scratch) {
+
 var servers = {}; ; // Server list
 let mWS = null;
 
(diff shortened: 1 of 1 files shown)


GarboMuffin:
I made some manual changes to how the extension registers itself until I decide on an API that I am more satisfied with (draft documentation: https://docs.turbowarp.org/development/unsandboxed-extensions). Otherwise it's unchanged.

https://extensions.turbowarp.org/cloudlink.js

GarboMuffin:
@MikeDev101 I would like for all of the extensions here to be under a consistent open source license. The Cloudlink extension appears to this license: https://github.com/MikeDev101/cloudlink/blob/master/LICENSE

I'm curious:

 - What is the motivation to writing your own license?
 - What is the motivation for the explicit lack of patent grant?

We would like to put everything in this repository under the MIT License (implicit patent grant) or Apache 2.0 (explicit patent grant). It appears that Cloudlink's license will cause conflicts with either of these choices. Is it possible to put Cloudlink under a license that doesn't have an explicit lack of patent grant?

(Disclaimer: I am not a lawyer)

GarboMuffin:
The open source initiative [was unable to come to consensus on whether the CC0 license meets the open source definition](https://opensource.org/faq#cc-zero) because it has a similar explicit lack of patent grant. I would like to avoid a situation where I have to remove Cloudlink because it uses a license that is not truly "open source".

MikeDev101:
If I recall correctly, the MEOW license was created as a derivative of 0BSD, but modified to deny any patent usage, as well as allowing users to change the license if they deem it necessary. I tried to get it reviewed by the OSI, but I forgot about it.

I originally made CL public domain. After reconsideration, I wanted to use a license that still gave me credit, but remains extremely permissive (I once joked about using the WTFPL), so MEOW was created.

In reality, I really don't care what people do with the software, as long as it isn't illegal, infringes upon other laws, or attempts to use it in a maliciously proprietary form.

I do see weaknesses in this custom license, so I'll yeild. MIT seems fine by me.

GarboMuffin:
Thanks

PR #19, opened 2022-11-06
New extension: File read and download

GarboMuffin:
image

which opens a modal that looks like this (design temporary, known to be ugly) because we can't reliably open an input dialog without user interaction

image

### DIFF
--- a/files.js
+++ b/files.js
@@ -1,0 +1,204 @@
+(function(Scratch) {
+  'use strict';
+
+  const vm = Scratch.vm;
+
+  const showFilePrompt = (accept) => new Promise((_resolve) => {
+    // We can't reliably show an <input> picker without "user interaction" in all environments,
+    // so we have to show our own UI anyways. We may as well use this to implement some nice features
+    // that native file pickers don't have:
+    //  - Easy drag+drop
+    //  - Reliable cancel button (input cancel event is still basically nonexistent)
+    //    This is important so we can make this just a reporter instead of a command+hat block.
+    //    Without an interface, the script would be stalled if the prompt was just cancelled.
+
+    /** @param {string} text */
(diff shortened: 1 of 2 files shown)


FilipeCollura:
Is the modal customizable? And if it cannot, can it be hidden? Thanks! :)

Mickmick2012:
> Is the modal customizable? And if it cannot, can it be hidden? Thanks! :)

not customizable but can be hidden with the block "set open file selector to (hide modal)"and you can do show it with the same block

FilipeCollura:
> 

That is not an available option. I can only choose to either show the modal or open the selector immediately.

GarboMuffin:
I'll restate what I said in https://github.com/TurboWarp/extensions/issues/28 (yes that's a bit buried and hard to find)

I'm hesitant to add a way to remove the prompt entirely because:

 - Browsers don't always let us open a file dialog, and they don't tell us when they refuse to, so the script would get stuck forever if there wasn't a way for people to manually open the dialog
 - Browsers don't reliably tell us when someone presses cancel on the dialog (only Firefox implements `oncancel`), so people need way to cancel the file dialog manually, otherwise the script gets stuck forever
 - Drag+drop is a very common way for people to select files. Operating system files prompts do not accommodate this.

Regardless, I did end up adding this as a hidden mode:

image

(the second text input there is completely empty, not even a space)

for the reasons mentioned previously, I don't recommend using this mode, so it hasn't been added to the dropdown at this time.

GarboMuffin:
If you want additional customization, please describe that more specifically

FilipeCollura:
Thanks for this info. I wanted to remove it completely because it really didn't fit with my projects theme. If there's a way to customize the popup so that it can fit my theme, it would be great! :)

PR #25, opened 2022-11-07
Create rixxyx.js

Alestore:
An extension that allows you to do a lot of interesting things, including transforming strings into binary code and then converting them back into binary strings (I use it to pass string values through the TurboWarp cloud server which only supports integer type cloud variables)

### DIFF
--- a/rixxyx.js
+++ b/rixxyx.js
@@ -1,0 +1,507 @@
+var count = 0
+var isMeasure = false
+var time = 0
+
+class RixxyX {
+  getInfo() {
+    return {
+      color1: "#773c00",
+      color2: "#5f3000",
+      id: "RixxyX",
+      name: "RixxyX",
+      description: '[deleted]',
+      blocks: [
+        {
+          opcode: 'notEquals',
+          blockType: Scratch.BlockType.BOOLEAN,
+          text: '[TEXT_1] != [TEXT_2]',
+          arguments: {
+            TEXT_1: {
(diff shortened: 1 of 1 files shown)


GarboMuffin:
see the part of https://docs.turbowarp.org/development/unsandboxed-extensions about wrapping the extension code in an IIFE to avoid leaking to global scope

GarboMuffin:
I am a bit curious about what 'rixxyx' means

Alestore:
> I am a bit curious about what 'rixxyx' means

I can't really tell you, I found it in a Scratch forum.

GarboMuffin:
So is this something you made or copied from somewhere else? If it's from somewhere else, link the original source, we need to know the license information and who needs to be attributed.

Alestore:
> So is this something you made or copied from somewhere else? If it's from somewhere else, link the original source, we need to know the license information and who needs to be attributed.

Yes, I got it from a Scratch forum, I'm sorry.

You can find it in this list under the letter R: https://scratch.mit.edu/discuss/m/topic/470021/

GarboMuffin:
I've asked the original creator for permission, so we'll see how that goes

Also, arbitrary JS execution blocks will have to be removed or altered: returnObject, execJsCode

GarboMuffin:
Compared to the original version of the extension, returnObject, returnJsCode, and execJsCode were removed.

Note that the color reporter block currently returns the wrong values until a compiler bug fix goes live soon-ish

PR #26, opened 2022-11-07
GameJolt API

softedco:
Might not be suitable for this but it's stable and it works


GarboMuffin:
move it to just `gamejolt.js`

GarboMuffin:
This is something quite a few people have asked for, so I'll be reviewing it more than some other extensions so that we can get this right the first time

softedco:
> This is something quite a few people have asked for, so I'll be reviewing it more than some other extensions so that we can get this right the first time

Wow, didn't know this was actually talked about

GarboMuffin:
It looks like this includes a minified version of a library. Where can I find the library's source and what modifications have you made?

softedco:
The library the extension is using is the [gamejolt api js library](https://github.com/MausGames/game-jolt-api-js-library)
by Martin Mauersics or @MausGames, it is under the unlicense license, as well as the extension, which, as far as I know, should not conflict with the MIT license. The modifications I made is migrating to the 1.2 version of the GameJolt API and also some methods related with the new namespaces presented in that version, as well as some unnecessary methods I made to optimize the code.

softedco:
I have a [repo](https://github.com/softedco/GameJolt-API-Scratch-extension) that contains all necessary information in terms of licensing and credits, as well as a link for quickly testing the extension on tw, if that could be of some help.

GarboMuffin:
Sorry for radio silence on this. It's fine to have your own fork of that GameJolt API library, but can you make your modifications to the unminified version of the library?

softedco:
Well, I definetely could, but I don't think that it would be very useful since the modifications made to the unminified version are not very significant. The custom requests are all unminified, and the changes I made are indirectly stated in the comments, except maybe success properties of the response object which are strings and not booleans for some reason, they are stated later when the library part ends. But basically, the modifications I made to the minified version are changing the api's url, adding GJAPI.sAPIO which stores the old api url for GJAPI.DataStoreFetch to work properly and changing the GJAPI.SendRequestEx method to work with the old url when DataStoreFetch is called. I also changed some `b.success` checks to `b.success == 'true'` for the reason said earlier. That's really all I changed, let me know if you would be ok with that or not, because it wouldn't take me much time to get all of the modifications into the unminified version, I already have a pr being discussed in the library repo.

GarboMuffin:
I will try to get this in by the end of the weekend. The big "Clean up" commit mostly just replaces minified code with unminified code for maintainability.

Some questions:

 - How do I test this?
 - What is the motivation for the version block and up-to-date checker? Is the GameJolt API known to change in a way that the extension breaks?
 - Do any existing projects use the previous version of this extension?

softedco:
To test the extension you will need a GameJolt account and a game page, since it requires game credentials for the extension to work properly as well as user credentials for some user related blocks, again, might not really be suitable for tw.
The version block is only there to check if the extension is outdated or not, it is just there for convenience when using packaged projects with the code, not necessary since there haven't been any cases of the extension breaking.
And yes, there are existing projects using older versions, but they are packaged with the code and not the link so even if something goes wrong they will not break.

GarboMuffin:
I'll merge this tonight so consider giving it another look through before then

GarboMuffin:
Welcome https://extensions.turbowarp.org/gamejolt.js

I haven't really tested this, assuming it works based on prior versions having users. If anything is broken, just send another PR

PR #38, opened 2022-11-15
Add Augmented Reality extension

Xeltalliv:
New blocks in block palette

An extension that allows scratch projects to use AR capabilities provided by [WebXR](https://developer.mozilla.org/en-US/docs/Web/API/WebXR_Device_API) api.
[Example project](https://github.com/Xeltalliv/extensions/tree/augmented-reality-extras/ar_extras)

https://user-images.githubusercontent.com/90340988/201976379-f030f85a-81bb-4427-a85e-2679ed509aaa.mp4

State of this extension: it works and can be used; it's buggy; some part are implemented unoptimally; I don't have sufficient knowledge about scratch internals to make it better.
I don't expect it to be accepted in it's current state, but you could probably tell me how I could improve it or do it directly yourself.


GarboMuffin:
This is the kind of extension I was hoping to see.

It looks like reviewing this extension will be quite a task. I don't have time to review the entire extension yet, but is there anything in particular you want me to look at?

Xeltalliv:
AR session is always in fullscreen and involves rendering everything to a framebuffer provided by the WebXR api, with additional ability to add DOM overlay. That means that to already existing normal player, mini player and fullscreen I had to essentially add another type of fullscreen (which runs at screen's aspect ratio). I don't know a proper way to do that, but with combination of vm.runtime.setStageSize, putting "stage-wrapper_stage-canvas-wrapper" with "transform: scale" in DOM overlay(including original canvas with "opacity: 0" to make touch detection work), and also changing _pickTarget to take into a count "transform: scale" I manged to make something that kind-of works. Currently that is the main source of all of the bugs and instabilities.

GarboMuffin:
This seems like the type of extension where the internals will always be a bit of a mess without extremely specific Scratch.* APIs because you need to mess with some fundamental parts of how rendering works.

Anonymous-cat1:
I know this is a weird idea but, what is the probability that this can lead to a Web-XR Turbowarp-VR extension? it would be awesome to use scratch to make VR content even if not in 3D.

Xeltalliv:
The thing is, I feel like it is more messy than it should be and that I'm modifying wrong parts of the code. For easier review, I've separated the problematic fullscreen code into [a separate extension](https://github.com/Xeltalliv/extensions/blob/augmented-reality-extras/ar_extras/fullscreen.js) which doesn't use WebXR. Can you take a look and think of whether there are better ways of implementing what it does?

Depending on whether the project is in normal player, mini player, or fullscreen mode, turbowarp recalculates and changes width and height of elements in pixels. I wasn't able to find a where that code is to modify it, so the way I'm doing it is using transform:scale on one of the outer elements. As a result, the resolution of canvas is different depending of whether it was entered from normal, mini or fullscreen mode.

Also, a question not related to fullscreen, is it possible to do anything about those?
https://github.com/TurboWarp/extensions/blob/f54dbde7b62b69183b79c9f38ae359b07732d5f3/ar.js#L320-L321
https://github.com/TurboWarp/extensions/blob/f54dbde7b62b69183b79c9f38ae359b07732d5f3/ar.js#L130-L150

Xeltalliv:
Should this extension be listed on homepage? [(Should I merge Xeltalliv:augmented-reality-featured -> Xeltalliv:augmented-reality ?)](https://github.com/Xeltalliv/extensions/commit/1fe054ab467b06fabd4515e2d492920ccbc86bdb)

GarboMuffin:
We can put this on the homepage

extensions/ar.js
```diff
@@ -0,0 +1,728 @@
...
+		}
+		turnSpaceBy(args) {
+			if(!xrRefSpace) return;
+			const r = +args.R || 0;
+			const i = +args.I || 0;
+			const j = +args.J || 0;
+			const k = +args.K || 0;
+			const len = Math.sqrt(r*r + i*i + j*j + k*k);
+			if(!isFinite(len)) return;
+			if(len === 0) len = r = 1;
```
GarboMuffin, line 689: `len` and `r` are both `const` so this won't work

GarboMuffin:
If making this work in the packager is too difficult, at least add something like:

```js
if (typeof scaffolding !== 'undefined') {
  alert("ar does not work in packager :(");
}
```

Xeltalliv:
> * Set a `docsURI` pointing somewhere to help people understand how to use this. You can drop an HTML file in `website` and use something like https://extensions.turbowarp.org/ar.html

I definitely need to add documentation
> * Give the extension a `blockIconURI`

It doesn't look that well (increased block width)
> * See if you can make it work in the packager

done. Also done the same changes to [fullscreen.js](https://github.com/Xeltalliv/extensions/commit/0da6f8ea4811a865532c27c8ac329d1837cb9739).



> If making this work in the packager is too difficult, at least add something like:
> 
> ```js
> if (typeof scaffolding !== 'undefined') {
>   alert("ar does not work in packager :(");
> }
> ```

it would've been useful if it was mentioned in [turbowarp docs](https://docs.turbowarp.org/development/unsandboxed-extensions)

Xeltalliv:
I think it's ready

GarboMuffin:
It may be a while before this appears on the website because GitHub Actions is currently not working properly https://www.githubstatus.com/incidents/n18j5g7wpg0s. I will try it again tonight.

PR #39, opened 2022-11-19
Mouse Cursor Extension

Samq64:
Source:  https://github.com/Samq64/scratch-extensions

It's just a test extension I made a while back and thought I would add it here in case someone finds it useful.

Tested locally, everything works except the cursor reporter is blank until running one of the other blocks.

### DIFF
--- a/cursor.js
+++ b/cursor.js
@@ -54,7 +54,7 @@
         }
 
         GetCur() {
-            return this.stage.style.cursor;
+            return this.stage.style.cursor || 'default';
         }
     }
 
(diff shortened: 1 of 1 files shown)


GarboMuffin:
Are you aware of any projects made using the original version of the extension that would justify backwards compatibility?

Samq64:
No. Someone did open an issue but have no way of knowing if it was used in a real project because the GitHub account was deleted.

Samq64:
There were accidentally some stashed changes in that last commit, but I think it's fine. Also, thanks for the fixes!

PR #50, opened 2022-11-29
Itch.io API

softedco:
A small Itch.io API reference extension
image
https://itch.io/docs/api/javascript
I tried tweaking the `Itch.attachBuyButton()` method to work as a command block but I couldn't get it to open the window so we have what we have


itchio.js
```diff
@@ -0,0 +1,199 @@
+(Scratch => {
+	'use strict';
+
+	/*!
+	 * Itch.io JavaScript API reference from https://static.itch.io/api.js
+	 * Softed by softed
+	 */
+	(() => {
+		window.ItchIo = {};
```
GarboMuffin, line 9: shouldn't be on window

GarboMuffin:
A little thing I've been wondering with both this and the gamejolt one: Both of them have some "secret" token that they use to fetch game data. Is that token actually supposed to be secret? Because in Scratch projects it very much is not secret

softedco:
The secret argument in this one is just for testing, it says in the documentation that it is not needed, you can basically fetch any game with the secret field empty because there shouldn't be any private information. But for the gamejolt extension the key you pass is private and the only way you can safely post the project somewhere is by posting a web only build of your project on your GameJolt game page. When accessing a private game key it says that it is private and I will additionally adress this as well as how to post the project in the extension's documentation which is currently work in progress.

GarboMuffin:
If secret is unused then let's remove it

softedco:
The secret argument is used in fetching draft games so it is still useful for development

PR #53, opened 2022-12-01
Add link to desktop app v1.7.0-beta.0

GarboMuffin:


softedco:
Is there an option to disable the automatic extension loading, it would be really nice for testing, for example testing how would the new version of an extension behave

PR #55, opened 2022-12-02
Add local storage extension

infernostars:
Basically just a wrapper for it but it's simple and works well enough


localStorageExtension.js
```diff
@@ -0,0 +1,54 @@
+(function(Scratch) {
+    'use strict';
+    if (!Scratch.extensions.unsandboxed) {
+        throw new Error('Box2D must be run unsandboxed');
```
softedco, line 4: why box2d tho

localStorageExtension.js
```diff
@@ -0,0 +1,54 @@
...
+    }
+    class LocalStorageExt {
+      getInfo () {
+        return {
+            id: "localstorageextension",
+            name: "Local Storage",
+            blocks: [
+                {
+                    opcode: 'lse_load_reporter',
```
GarboMuffin, line 13: you don't need to prefix them on your own; the VM automatically inserts the extension's ID first. So you can just call this `load` and the other one `set`

localStorageExtension.js
GarboMuffin, line 4:
```suggestion
        throw new Error('Local Storage must be run unsandboxed');
```

infernostars: accidental

softedco:
Might I suggest adding a remove key block

image

image

That way clearing site's data or interacting with the console is not needed to remove the keys

local-storage.js
```diff
@@ -0,0 +1,69 @@
...
+              key: {
+                type: Scratch.ArgumentType.STRING,
+                defaultValue: 'example',
+              },
+              value: {
+                type: Scratch.ArgumentType.STRING,
+                defaultValue: '1234',
+              },
+            },
+          },
```
softedco, line 40:
```suggestion
          },
          {
            opcode: 'remove',
            blockType: Scratch.BlockType.COMMAND,
            text: 'remove key [key]',
            arguments: {
              key: {
                type: Scratch.ArgumentType.STRING,
                defaultValue: 'example'
              },
            },
          },
```

local-storage.js
```diff
@@ -0,0 +1,69 @@
...
+          }
+          // Return the raw value as a string.
+          return storedValue;
+        }
+      } catch (e) {
+        // localStorage.getItem failed, ignore
+      }
+      return '';
+    }
+  }
```
softedco, line 67:
```suggestion
  }
  remove(args) {
    localStorage.removeItem(PREFIX + args.key.toString());
  }
```

infernostars:
> Might I suggest adding a remove key block
> 
> image
> 
> image
> 
> That way clearing site's data or interacting with the console is not needed to remove the keys

To be honest I completely forgot. Thanks!

local-storage.js
softedco: completely missed that it's not in a class nest, good thing it's fixed

GarboMuffin:
My current concern with this extension is that the storage keys are global instead of per-project, so if one project uses something generic like "data" then it will probably break some other project

I'm not sure the best way to resolve that

infernostars:
> My current concern with this extension is that the storage keys are global instead of per-project, so if one project uses something generic like "data" then it will probably break some other project
> 
> I'm not sure the best way to resolve that

hm, is there a way to get the project id or smth?

infernostars:
then we could do something like `prefix:1234567890:value`

softedco:
> > My current concern with this extension is that the storage keys are global instead of per-project, so if one project uses something generic like "data" then it will probably break some other project
> > I'm not sure the best way to resolve that
> 
> hm, is there a way to get the project id or smth?

You can't post projects with custom extensions on scratch right

CST1229:
> > > My current concern with this extension is that the storage keys are global instead of per-project, so if one project uses something generic like "data" then it will probably break some other project
> > > I'm not sure the best way to resolve that
> > 
> > 
> > hm, is there a way to get the project id or smth?
> 
> You can't post projects with custom extensions on scratch right

No, you can't,

softedco:
> > > > My current concern with this extension is that the storage keys are global instead of per-project, so if one project uses something generic like "data" then it will probably break some other project
> > > > I'm not sure the best way to resolve that
> > > 
> > > 
> > > hm, is there a way to get the project id or smth?
> > 
> > 
> > You can't post projects with custom extensions on scratch right
> 
> No, you can't,

Yeah so It won't work, and there probably aren't any other non-exploitable ways.

infernostars:
maybe based on project name? this would cause collisions with same project names but that is pretty rare (especially with how many people use extensions)

infernostars:
actually if we're loading, just use the url. this would be nonexploitable [unless you have 2 .sb3s in the exact same spot] and probably work fine. in editor we just read project name. if no name somehow then just some default. i think that would work

softedco:
> actually if we're loading, just use the url. this would be nonexploitable [unless you have 2 .sb3s in the exact same spot] and probably work fine. in editor we just read project name. if no name somehow then just some default. i think that would work

I don't know any ways to access the name of the project in the editor via extensions, it might even just be impossible. The URL sounds fine, but you will lose your data when the url changes, and you would have to  restore or remove it manually. To be truly unexploitable and safe, the project should have a unique and an independent parameter, that is also accesible to the extension, basically project ID but turbowarp doesn't have that.

softedco (commented):
requested by mistake?

infernostars:
> > actually if we're loading, just use the url. this would be nonexploitable [unless you have 2 .sb3s in the exact same spot] and probably work fine. in editor we just read project name. if no name somehow then just some default. i think that would work
> 
> I don't know any ways to access the name of the project in the editor via extensions, it might even just be impossible. The URL sounds fine, but you will lose your data when the url changes, and you would have to restore or remove it manually. To be truly unexploitable and safe, the project should have a unique and an independent parameter, that is also accesible to the extension, basically project ID but turbowarp doesn't have that.

urls changing is pretty unlikely, especially for a static file

StioStudio:
You can add a block (local ID[ID]) and that is the ID for the project. Really easy to use and make.

nikeedev:
> My current concern with this extension is that the storage keys are global instead of per-project, so if one project uses something generic like "data" then it will probably break some other project
> 
> I'm not sure the best way to resolve that

Those keys arent saved on the whole [turbowarp.org](https://turbowarp.org/), they are too, saved on slash domains like: [turbowarp.org/834792364](https://turbowarp.org/603265289), so it well save those keys on specific domains, it won't cause any issues.

nikeedev:
> > actually if we're loading, just use the url. this would be nonexploitable [unless you have 2 .sb3s in the exact same spot] and probably work fine. in editor we just read project name. if no name somehow then just some default. i think that would work
> 
> I don't know any ways to access the name of the project in the editor via extensions, it might even just be impossible. The URL sounds fine, but you will lose your data when the url changes, and you would have to restore or remove it manually. To be truly unexploitable and safe, the project should have a unique and an independent parameter, that is also accesible to the extension, basically project ID but turbowarp doesn't have that.

I'm agreeing with you

softedco:
> You can add a block (local ID[ID]) and that is the ID for the project. Really easy to use and make.

This might still introduce collision

softedco:
> > My current concern with this extension is that the storage keys are global instead of per-project, so if one project uses something generic like "data" then it will probably break some other project
> > 
> > I'm not sure the best way to resolve that
> 
> Those keys arent saved on the whole [turbowarp.org](https://turbowarp.org/), they are too, saved on slash domains like: [turbowarp.org/834792364](https://turbowarp.org/603265289), so it well save those keys on specific domains, it won't cause any issues.

Unfortunately, you can't upload projects with custom extensions on scratch, so slash domains are not an option

nikeedev:
> > > My current concern with this extension is that the storage keys are global instead of per-project, so if one project uses something generic like "data" then it will probably break some other project
> > > 
> > > I'm not sure the best way to resolve that
> > 
> > Those keys arent saved on the whole [turbowarp.org](https://turbowarp.org/), they are too, saved on slash domains like: [turbowarp.org/834792364](https://turbowarp.org/603265289), so it well save those keys on specific domains, it won't cause any issues.
> 
> Unfortunately, you can't upload projects with custom extensions on scratch, so slash domains are not an option

I never mentioned scratch in the comment, I even used turbowarp links.

nikeedev:
Those slash domains have own localStorage. They won't collide with other projects. 

turbowarp.org/1 has own, and turbowarp.org/2 has completely own localStorage too.

GarboMuffin:
I think the path forward is to add some API to the VM that lets the extension figure out the project ID. In the packager any projects with the same ID on the same domain will share local cloud variables; I think we want to copy that here.

softedco:
> Those slash domains have own localStorage. They won't collide with other projects. 
> 
> turbowarp.org/1 has own, and turbowarp.org/2 has completely own localStorage too.

You can't store a project on a turbowarp's slash domain if it is not on scratch under the same ID

softedco:
> > > > My current concern with this extension is that the storage keys are global instead of per-project, so if one project uses something generic like "data" then it will probably break some other project
> > > > 
> > > > I'm not sure the best way to resolve that
> > > 
> > > Those keys arent saved on the whole [turbowarp.org](https://turbowarp.org/), they are too, saved on slash domains like: [turbowarp.org/834792364](https://turbowarp.org/603265289), so it well save those keys on specific domains, it won't cause any issues.
> > 
> > Unfortunately, you can't upload projects with custom extensions on scratch, so slash domains are not an option
> 
> I had never mentioned scratch in the comment, I even used turbowarp links.

Turbowarp links are connected with scratch project IDs

nikeedev:
> Turbowarp links are connected with scratch project IDs

Aaaa, okay.

nikeedev:
> Turbowarp links are connected with scratch project IDs

Well that tells that we cant have localStorage then. You can probably close the case/issue.

softedco:
> > Turbowarp links are connected with scratch project IDs
> 
> Well that tells that we cant have localStorage then. You can probably close the case/issue.

Or we can

> I think the path forward is to add some API to the VM that lets the extension figure out the project ID. In the packager any projects with the same ID on the same domain will share local cloud variables; I think we want to copy that here.

nikeedev:
It could be the solution...

SIPC:
> I think the path forward is to add some API to the VM that lets the extension figure out the project ID. In the packager any projects with the same ID on the same domain will share local cloud variables; I think we want to copy that here.

That's a good one.

DT-is-not-available:
is it possible to store a value in a project via an extension without creating a scratch variable? if so, maybe you could generate a UUID for the project once the extension is loaded, and store keys based on that.

nikeedev:
@GarboMuffin, let’s close this case, doesn’t look like it’s active anymore. I don't think there are any solutions for adding stable local storage to turbowarp as an extension.

GarboMuffin:
Let's leave this open

DT-is-not-available:
> I don't think there are any solutions for adding stable local storage to turbowarp as an extension.

if the extension generates a UUID for the project file (or better yet uses the timestamp the extension was installed at to guarantee no conflict) and save keys via `localStorage.setItem(UUID+key, value)` it would work

infernostars:
Nice! took a while but we got it finally, major W [been busy so I haven't taken a look in a while]

PR #57, opened 2022-12-04
Add Notifications extension

mdwalters:


softedco:
This will only work when the project is running?

mdwalters:
uh

GarboMuffin:
My current concern with this extension is that notifications will appear as "from turbowarp.org", which puts the blame on us if someone makes a project that uses this maliciously.

tutacat:
@GarboMuffin Same as scratch.mit.edu wants permission to access your camera

@softedco It can't display notifications when not running, because a project doesn't have a 

@mdwalters Maybe just use a sound and an alert(); even that is quite a lot; and alert will normally show a notification of some sort when the tab is out of focus.

Cbr2011:
I like this to be an turbowarp extension

PR #59, opened 2022-12-05
Itch IO API not working!!!!!!!

imya505222:
itch io api not working i trying to add this but it dont showing


softedco:
I made a PR that fixes it

GarboMuffin:
Should work now

PR #63, opened 2022-12-08
GameJolt API - Many updates

softedco:
### stuff:
- added documentation, but it's still lacking examples and error descriptions, don't expect much
- added indents between block sections
- added 3 new blocks:
  - automatic login
  allows for automatic login after logout, which wasn't previously possible
  - autologin boolean
  allows to check if automatic login is available, different from GameJolt boolean since GameJolt allows guests to play games
  - logged in user's username reporter
    allows for the following:

    newblock

    which was previosly possible only after manual login with input username saved in a variable
- changed data storage get keys block, it now supports patterns, tested it, it doesn't break projects,
  but in old projects it would look something like this:
  
  uhoh

  is there a way to fix this?
- changed errors, they are now more accurate
- fixed data storage blocks not working with user data storage
- the old API link is not needed anymore
- some small optimizations
- other small stuff

### DIFF
--- a/gamejolt.js
+++ b/gamejolt.js
@@ -229,8 +229,7 @@
     GJAPI.bAutoLogin = true;
 
     GJAPI.sAPI = "https://api.gamejolt.com/api/game/v1_2";
-    GJAPI.sAPIold = "https://gamejolt.com/api/game/v1"; // This is for fixing DataStoreFetch
-    GJAPI.sLogName  = "[Game Jolt API]";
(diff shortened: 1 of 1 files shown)


GarboMuffin:
> changed data storage get keys block, it now supports patterns, tested it, it doesn't break projects,

would it make sense to have a separate block with a pattern input and another one without?

softedco:
The updated block has a placeholder pattern as a default value, if it is untouched, it will work the same as before.

PR #64, opened 2022-12-09
Stretch (scale x and y) extension

TheStarWorld:
probably not the best way of making it work

### DIFF
--- a/stretch.js
+++ b/stretch.js
@@ -1,9 +1,8 @@
-(function(Scratch) {
+(function (Scratch) {
   'use strict';
-  const vm = Scratch.vm;
 
-  class morescale {
-    getInfo () {
+  class Stretch {
+    getInfo() {
       return {
         id: 'moreScale',
         name: 'Stretch',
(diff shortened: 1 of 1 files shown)


dinoscove:
god i love you

GarboMuffin:
I think the big question here is what "unit" to use for how stretched a sprite is. Should "100" mean "100% of original size" (normal length) or "stretched by 100%" (double length)?

CST1229:
> I think the big question here is what "unit" to use for how stretched a sprite is. Should "100" mean "100% of original size" (normal length) or "stretched by 100%" (double length)?

Probably "100% = original size"; that's what Scratch 1.4 used for its (unused) stretch block.

GarboMuffin:
100

image

200

image

0

image

-100

image

-200

image

works the same in the Y direction too

GarboMuffin:
The Scratch 1.4 stretch block only has one number that is horizontal stretch. I'm not sure if that's something we want to copy with this.

Also, Scratch 3 technically has stretch blocks, they just don't do anything. Nothing would stop us from making them do something, though

CST1229:
> Also, Scratch 3 technically has stretch blocks, they just don't do anything. Nothing would stop us from making them do something, though

I am aware.

> The Scratch 1.4 stretch block only has one number that is horizontal stretch. I'm not sure if that's something we want to copy with this.

Of course we probably wouldn't remove vertical stretching.

PR #65, opened 2022-12-10
Add extension: Text

CST1229:
A bunch of blocks for handling and manipulating text.
image
> ```js
> // Modified port of an extension from a mod I worked on.
> ```

### DIFF
--- a/text.js
+++ b/text.js
@@ -474,7 +474,7 @@
 					));
 				break;
 				case CaseParam.TITLECASE:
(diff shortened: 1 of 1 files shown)


TrueFantom:
I started doing similar extension in parallel. The fact that there is analogous extension will prevent its posting?

GarboMuffin:
Consider adding those blocks to this one, at least the ones that don't already exist

PR #67, opened 2022-12-16
Add network extension

TrueFantom:
Several blocks for working with Internet:
Screenshot_164

I didn't add "fetch DELETE" and "fetch PATCH" blocks because I don't understand how these methods work. If someone needs it, I will add them.

*Thanks to  @TheShovel  for code of ping block.

*Method String() for arguments is need.


showierdata9978:
post should have a content-type option that is default text, for stuff like POST`in JSON

showierdata9978 (changes requested):
post should have a content-type option that is default text, for stuff like POST`in JSON

TrueFantom:
@showierdata9978

As I understand it, you are proposing to add menu with options "string" and "json". If "string" is selected, the code that exists now is executed. If "json" is selected, BODY is converted to JSON by method: "BODY = JSON.stringify(BODY);" and then same code. Did I understand you correctly?

showierdata9978:
it should be arbitrary, as the header in html spec is arbitrary

CST1229:
> s I understand it, you are proposing to add menu with options "string" and "json". If "string" is selected, the code that exists now is executed. If "json" is selected, BODY is converted to JSON by method: "BODY = JSON.stringify(BODY);" and then same code. Did I understand you correctly?

I think they mean a content type option that sets the `Content-Type` header to any string.

Samq64:
This should probably be merged with the fecth extension.

GarboMuffin:
Merge this into the existing `fetch` extension

GarboMuffin:
And keep the websocket pinging stuff in the websocket pinging extension, we don't need to duplicate it in the fetch extension.

A way to specify the Content-Type of the body will be required. A way to get the status code of the response and other response headers would also be interesting.

GarboMuffin (changes requested):
See my previous comments

PR #70, opened 2022-12-18
Itch.io API - Got open window to work

softedco:
image

This can be used maliciously, but the functionality of this block is limited due to it opening windows with this:
`/purchase?popup=1`

To limit this block even further we can remove the domain argument which is replaced with default itch.io, and since this is an entire new block, nothing breaks. This won't allow users to open their game purchase popup from other websites, but there is still `☁ open window` and `☁ redirect` stuff.

### DIFF
--- a/itchio.js
+++ b/itchio.js
@@ -208,18 +208,18 @@
 				}
 			};
 		}
-        openGameWindow(args) {
-		/* Passing just args will break this in the extension manager,
-		 * probably because of explicit name collision.
-		 */
-		openGameWindow({
-			user: args.user,
-			domain: args.domain,
-			game: args.game,
-			width: args.width,
-			height: args.height
-		});
-        }
+		openGameWindow(args) {
+			/* Passing just args will break this in the extension manager,
+			 * probably because of explicit name collision.
+			 */
+			openGameWindow({
+				user: args.user,
+				domain: args.domain,
+				game: args.game,
+				width: args.width,
+				height: args.height
+			});
+		}
 		getGameData(args) {
 			getGameData({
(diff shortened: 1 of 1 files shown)


GarboMuffin:
Does there really need to be an input for "itch.io"? Ideally that would be hardcoded

GarboMuffin:
I think it would be cool if the extension could provide a way for the project to choose whether to open the purchase page (eg https://night-school-studio.itch.io/oxenfree/purchase?popup=1), or just the project page (eg https://trianull.itch.io/starlight), or maybe whatever other pages itch has.

Is there a particular reason why it makes more sense to open a popup window instead of a new tab?

softedco:
> I think it would be cool if the extension could provide a way for the project to choose whether to open the purchase page (eg https://night-school-studio.itch.io/oxenfree/purchase?popup=1), or just the project page (eg https://trianull.itch.io/starlight), or maybe whatever other pages itch has. Probably a dropdown makes sense for that.
> 
> Is there a particular reason why it makes more sense to open a popup window instead of a new tab?

When adding that I reused attachBuyButton code from the API reference, it maked sense to be a popup since it was made specifically for a purchase button. Now that I think about it and now that this is a different block, I agree that we can expand it.

softedco:
image
These do the same thing.

image
This one opens the user page.

There isn't any way to fix this stuff well because it is not actually a bug.

GarboMuffin:
That's fine

GarboMuffin:
Just let me know when this is done and I'll press the green button

softedco:
It's done

PR #80, opened 2022-12-23
Add localStorage extension :)

StioStudio:
First time making an extension. I saw that someone else tried to make an localStorage extension, but didn't have a way do make an ID. So I made an ID block :D
Hope you like what I made :D

The ID block just makes it so that the thing you storage has an ID. So if you have an ID of "hi", then the ID of "hello" can not accses "hi".

### DIFF
--- a/localStorage.js
+++ b/localStorage.js
@@ -137,8 +137,3 @@ let localStorage_ID = undefined;
   }
   Scratch.extensions.register(new MyExtension());
 })(Scratch);
-
-
-// Call Scratch.extensions.register to register your extension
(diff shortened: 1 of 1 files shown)


softedco:
This is still exploitable and can introduce collision,
ideally, an ID should be unique and inaccessible to the end user. What this does can already be done in the original extension by adding a prefix to each key input, although thanks for your effort

GarboMuffin:
see above comment

we want collision avoidance to be "automated"; users shouldn't have to think about it at all

PR #91, opened 2022-12-25
Add cloudlink icon

softedco:


softedco:
> Note that that image is probably of CloudLink 3; CL4 has newer branding (see: [Cloudlink 4's official repo](https://github.com/MikeDev101/cloudlink))

Yeah I know, there is just no other extension icon I could find.

GarboMuffin:
I'd like to avoid touching GPL'd assets as much as possible so that this repository is unambiguously not affected by copyleft. I've included a different image in #93 that is similar but legally distinct and also an SVG.

Issue #46, opened 2022-11-24
For the cursor blocks, add an option to set the cursor to the current image of the sprite


PPPDUD:
Sorry for the long title.

PR #95, opened 2022-12-25
cursor: Allow setting cursor to custom images

GarboMuffin:
Closes https://github.com/TurboWarp/extensions/issues/46

https://user-images.githubusercontent.com/33787854/209484836-f11d7b07-80df-4934-a46a-691564ee7074.mp4

Also fixes cursor being reset when scratch-gui resizes or goes in/out of fullscreen

### DIFF
--- a/extensions/cursor.js
+++ b/extensions/cursor.js
@@ -30,23 +30,18 @@
     }
   };
   const getRawSkinCanvas = lazilyCreatedCanvas();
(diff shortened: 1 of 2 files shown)


GarboMuffin:
The big remaining question is: when someone sets the cursor to a sprite's costume, what should the `cursor` block return? I'm probably just gonna go with `costume1`.

Issue #92, opened 2022-12-25
MP3 player extension


Mickmick2012:
An extension that can can play mp3s via url and hard drive
i know that there is a Spotify extension from Scratch X
Example block:
![g891](https://user-images.githubusercontent.com/110345721/209468338-26598c2b-3578-49fb-96c7-1c1e0b203e6a.svg)

menu selection:
From URL
From File

SIPC:
wow

Mickmick2012:
i tried to do a prototype (with ABSOLUTELY NO KNOWLEDGE of java) with the spotify extension

SIPC:
Lazy loading?

softedco:
> Lazy loading?

I think it's just a suggestion

Mickmick2012:
Closing this as the extension is added

PR #100, opened 2022-12-27
Add sound extension

softedco:
image
As suggested in #92
Playing sound from device didn't make it there because chrome blocks it.


Mickmick2012:
is it possible to add "play sound from url until done" ?

softedco:
> is it possible to add "play sound from url until done" ?

Yes

CST1229:
#100

GarboMuffin:
Rewrite this to use Scratch's AudioContext `vm.runtime.audioEngine.audioContext` instead of plain Audio elements. I think there's something like a "media element source". Otherwise this won't be able to reliably play audio in browsers that have autoplay disabled.

GarboMuffin (changes requested):
See above

GarboMuffin (commented):
Made some changes:

 - It will first try to play the sound using Scratch's audio engine, which means it'll integrate with volume, pitch, pan left/right, etc.
 - If the sound can't be played that way (most likely because the URL does not support CORS), it will instead just use an <audio> element, like the original version.

SIPC:
Is it possible to add "pause [url]" and 'suspend all'?

softedco:
> Is it possible to add "pause [url]" ？

From the audio element it is, but I don't know if it is possible with the audioEngine.

GarboMuffin:
It will be possible in audio engine, just a good bit more annoying

GarboMuffin:
We'll do pausing later

GarboMuffin:
Whoever wants pausing: Open a new issue and describe which exact blocks you would like. There's a lot of ways this could be done and I don't want to have to change it later.

PR #113, opened 2023-01-05
CrazyGames SDK Extension

softedco:
[SDK](https://sdk.crazygames.com/crazygames-sdk-v1.js)
[Docs](https://docs.crazygames.com/sdk/html5)

crazygames

Should work from the editor right away, but it won't get link data, check if the user has AdBlock and display ads, instead there would be an overlay:

overlay

Unfortunately, once you request an ad, the only way to get rid of the overlay is to refresh the page.


GarboMuffin:
I imagine that their SDK is very much not open source, yes?

softedco:
> I imagine that their SDK is very much not open source, yes?

As far as I can tell, the only publicly available version is minified.

imya505222:
so extension will be added ?

softedco:
> so extension will be added ?

I emailed crazygames developer support for license information, we'll see how it goes

GarboMuffin:
I am assuming that the library is not available under a license that is GPL compatible, so unfortunately I am unable to include it here.

I will say that, regardless of license, accommodating a purely commercial game site littered with ads in this free open source project is not a high priority.

Anyone who is interested in this sort of integration is free to contact: contact@turbowarp.org

Martinelplayz:
i imagine i could make use if this if everything were to work the way it should.

PR #115, opened 2023-01-06
Add Cookie.js Extension

AlexSchoolOH:
This is a port of my extension Cookie.js yes this is david-orangemoon(ObviousAlexC) I am not at home right now so I am making this pull request on my alt account.
Here is a concept logo that I made for it but I might make a better one when I get home.
CookieLogo

### DIFF
--- a/extensions/Cookie.js
+++ b/extensions/Cookie.js
@@ -57,13 +57,6 @@
                             }                  
                         },
                         {
-                            "opcode": "readallcookies",
-                            "blockType": "reporter",
-                            "text": "read all cookies",
-                            "arguments": {
-                            }
-                        },
-                        {
                             "opcode": "readcookieofname",
                             "blockType": "reporter",
                             "text": "read cookie by the name of [c]",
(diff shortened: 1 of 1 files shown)


softedco:
This extension has the same problem as the localStorage extension #55, which is the fact that it can cause collision and is exploitable, but thanks for your effort, if this issue resolves, cookies can become a good addition to the localStorage extension referenced above.

AlexSchoolOH:
> This extension has the same problem as the localStorage extension #55, which is the fact that it can cause collision and is exploitable, but thanks for your effort, if this issue resolves, cookies can become a good addition to the localStorage extension referenced above.

Oh thanks! I should probably try to make a fix for the exploits! I think I have an idea for that too.

softedco:
> Oh thanks! I should probably try to make a fix for the exploits! I think I have an idea for that too.

There is already a fix for that, see https://github.com/TurboWarp/extensions/pull/55#issuecomment-1364454640

AlexSchoolOH:
> > Oh thanks! I should probably try to make a fix for the exploits! I think I have an idea for that too.
> 
> There is already a fix for that, see [#55 (comment)](https://github.com/TurboWarp/extensions/pull/55#issuecomment-1364454640)

I didn't see that I will see if I could commit the cookie part of cookies (and get permission) to do that when I get home. I got to get back to class.

showierdata9978 (changes requested):
you should expand the single char argument names, as that makes it hard to know what it means.

CST1229:
I don't think you should use cookies for local storage (`localStorage` is the intended way to store things locally; cookies are sent to the server every request).

AlexSchoolOH:
> I don't think you should use cookies for local storage (`localStorage` is the intended way to store things locally; cookies are sent to the server every request).

Cookies are good for people who want to link accounts to their games though like if somebody downloads a game off gamejolt or even has an api that uses another site for login eg google, or for custom social sites like meower they could be good for logins that eventually wear off. But I am going to close this request now.

PR #122, opened 2023-01-08
Add network and network-m extensions

ghost:
network banner

Extension contains a lot of blocks for working with the network:

Screenshot_192

`network` is now more global, so combine it with `fetch` would be wrong. Please add it separately.

PS: I tried to add all the suggestions here, if I missed something or misunderstood, please write.

PPS: I think it would be good to leave the ping block, because adding it as a separate extension is not very convenient.

PPPS: Thanks @softedco for optimizing the code


softedco:
I believe most of the blocks won't be added because:
- Server pinging extension already exists
- Fetch related blocks can be moved to the fetch extension
- Open window blocks can be replaced with special behaviour cloud variables and also they can be used mauliciously but I'm not sure if that's the case

extensions/true-fantom/network/network-v001.js
```diff
@@ -0,0 +1,477 @@
...
+    }
+    current_url_block() {
+      return document.URL || '';
+    }
+    network_type_block() {
+      let type = navigator.connection.type;
+      return type === undefined || type === 'none' || type === 'unknown' ? '' : type;
+    }
+    network_generation_block() {
+      return navigator.connection.effectiveType || '';
```
GarboMuffin, line 431: These blocks need to gracefully fallback when someone using a browser that does not have `navigator.connection` such as Firefox

extensions/true-fantom/network/network-v001.js
```diff
@@ -0,0 +1,477 @@
...
+            }
+          }
+        ],
+        
+        menus: {
+          content_type: { 
+            acceptReporters: true,
+            items: [
+              {
+                text: '(1) text',
```
GarboMuffin, line 304: You don't *need* to use numbers here, by the way

extensions/true-fantom/network/network-v001.js
```diff
@@ -0,0 +1,477 @@
...
+    open_window_block({URL,WIDTH,HEIGHT,LEFT,TOP}) {
+      let params = 'popup=1';
+      params += isNaN(WIDTH) ? '' : `,width=${Number(WIDTH) < 100 ? 100 : Number(WIDTH) > window.screen.width ? window.screen.width : Number(WIDTH)}`;
+      params += isNaN(HEIGHT) ? '' : `,height=${Number(HEIGHT) < 100 ? 100 : Number(HEIGHT) > window.screen.height ? window.screen.height : Number(HEIGHT)}`;
+      params += isNaN(LEFT) ? '' : `,left=${Number(LEFT) < 0 ? 0 : Number(LEFT) > window.screen.width ? window.screen.width : Number(LEFT)}`;
+      params += isNaN(TOP) ? '' : `,top=${Number(TOP) < 0 ? 0 : Number(TOP) > window.screen.height ? window.screen.height : Number(TOP)}`;
+      window.open(URL, '_blank', params);
+    }
+    redirect_link_block({URL}) {
+      window.open(URL, '_self');
```
GarboMuffin, line 472: This needs to protect against code execution when someone navigates to `javascript:alert(1)`

extensions/true-fantom/network/network-v001.js
```diff
@@ -0,0 +1,477 @@
...
+      return navigator.connection.downlink || '';
+    }
+    downlink_max_speed_block() {
+      return navigator.connection.downlinkMax || '';
+    }
+    network_rtt_block() {
+      return navigator.connection.rtt || '';
+    }
+    ping_block({SERVER}) {
+      return cached_ping_web_socket(String(SERVER));
```
GarboMuffin, line 443: I question whether this needs to be here, but since it seems like you want this to be a personal extension more than anything, I won't stop you from including it

extensions/true-fantom/network/network-v001.js
```diff
@@ -0,0 +1,477 @@
...
+          }
+        }
+      }
+    }
+    
+    connected_to_internet_block() {
+      return navigator.onLine;
+    }
+    browser_block() {
+      let has = input => navigator.userAgent.includes(input);
```
GarboMuffin, line 408: Did you write this yourself or copy it from somewhere else? We should credit where we copy things from if that's the case

softedco: Most of the changes from the update commit are from me, including this one

extensions/true-fantom/network/network-v001.js
True-Fantom: Yeah. This was in the version before "optimization"
I thought that `return something || ''` would work as `try catch`. I apologize for not checking this change.

And judging by the compatibility table, `try catch` will also be necessary for `navigator.onLine` block.

Ps: Can errors occur with `navigator.userAgent` `document.URL` `window.open` and `fetch()`?

extensions/true-fantom/network/network-v001.js
True-Fantom: Why? It seems convenient

extensions/true-fantom/network/network-v001.js
True-Fantom: Thanks for understanding)

Ps: It's not that I want to make an extension just for me. On the contrary, I would be glad if someone will use it. What I want is to achieve practicality and flexibility, but this cannot be fully achieved without a personal directory.

extensions/true-fantom/network/network-v001.js
True-Fantom: I copied and modified this code `https://developer.mozilla.org/en-US/docs/Web/API/Window/navigator`
(in version before @softedco 's changes)

extensions/true-fantom/network/network-v001.js
True-Fantom: Can you please clarify how I can do this?

GarboMuffin: Make a URL object:

`const url = new URL(URL)`

Make sure the protocol is something that makes sense:

`if (url.protocol === 'https:' || url.protocol === 'http:' || url.protocol === 'data:' || url.protocol === 'file:') window.open(...)`

extensions/true-fantom/network/network-v001.js
GarboMuffin: You may want to keep in mind that:

 - new URL() throws if the URL is invalid
 - fetch() returns a rejected promise in several conditions. you should attempt to handle this.

GarboMuffin: navigator.onLine and document.URL are supported basically everywhere

 - https://caniuse.com/mdn-api_navigator_online
 - https://caniuse.com/mdn-api_document_url

True-Fantom:
I have temporarily deleted ping block. In the future, I want to refine it a lot.

True-Fantom:
It seems that the extension is ready for publication. I hope there are no mistakes in it.

extensions/true-fantom/network/network-v001.js
```diff
@@ -0,0 +1,503 @@
...
+  const browser_protocols = ['chrome:', 'edge:', 'brave:', 'browser:', 'about:', 'extension:', 'chrome-extension:'];
+  const special_protocols = ['steam:', 'spotify:', 'zoommtg:', 'zoomus:', 'viber:', 'slack:', 'trueconf:', 'rdar:', 'msteams:', 'teamspeak:', 'magnet:', 'streamdeck:', 'skype:', 'sms:', 'comgooglemaps:', 'comgooglemapsurl:', 'comgooglemaps-x-callback:'];
+  const ms_protocols = ['ms-help:', 'ms-settings', 'ms-settings-airplanemode:', 'ms-settings-bluetooth:', 'ms-settings-camera:', 'ms-settings-cellular:', 'ms-settings-cloudstorage:', 'ms-settings-emailandaccounts:', 'ms-settings-language:', 'ms-settings-location:', 'ms-settings-lock:', 'ms-settings-nfctransactions:', 'ms-settings-notifications:', 'ms-settings-power:', 'ms-settings-privacy:', 'ms-settings-proximity:', 'ms-settings-screenrotation:', 'ms-settings-wifi:', 'ms-settings-workplace:', 'ms-access:', 'ms-excel:', 'ms-infopath:', 'ms-powerpoint:', 'ms-project:', 'ms-publisher:', 'ms-spd:', 'ms-visio:', 'ms-word:', 'ms-clock:', 'ms-calculator:', 'ms-windows-store:'];
+  const protocols = [...main_protocols, ...browser_protocols, ...special_protocols, ...ms_protocols];
```
GarboMuffin, line 10: credit where you got these lists from if you didn't come up with it yourself

GarboMuffin:
Sorry for all the massive delays on this

I question whether it is necessary to name it `network-m/network-m-v001.js`, but okay, this is your namespace, do as you please

See the one review comment above and I'll merge it

GarboMuffin:
you could consider changing the ID of the extensions internally as `network` is vague and sort of defeats the purpose of the username-specific folders

extensions/true-fantom/network/network-v001.js
GarboMuffin: reminder

GarboMuffin:
I attempted to finish this up for you but it appears you did not check the box to let maintainers edit this PR for you

True-Fantom:
Sorry for the delay, I didn't answer due to my health condition.

extensions/true-fantom/network/network-v001.js
True-Fantom: I used https://en.wikipedia.org/wiki/List_of_URI_schemes and https://www.youtube.com/watch?v=WcLmMLqnKCM

True-Fantom: I think the list should be left, since without it the block's capabilities are severely limited.

True-Fantom:
> you could consider changing the ID of the extensions internally as `network` is vague and sort of defeats the purpose of the username-specific folders

Yes, i agree

True-Fantom:
> Sorry for all the massive delays on this
> 
> I question whether it is necessary to name it `network-m/network-m-v001.js`. That's a bit of a mouthful and largely redundant, but okay, this is your namespace, do as you please
> 
> See the one review comment above and I'll merge it

I propose to leave the versions, but reduce the spelling to `network/network-01.js` and `network/network-01m.js`

True-Fantom:
And I also think that these blocks
Screenshot_195
go beyond the network, so can I move them to the `window.js`?
With additional blocks in this topic of course.

GarboMuffin:
Feel free to submit whatever extensions to your user folder that you see fit

I assume you are okay with having your name and Scratch profile listed like this?

image

True-Fantom:
> Feel free to submit whatever extensions to your user folder that you see fit
> 
> I assume you are okay with having your name and Scratch profile listed like this?
> 
> image

Yes, I'm all for it!

DNin01:
Scratch have apparently added the "connected to internet?" block officially!
https://github.com/scratchfoundation/scratch-editor/blob/f1e1278bff8dc89cedbf52098c5a44fb513ceabd/packages/scratch-vm/src/blocks/scratch3_sensing.js#L335-L343

Brackets-Coder:
> Scratch have apparently added the "connected to internet?" block officially! https://github.com/scratchfoundation/scratch-editor/blob/f1e1278bff8dc89cedbf52098c5a44fb513ceabd/packages/scratch-vm/src/blocks/scratch3_sensing.js#L335-L343

So it seems they have

PR #125, opened 2023-01-09
Merge branch 'master' of https://github.com/TurboWarp/extensions

obaydmerz:


Mickmick2012 (changes requested):
Why Change "Turbowarp" to "TinyPatch" Everywhere ? There is no change apart from this !

Mickmick2012:
I don't think "Turkbowarp" is a thing

GarboMuffin:
It looks like TinyPatch intends to be a mod of TurboWarp.

You shouldn't send your branding changes to us (but do consider sending us actual code if you're in the mood)

PR #128, opened 2023-01-10
The warping update and the depracation of 3 blocks!

David-Orangemoon:
3 blocks relating to stage size have been deprecated but not removed just to keep compatibility with older versions of pen+. A new block has been added allowing for creators to fix texture warping in their 3d projects.

### DIFF
--- a/extensions/penplus.js
+++ b/extensions/penplus.js
@@ -1,2185 +1,2255 @@
 (function(Scratch) {
-  'use strict';
-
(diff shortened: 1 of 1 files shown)


David-Orangemoon:
Also I have sorted the blocks based on their catagory and what they do.

David-Orangemoon:
I had the code wrong oops.

PR #129, opened 2023-01-10
Update for penplus.

David-Orangemoon:


David-Orangemoon:
3 blocks relating to stage size have been deprecated but not removed just to keep compatibility with older versions of pen+. A new block has been added allowing for creators to fix texture warping in their 3d projects.Also I have sorted the blocks based on their catagory and what they do.

Issue #90, opened 2022-12-25
Create images for all featured extensions


GarboMuffin:
If you look at https://extensions.turbowarp.org/ you'll notice that a of extensions use a placeholder image instead of having a custom one.

The problem is I genuinely don't know what to put there for most of them, but maybe some of you have ideas.

Extensions must be:

 - 2:1 aspect ratio
 - SVG (preferred), PNG, or JPG
 - If PNG or JPG, 600x300 resolution

You can post proposed images here or try to figure out how to submit a pull request if you're in the mood

CST1229:
Files:
![](https://user-images.githubusercontent.com/68464103/209532116-e9836525-4afd-4d89-a424-c7d72e0c1ca8.png)

DNin01:
How about this for Runtime Options?

![60 FPS option turned on](https://user-images.githubusercontent.com/106490990/209746149-a83fb609-1396-4131-a2fe-6e128f84a8ca.png)

GarboMuffin:
Can you make that again but without the top few pixels of the next checkbox visible? (better yet, you might be able to just fit it in?)

GarboMuffin:
Actually, I think this will look a bit better if I remake it as an SVG and make some changes to the design, removing the question marks, making the highlight a round rectangle on both sides, etc.

DNin01:
> Actually, I think this will look a bit better if I remake it as an SVG and make some changes to the design, removing the question marks, making the highlight a round rectangle on both sides, etc.

Go ahead.

David-Orangemoon:
![UtilitiesBITMAP](https://user-images.githubusercontent.com/76855369/210033040-4adf3d2e-903e-416f-8b3b-32fbe6607e53.png)
Utilities!

David-Orangemoon:
![FETCHBITMAp](https://user-images.githubusercontent.com/76855369/210033794-2833d868-8ed1-4423-84f9-0d3da825b529.png)
Fetch!

David-Orangemoon:
![CloudPingBITMAP](https://user-images.githubusercontent.com/76855369/210034608-14bca117-b74a-4ebd-9b3b-ead178be8dec.png)
Cloud Ping
Its simple but at least it conveys its purpose!

David-Orangemoon:
![PointerLockBitmap](https://user-images.githubusercontent.com/76855369/210035268-59b0a1c7-c365-4f10-8bc5-2b6dacd44d8d.png)
[...]

PR #132, opened 2023-01-10
Add Encoding extension

SIPC:
image


Xeltalliv:
Suggestion:
First 6 blocks in this extension are about different types of text encoding and the other 3 are about working with time, which isn't what most people expect from the extension named math. Maybe it is worth to split it into 2 extensions called: _Encoding_ and _Time_.
_Time_ extension could be extended to have blocks like:
image

SIPC:
> Suggestion: First 6 blocks in this extension are about different types of text encoding and the other 3 are about working with time, which isn't what most people expect from the extension named math. Maybe it is worth to split it into 2 extensions called: _Encoding_ and _Time_? _Time_ extension could be extended to have blocks like: image

I will try to improve!

SIPC:
image

GarboMuffin:
I think that the encoding blocks and the time blocks should be put in separate extensions

SIPC:
> I think that the encoding blocks and the time blocks should be put in separate extensions

Removed blocks over time

SIPC:
image

GarboMuffin:
Where did you get the code for the md5 function here? We need to credit where we copy code from.

GarboMuffin:
I am not sure how much the random generation blocks belong here, but once it's clarified where the md5 script comes from, I can merge this

SIPC:
md5 scripts can be found everywhere on the web, here is the md5 script from [Wie erzeugt man einen MD5-Dateihash in JavaScript/Node.js?](https://stackoverflow.com/questions/14733374/how-to- generate-an-md5-file-hash-in-javascript-node-js)

GarboMuffin:
I'm very sorry for all the delay on this

If you want, we can probably rename this to `SIPC/encoding.js`, and you can go add back all the blocks that I had asked to be removed, or I can just merge it as-is

About the MD5 function: Note that StackOverflow's content license is not compatible with the GPLv3 that we require code to be compatible with, so I've replaced the the function with one that is. Note that just because there are a lot of MD5 functions on the internet doesn't mean we can just copy one without checking its license.

SIPC:
That's it! :)

GarboMuffin:
Do you want your name and/or link to be listed here, and if so, what should those be? (scratch.mit.edu links okay, other links may not be)

image

SIPC:
> Do you want your name and/or link to be listed here, and if so, what should those be? (scratch.mit.edu links okay, other links may not be)
> 
> image

yes

SIPC:
I have posted pictures about the extension in [Help us create images for all of the extensions](https://github.com/TurboWarp/extensions/issues/90)

PR #142, opened 2023-01-11
Add JS Extension

RayFB-13:
Adds an extension that lets you run Javascript from Turbowarp.


softedco:
From [CONTRIBUTING.md](https://github.com/TurboWarp/extensions/blob/master/CONTRIBUTING.md)
> Extensions must not use `eval()`, `new Function()`, untrusted `<script>` or `<iframe>` tags, or similar arbitrary JS/CSS/HTML/etc.

PR #143, opened 2023-01-11
Add Run JS extension [Replaced eval-no new function()]

RayFB-13:
Adds an extension that runs JS.


showierdata9978:
> It's still arbitrary js which can be used mauliciously.

it would really only work with already injected js.  This only runs functions.

GarboMuffin:
Which functions is this actually useful for?

showierdata9978:
basically none, as it can only run global objects

CST1229:
> > It's still arbitrary js which can be used mauliciously.
> 
> it would really only work with already injected js. This only runs functions.

`eval (eval) (alert("hakd"))`

GarboMuffin:
It does seem the code refuses to run eval now so maybe it is secure (maybe)

My concern here is that if this is only useful for things like `alert`, `confirm`, and `prompt`, it would make more sense to just have an extension with specific blocks for those.

Xeltalliv:
> It does seem the code refuses to run eval now so maybe it is secure (maybe)

image

GarboMuffin:
I'm closing this because making this safe is not trivial and the usefulness is unclear per above discussions.

You are welcome and encouraged to submit extensions that expose safe and intentional wrappers around functions on window or any other Web API

PR #153, opened 2023-01-17
Add extension Outfitted.js

David-Orangemoon:
Outfitted is a scratch extension that allows for costume modification through your scratch blocks.

This extension will get updates in the future if accepted.
[SVG]
Outfitted
[PNG]
Outfitted-Bitmap

### DIFF
--- a/extensions/Outfitted.js
+++ b/extensions/Outfitted.js
@@ -1,0 +1,110 @@
+(function(Scratch) {
+    'use strict';
+  
+    if (!Scratch.extensions.unsandboxed) {
+      throw new Error('Pen+ must be run unsandboxed');
+    }
+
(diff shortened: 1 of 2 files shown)


GarboMuffin:
haven't tried this yet (really busy), but first look through the code:

 - eslint is complaining about style issues; not critical but easy to fix
 - replace read me block with `docsURI` to get a view documentation. You can put something like `outfitted.html` in `website` (until we get a more proper solution) then link to https://extensions.turbowarp.org/outfitted.html

GarboMuffin:
Also see what happens if someone tries to set their costumes to weird things like `<script>alert(1)</script>`. I don't remember whether sanitization is something that happens in VM or renderer

David-Orangemoon:
codebase. Didnt think about that thanks for notifying me

GarboMuffin:
I didn't actually test what happened if someone tries to put script tags; it's just something that the extension needs to ensure is handled as it is possible to handle

Martinelplayz:
this looks cool

SurvExe1Pc:
> I didn't actually test what happened if someone tries to put script tags; it's just something that the extension needs to ensure is handled as it is possible to handle

it seems to be good!
image

Code used:
image
xss attack i used to test this: https://youtu.be/gVrdE6g_fa8

LilyMakesThings:
> this looks cool

It's superseded by Looks Plus, which has similar functionality

SurvExe1Pc:
when will there be bitmap support?

Drago-Cuven:
ah I remember this. outdated by looks+ correct?

PR #155, opened 2023-01-17
Update penplus.js to version 5

David-Orangemoon:
Changelog:
Added Line blocks
Fixed Color Blocks
Seperated Blocks into catagories just so I could edit them in the future without having to dig through a large blocks array! Fixed support for the Plugin Loader
Made cross platform with the same file by using conditional assignments! Confirmed some doubts (see line 52) 
Learned that javascript had C++ like conditional assignments Depracated spacial transformation block
Added spacial changing block
Other various small fixes

### DIFF
--- a/extensions/penplus.js
+++ b/extensions/penplus.js
@@ -2663,7 +2663,7 @@ Other various small fixes
 
   if (isPlugin) {
     (function () {
-      var extensionInstance = new PenPlus(window.vm.extensionManager.runtime);
+      var extensionInstance = new PenPlus();
       var serviceName = window.vm.extensionManager._registerInternalExtension(extensionInstance);
       window.vm.extensionManager._loadedExtensions.set(extensionInstance.getInfo().id, serviceName);
     })();
--- a/extensions/penplus.js
+++ b/extensions/penplus.js
@@ -19,2585 +19,2666 @@ Added spacial changing block
 Other various small fixes
 */
 (function(Scratch,isPlugin) {
-    'use strict';
+  'use strict';
 
-    let canvas = null
-    let gl = null
-    
-    const ScratchRuntime = (isPlugin) ? Scratch.runtime:Scratch.vm.runtime
-    //Detect if we are using the plugin loader or extensions.turbowarp.org!
-    if (Scratch) {
-      if (!isPlugin){
-        canvas = Scratch.renderer.canvas
-        gl = Scratch.renderer._gl
-        if (!Scratch.extensions.unsandboxed) {
-          throw new Error('Pen+ must be run unsandboxed');
-        }
-      }
-      else{
-        canvas = document.querySelector("canvas")
-        gl = canvas.getContext("webgl")
+  let canvas = null;
+  let gl = null;
+  
+  const ScratchRuntime = (isPlugin) ? Scratch.runtime : Scratch.vm.runtime;
+  //Detect if we are using the plugin loader or extensions.turbowarp.org!
+  if (Scratch) {
+    if (!isPlugin) {
+      canvas = Scratch.renderer.canvas;
+      gl = Scratch.renderer._gl;
+      if (!Scratch.extensions.unsandboxed) {
+        throw new Error('Pen+ must be run unsandboxed');
       }
     }
-    //if we can't detect scratch alert the user
     else{
-      alert("Pen+ must be ran in a scratch based enviornment!")
+      canvas = document.querySelector("canvas");
+      gl = canvas.getContext("webgl");
     }
-
-    const EXAMPLE_IMAGE = 'https://extensions.turbowarp.org/dango.png';
-  
-    const blankImage = "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAIAAAACCAYAAABytg0kAAAAAXNSR0IArs4c6QAAABRJREFUGFdj/P///38GBgYGRhgDAFfVB/vDfnUlAAAAAElFTkSuQmCC"
-
-    // TODO: see how these differ from Scratch, if at all
-    // Note to Garbo or any code checker it does it uses bilinear filtering!
-    gl.enable(gl.BLEND);
-    gl.blendEquation(gl.FUNC_ADD);
-    gl.blendFunc(gl.ONE_MINUS_CONSTANT_ALPHA, gl.ONE_MINUS_SRC_ALPHA);
-  
-    var stampWidth = 64;
-    var stampHeight = 64;
-
-    var lineWidth = 1;
-    var lineColor = {
-      r:1,
-      g:1,
-      b:1,
-      a:1
-    }
-  
-    var screenWidth = 480;
-    var screenHeight = 360;
-    var coordinateSpace = "Canvas"
-  
-    var stampRotation = 90;
-    var stampOffset = [0,0]
-  
-    const m4 = (function() {
-      /*!
-       * 4x4 matrix operation code is from https://webglfundamentals.org/webgl/resources/m4.js
-       * We have made some changes:
-       *  - Fixed type errors
-       *  - Changed code formatting
-       *  - Removed unused functions
-       *
-       * Copyright 2021 GFXFundamentals.
-       * All rights reserved.
-       *
-       * Redistribution and use in source and binary forms, with or without
-       * modification, are permitted provided that the following conditions are
-       * met:
-       *
-       *     * Redistributions of source code must retain the above copyright
-       * notice, this list of conditions and the following disclaimer.
-       *     * Redistributions in binary form must reproduce the above
-       * copyright notice, this list of conditions and the following disclaimer
-       * in the documentation and/or other materials provided with the
-       * distribution.
-       *     * Neither the name of GFXFundamentals. nor the names of his
-       * contributors may be used to endorse or promote products derived from
-       * this software without specific prior written permission.
-       *
-       * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
-       * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
-       * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
-       * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
-       * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
-       * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
-       * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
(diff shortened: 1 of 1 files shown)


GarboMuffin:
I will have to think about this one a bit more:

 - One extension adding multiple categories is not something we've wanted before
 - We generally don't want "plugin" related code; this system is designed to replace plugins

Not saying that can't happen. I'll just have to think about it

David-Orangemoon:
> I will have to think about this one a bit more:
> 
>  - ~~One extension adding multiple categories is not something we've wanted before~~ I misunderstood what you did with categories. Disregard this point.
>  - We generally don't want "plugin" related code; this system is designed to replace plugins
> 
> Not saying that can't happen. I'll just have to think about it

I just made the code work for both (it's still mostly unsandboxed extension code) but I have it on both lstv's plugin loader and extensions.turbowarp.org and I didnt want to have to maintain 2 different codebases.

showierdata9978:
i agree that both ways should be allowed; so other scratch mods can still load via the browser console, and script injectors.

(the code that I specifically use)

```js
        // major credit to CST for this (I did modify it some to allow sandboxed if you only want the speed boost (:
	const extensionClass = YourExt;
	if (Scratch) {
		if (Scratch.extensions.unsandboxed) {
			Scratch.extensions.register(new extensionClass(Scratch.vm));
		} else {
                        // comment out if you use VM things
			Scratch.extensions.register(new extensionClass());
                        // throw new Error("YourExt cannot run in sandboxed mode.");
		} 
	} else if (globalThis.vm) {
		// Support loading the extension "the old way"
		// (running the code in something like the browser console
		// or E羊icques' load_plugin URL parameter)
		const extensionInstance = new extensionClass(globalThis.vm);
		const serviceName = globalThis.vm.extensionManager._registerInternalExtension(
			extensionInstance
		);
		globalThis.vm.extensionManager._loadedExtensions.set(
			extensionInstance.getInfo().id, serviceName
		); 
	} else {
		throw new Error("Scratch Not detected"); // no idea if there is anything else i can do here
	};
```

showierdata9978:
> > i agree that both ways should be allowed; so other scratch mods can still load via the browser console, and script injectors.
> > (the code that I specifically use)
> > ```js
> >         // major credit to CST for this (I did modify it some to allow sandboxed if you only want the speed boost (:
> > 	const extensionClass = YourExt;
> > 	if (Scratch) {
> > 		if (Scratch.extensions.unsandboxed) {
> > 			Scratch.extensions.register(new extensionClass(Scratch.vm));
> > 		} else {
> >                         // comment out if you use VM things
> > 			Scratch.extensions.register(new extensionClass());
> >                         // throw new Error("YourExt cannot run in sandboxed mode.");
> > 		} 
> > 	} else if (globalThis.vm) {
> > 		// Support loading the extension "the old way"
> > 		// (running the code in something like the browser console
> > 		// or E羊icques' load_plugin URL parameter)
> > 		const extensionInstance = new extensionClass(globalThis.vm);
> > 		const serviceName = globalThis.vm.extensionManager._registerInternalExtension(
> > 			extensionInstance
> > 		);
> > 		globalThis.vm.extensionManager._loadedExtensions.set(
> > 			extensionInstance.getInfo().id, serviceName
> > 		); 
> > 	} else {
> > 		throw new Error("Scratch Not detected"); // no idea if there is anything else i can do here
> > 	};
> > ```
> 
> (this is david-orangemoon) Personally use a different method but this could work too.

it looks like mine works with the console, as if there is no `Scratch` object , it looks for the VM in globalThis

AlexSchoolOH:
> > > i agree that both ways should be allowed; so other scratch mods can still load via the browser console, and script injectors.
> > > (the code that I specifically use)
> > > ```js
> > >         // major credit to CST for this (I did modify it some to allow sandboxed if you only want the speed boost (:
> > > 	const extensionClass = YourExt;
> > > 	if (Scratch) {
> > > 		if (Scratch.extensions.unsandboxed) {
> > > 			Scratch.extensions.register(new extensionClass(Scratch.vm));
> > > 		} else {
> > >                         // comment out if you use VM things
> > > 			Scratch.extensions.register(new extensionClass());
> > >                         // throw new Error("YourExt cannot run in sandboxed mode.");
> > > 		} 
> > > 	} else if (globalThis.vm) {
> > > 		// Support loading the extension "the old way"
> > > 		// (running the code in something like the browser console
> > > 		// or E羊icques' load_plugin URL parameter)
> > > 		const extensionInstance = new extensionClass(globalThis.vm);
> > > 		const serviceName = globalThis.vm.extensionManager._registerInternalExtension(
> > > 			extensionInstance
> > > 		);
> > > 		globalThis.vm.extensionManager._loadedExtensions.set(
> > > 			extensionInstance.getInfo().id, serviceName
> > > 		); 
> > > 	} else {
> > > 		throw new Error("Scratch Not detected"); // no idea if there is anything else i can do here
> > > 	};
> > > ```
> > 
> > 
> > (this is david-orangemoon) Personally use a different method but this could work too.
> 
> it looks like mine works with the console, as if there is no `Scratch` object , it looks for the VM in globalThis
The method I used also works in the console by using conditional assignments for variables. but it still pretty simple when it gets to the basics  of things.

`
(function(Scratch,isPlugin) {
    'use strict';

    let canvas = null // example of enviornment variables setup just setting them to constant
    let gl = null // probaby a better way to do this part with constants but too lazy
    
    const ScratchRuntime = (isPlugin) ? Scratch.runtime:Scratch.vm.runtime // grabbing the runtime
    //Detect if we are using the plugin loader or extensions.turbowarp.org!
    if (Scratch) {
      if (!isPlugin){
        // define the normal non plugin variables here
        canvas = Scratch.renderer.canvas
        gl = Scratch.renderer._gl
        if (!Scratch.extensions.unsandboxed) {
          throw new Error('Pen+ must be run unsandboxed');
        }
      }
      else{
        // You can define you extension variables for the plugin variation here because paths are slightly different
        canvas = document.querySelector("canvas")
        gl = canvas.getContext("webgl")
      }
    }
    //if we can't detect scratch alert the user
    else{
      alert("Extension must be ran in a scratch based enviornment!")
    }

    const TWTypes = (isPlugin) ? 
    //PluginLoader
    {
      BlockType:{
        COMMAND:'command',
        REPORTER:'reporter',
        BOOLEAN:'boolean',
        HAT:'hat'
      },
      ArgumentType:{
        STRING:'string',
        NUMBER:'number',
        COLOR:'color',
        ANGLE:'angle',
        BOOLEAN:'boolean',
        MATRIX:'matrix',
        NOTE:'note'
      }
    } 
    : 
    //Extensions.Turbowarp.org
    {
      BlockType:{
        COMMAND:Scratch.BlockType.COMMAND,
        REPORTER:Scratch.BlockType.REPORTER,
        BOOLEAN:Scratch.BlockType.BOOLEAN,
        HAT: Scratch.BlockType.HAT
      },
      ArgumentType:{
        STRING:Scratch.ArgumentType.STRING,
        NUMBER:Scratch.ArgumentType.NUMBER,
        COLOR:Scratch.ArgumentType.COLOR,
        ANGLE:Scratch.ArgumentType.ANGLE,
        BOOLEAN:Scratch.ArgumentType.BOOLEAN,
        MATRIX:Scratch.ArgumentType.MATRIX,
        NOTE:Scratch.ArgumentType.NOTE
      }
    }

    class Extension {
      getInfo () {
        return { /* ... */ };
      }
    }

    if(isPlugin){
      (function() {
        var extensionInstance = new Extension(window.vm.extensionManager.runtime)
        var serviceName = window.vm.extensionManager._registerInternalExtension(extensionInstance)
        window.vm.extensionManager._loadedExtensions.set(extensionInstance.getInfo().id, serviceName)
        })()
    }
    else{
      Scratch.extensions.register(new PenPlus());
    }
    
  })((typeof Scratch === 'undefined') ? window.vm:Scratch,typeof Scratch === 'undefined');
`

AlexSchoolOH:
> > > i agree that both ways should be allowed; so other scratch mods can still load via the browser console, and script injectors.
> > > (the code that I specifically use)
> > > ```js
> > >         // major credit to CST for this (I did modify it some to allow sandboxed if you only want the speed boost (:
> > > 	const extensionClass = YourExt;
> > > 	if (Scratch) {
> > > 		if (Scratch.extensions.unsandboxed) {
> > > 			Scratch.extensions.register(new extensionClass(Scratch.vm));
> > > 		} else {
> > >                         // comment out if you use VM things
> > > 			Scratch.extensions.register(new extensionClass());
> > >                         // throw new Error("YourExt cannot run in sandboxed mode.");
> > > 		} 
> > > 	} else if (globalThis.vm) {
> > > 		// Support loading the extension "the old way"
> > > 		// (running the code in something like the browser console
> > > 		// or E羊icques' load_plugin URL parameter)
> > > 		const extensionInstance = new extensionClass(globalThis.vm);
> > > 		const serviceName = globalThis.vm.extensionManager._registerInternalExtension(
> > > 			extensionInstance
> > > 		);
> > > 		globalThis.vm.extensionManager._loadedExtensions.set(
> > > 			extensionInstance.getInfo().id, serviceName
> > > 		); 
> > > 	} else {
> > > 		throw new Error("Scratch Not detected"); // no idea if there is anything else i can do here
> > > 	};
> > > ```
> > 
> > 
> > (this is david-orangemoon) Personally use a different method but this could work too.
> 
> it looks like mine works with the console, as if there is no `Scratch` object , it looks for the VM in globalThis

Also I might implement that in a quick patch I'm going to make when I get home that hopefully fixes most of the es errors

David-Orangemoon:
I got everything working and formatted so the linter won't throw a fit now

David-Orangemoon:
Ok I'm going to add buttons that link to the new docs.

David-Orangemoon:
Nevermind version 5 is ready

GarboMuffin:
I will allow this extension to have plugin loader compatibility code because it used to be a plugin prior to unsandboxed extensions existing. Other extensions will not be allowed to have similar code without justification as the compatibility code is hard to read and prone to bugs.

GarboMuffin:
A quick note on how I think you should go about making your extensions compatible with all the different environments, if you have an actual reason to:

 - Your extension either works in the sandbox or throws an error. I am not aware of any case where a middle ground is useful.
 - It is for the best that you use APIs like `Scratch.ArgumentType.BOOLEAN` so you don't spend time trying to debug why your block is broken because it's `"Boolean"` instead of `"boolean"` like all the others
 - It is perfectly fine to do things like `Scratch.vm` whenever you need to access the VM instead of a class property if you want

With these in mind, you should put something like this in the inner function right at the top: (incomplete sample)

```js
  if (!Scratch) {
    Scratch = {
      BlockType: {
        COMMAND: 'command',
        REPORTER: 'reporter',
        BOOLEAN: 'Boolean',
        HAT: 'hat'
      },
      ArgumentType: {
        STRING: 'string',
        NUMBER: 'number',
        COLOR: 'color',
        ANGLE: 'angle',
        BOOLEAN: 'Boolean',
        MATRIX: 'matrix',
        NOTE: 'note'
      },
      vm: window.vm,
      extensions: {
        unsandboxed: true,
        register: (object) => {
          const serviceName = vm.extensionManager._registerInternalExtension(object);
          vm.extensionManager._loadedExtensions.set(object.getInfo().id, serviceName);
        }
      }
    };
    if (!Scratch.vm) {
      throw new Error('The VM does not exist');
    }
  }
```

and you change the last line that actually calls the IIFE to

```js
})(window.Scratch);
```

then just write an extension like you would any other custom extension

Issue #172, opened 2023-01-27
String to image/sound extension


TheShovel:
A extension that creates sprites and sounds in sprites with string input, and also gives string data for a selected costume or sound. This could be used to save and edit user made images in cloud variables, and make it possible to import image/sound files using the Files extension.

TheShovel:
Related to #149

GarboMuffin:
This is now possible as https://github.com/TurboWarp/extensions/pull/173 was merged

Issue #149, opened 2023-01-14
Files extension issue with images and audio


TheShovel:
I wanted to make a extension that goes with the files extension.
Basically you can use the open file reporter block to open a image or sound, and have it be applied to the current editing target. Issue is that the block doesn't return image data, it returns a string.
A check for the extension of the file, and the type of data it sends would be nice, since having 2 extensions with different ways of opening files would be really messy.
If that's not something that you wanna do or can't do because of some kind of limitations, a way for me to add that in my extension is appreciated.

TheAwesome98-Real:
for sounds, maybe you could convert it into a data url and play it with the sounds extension?

TheShovel:
> for sounds, maybe you could convert it into a data url and play it with the sounds extension?

I have some blocks that can load sounds and sprites from a link in my utils extension, but I tried to make a data url to do what I just said, and it just doesn't work. It's not a valid link, even tho the code executes with no errors or warnings.

GarboMuffin:
This is now possible as https://github.com/TurboWarp/extensions/pull/173 was merged

PR #173, opened 2023-01-28
Block to output data uri for media files

TheShovel:
Adds a block that outputs data url for files.


TheShovel:
This is related to #149  and #172

softedco:
You mean data uri?

TheShovel:
> You mean data uri?

Yeah. I actually thought it was url until now because in js it is called dataurl.

GarboMuffin:
we will use an alternative approach by adding blocks like these:

image

which is also future-proof if we ever want a pure base64 option or whatever else

for now there will be some redundancy in here with the old blocks, but that's okay, eventually we can find a way to make the old blocks keep working but without being listed in the flyout

PR #178, opened 2023-02-02
Add Unicode Encoding Extension

CS2627883:
Add [Unicode Encoding](https://github.com/CS2627883) extension. This is an unsandboxed extension. It has been added to the extensions folder as well as the website.


GarboMuffin:
to list this on the website you will have to make an icon that is 3:2. it would be preferred to not use trademarks when possible

CS2627883:
> to list this on the website you will have to make an icon that is 3:2. it would be preferred to not use trademarks such as the unicode icon when possible

I can't think of a good alternative logo - I used unicode because that's what it is. Do you have any suggestions? I want the logo to look clean and professional - that's why I took it from the internet

GarboMuffin:
Having an extension with an icon that is the Unicode trademark and a name "Unicode Encoding" may be seen to imply a certain endorsement that does not exist. Is it not possible to find another icon?

It would seem to me that this extension may be better with a name like "Cloud variable encoding" instead of "Unicode Encoding". It encodes text into a sequence of 6 base-10 numbers, which is perfectly fine, but that's not what people think of when you say "unicode"

CS2627883:
> Having an extension with an icon that is the Unicode trademark and a name "Unicode Encoding" may be seen to imply a certain endorsement that does not exist. Is it not possible to find another icon?
> 
> It would seem to me that this extension may be better with a name like "Cloud variable encoding" instead of "Unicode Encoding". It encodes text into a sequence of 6 base-10 numbers, which is perfectly fine, but that's not what people think of when you say "unicode"
> 
> That it uses the unicode mappings to generate the string can be considered an implementation detail

Yes, I agree that the name is misleading. The original name was 'Encoding', but this conflicted with [another extension](https://github.com/TurboWarp/extensions/blob/master/extensions/encoding.js). I'm absolutely willing to choose a new icon based on the new name. Maybe numerical encoding would be a better name...? I'll change it to that.

CS2627883:
Why'd you [move my extension into a user folder(https://github.com/CS2627883/extensions/commit/2b1315b21140b159147375b6a05d25ee2bd48b5a)? Is this temporary or permanent?

GarboMuffin:
Is there any reason you don't want this to be in a user folder? The extension can still be listed on the website the same and people shouldn't be typing URLs manually anyways. Being in a user folder has the advantage for you that I'm not going to nitpick it until the end of time.

GarboMuffin:
The other encoding extension was added before we started doing user folders. If it was opened today, we would've put in a user folder.

showierdata9978:
> people shouldn't be typing URLs manually anyways.

apart from https://scratch.mit.edu/discuss/topic/653955/ lmao

CS2627883:
> Is there any reason you don't want this to be in a user folder? The extension can still be listed on the website the same and people shouldn't be typing URLs manually anyways. Being in a user folder has the advantage for you that I'm not going to nitpick it until the end of time.

I'm fine with it, and I appreciate the feedback you're giving me. I just wanted to know why.

CS2627883:
@GarboMuffin I have changed the name, id, logo, etc. to Numerical Encoding.

GarboMuffin:
I assume you are okay being credited something like this?

image

PR #179, opened 2023-02-02
New line block

TheShovel:
Does this need explanation?
image

### DIFF
--- a/extensions/utilities.js
+++ b/extensions/utilities.js
@@ -202,9 +202,9 @@
         {
           opcode: 'newline',
 
-          blockType: Scratch.BlockType.BOOLEAN,
+          blockType: Scratch.BlockType.REPORTER,
 
-          text: 'new line',
+          text: "\\n",
           arguments: {}
         },
           {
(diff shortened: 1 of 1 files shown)


softedco:
I was thinking about this too, btw you can make the text "\\n" so it's not that long for what's just a newline character

TheShovel:
It's easier to understand I think.

softedco:
> It's easier to understand I think.

Yeah but it would also look unnecessarily huge when multiple of these are used in a join nest, if you're still fine with that, then might I suggest adding a block that replaces "\\\\n" with just "\\n"? Also, why is it a boolean? I know booleans are practically reporters but better, but I can't imagine a scenario where it would be used as a condition and not just replaced with `<(true)>` or `<not<>>`.

TheShovel:
image

GarboMuffin:
call it newline, not \n, average Scratch users will see that and try to report a bug

TheShovel:
Changed it, again.

PR #185, opened 2023-02-10
Add another text to speech extension

WP-Studio01:
因为文字朗读扩展被墙，所以开发了这个文字朗读

Because the text reading extension is walled, this text reading is developed.

> Edit from maintainers: A more accurate translation may be "Because the text-to-speech extension was blocked, this text-to-speech was developed"


GarboMuffin:
Rename the extension file to something like `extensions/WP-Studio01/text2speech.js`, update the extension ID to include your username, and give it a more specific name than "tools"

Frankly it might be better for us to just use trampoline for all TTS requests, though

GarboMuffin:
https://extensions.turbowarp.org/WP-Studio01/text2speech.js

Issue #169, opened 2023-01-26
Request for comment: Proposed APIs to give users control over what websites projects can access


GarboMuffin:
# The problem

A project with eg. the fetch extension can talk to any server it wants to without confirmation. This can have privacy and security implications that are not obvious. Should the fetch extension really allow port scanning and IP logging by default? Probably not.

# Proposed solution

Part 1 is to replace every use of `fetch`, `WebSocket`, `Image`, etc. with eg. `Scratch.fetch`, `Scratch.WebSocket`, `Scratch.Image`, etc. Initially these could just be copies of their window.* counterparts. We can probably configure ESLint to error if using the old APIs.

Part 2 is to make those APIs check if the origin they are trying to load is allowed or not. We might present a prompt like this:

> This project is trying to access the website "trampoline.turbowarp.org". 
> Allowing this may share your IP address and approximate location with the website.
> [Deny]
> [Allow once]
> [Always allow trampoline.turbowarp.org]
> [Temporarily allow all websites]

Sort of like our own "google.com wants to know your location [Deny] [Allow]" prompt

Until the user chooses, fetch()'s promise will remain unresolved, WebSocket won't emit open, etc. If user denies, an error will be returned in the same way that the native fetch() would return a rejected promise, WebSocket would emit an error event, etc.

Having an option to allow forever is safe here as this is just fetching; we're not executing arbitrary JS

Depending on how fetch() handles HTTP redirects, we can likely add a builtin exception for trusted static site hosts like *.github.io where there is no privacy concern

We can do a similar thing for window opening and page redirecting:

> This project is trying to open a new tab to "github.com"
> [... the same options ...]

For these we can also add checks to ensure URLs like `javascript:alert(1)` do not run code in just one spot instead of everywhere. We can also probably detect popup blockers.

Extensions running in a sandbox will be exempted from these checks (there is nothing we can do to force them to use Scratch.fetch instead of window.fetch). The prompt to confirm loading them will be updated to mention this.

The packager will always be exempted from these checks.

## Concerns

 - These confirmations will be very annoying very quickly
 - Notification fatigue

showierdata9978:
The only problem i have with this is that users could break projects, because they have required data on another server...

This also breaks CL. As its entire use is connecting to a Cloudlink websocket server

GarboMuffin:
It doesn't "break" cloud link; the user just has to consent to connecting. Perhaps one of the default servers can be added to the list of always allowed websites.

[...]

PR #188, opened 2023-02-12
Use Scratch.fetch, Scratch.openWindow, Scratch.redirect, etc.

GarboMuffin:
https://github.com/TurboWarp/extensions/issues/169

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -76,6 +76,75 @@ module.exports = {
     // significantly impacts the behavior of projects.
     'require-await': 'error',
     // Require each extension to use strict mode
-    'strict': ['error', 'function']
+    'strict': ['error', 'function'],
+    // Disallow APIs that are replaced by Scratch.* APIs
+    'no-restricted-syntax': [
+      'error',
+      // This is not comprehensive, but it should be enough to prevent the most common ways for these to be written.
+      {
+        selector: 'CallExpression[callee.name=fetch]',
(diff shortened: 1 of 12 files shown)


showierdata9978:
Cloudlink should be able to fetch example servers without using `Scratch.fetch`. My reasoning is that, before scratch even knows about the extension, it loads it. (plus it only goes to mikes gh pages)

That URL is also static, and would only get changed if the CL repo changed, or mike added a gh pages domain (unlikely)

people are also likely to decline it, not knowing what it does. Then not being able to see any default servers (Then complaining that its broken).

GarboMuffin:
There will be a list of websites that are always exempted from the permission prompt. *.github.io and raw.githubusercontent.com will be in that list.

showierdata9978:
Ah ok.

showierdata9978:
Prob easier to maintain then having it in the ext.

But best just to ask @MikeDev101

MikeDev101:
Server list not being built directly into the extension is mainly just for convenient updating. On that note, I believe the changed fetch methods are for providing a user consent prompt of some sort?

showierdata9978:
Yeah, it opens an alert with the domain

True-Fantom:
@GarboMuffin There will be no alerts when the project is in electron format? If they are, it will be very inconvenient, considering that the user already provides full access to the device by opening `.exe` file.

GarboMuffin:
There will be no permission prompts in any packaged format

Issue #99, opened 2022-12-27
[Feature request] migrate extensions from machinelearningforkids 


Zemolu:
Any plans to migrate extensions from machinelearningforkids?
There have really cool extensions there.

https://machinelearningforkids.co.uk/#!/pretrained

Source codes:
https://github.com/IBM/taxinomitis

SIPC:
I migrated an extension from that site and solved the cross-domain problem！

OeildeLynx31:
I have an other suggestion: Can you also migrate extensions from Adacraft?
I think it is more easy and simple, because Adacraft is based on TurboWarp, and there are a lot of cool extensions, like open link, redirect, localstorage...
Thank you!

CST1229:
> I think it is more easy and simple, because Adacraft is based on TurboWarp, and there are a lot of cool extensions, like open link, redirect, localstorage... Thank you!

Additionally, [adacraft is open-source](https://gitlab.com/adacraft), unlike the mod on machinelearningforkids (AFAIK).

TheShovel:
It looks like most Adacraft extensions have dependencies that can't be loaded with TurboWarp (at least not yet).

OeildeLynx31:
> It looks like most Adacraft extensions have dependencies that can't be loaded with TurboWarp (at least not yet).

Ogadaki, the creator, said it was doable. Moreover, I can contact him to ask what changes he made exactly. But, he doesn't has a Github account. You can contact him at [@ogadaki on scratch](https://scratch.mit.edu/users/ogadaki/). If you need help, I can a little bit help you, but mainly ask @ogadaki.

BlueDome77:
> I wanna try to migrate the Spotify extension, where do I find that on the GitHub page?

https://github.com/TurboWarp/extensions/pull/1143

GarboMuffin:
Different issues can be made for specific requests, we don't have a blanket issue to transfer everything from some other website with different goals and requirements

PR #194, opened 2023-02-13
Profanity remover extension

TheShovel:
A similar extension to one of the ones mentioned in #99


GarboMuffin:
Please don't put example profanity here...

TheShovel:
Sorry, didn't think that word was that big of a deal.

TheShovel:
There are a lot of things already made that can do this, but none (that I can find) that are gonna work with just pure js. I will draft this, and if you or someone else has another idea I'll be sure to try it.

GarboMuffin:
You can absolutely do what I described with pure JS

showierdata9978:
AFAIK You should not do it client side, but this could be useful in a no "server" environment. (like if your using an example CL server)

TheShovel:
Ok, so I am trying to make it client side, and translate the array to binary, but when I translate it back, and use regex I get
`Invalid regular expression: /[/: Unterminated character class`

TheShovel:
> AFAIK You should not do it client side, but this could be useful in a no "server" environment. (like if your using an example CL server)

Yeah, but Garbo asked for it. I'll host the api version on my github, but I still wanna get this one done.

softedco:
> Ok, so I am trying to make it client side, and translate the array to binary, but when I translate it back, and use regex I get
> `Invalid regular expression: /[/: Unterminated character class`

[ is a meta character so you write it like so \\[

TheShovel:
I have made it client side, and the words are base64 encrypted.
Also added the option to change what it is replaced with.

GarboMuffin:
https://extensions.turbowarp.org/TheShovel/profanity.js

I do not know if we want this listed on the website

TheShovel:
It's a harmless extension to help people make stuff like chat apps safer for younger people. I don't see a reason why this shouldn't be on the website.

softedco:
> It's a harmless extension to help people make stuff like chat apps safer for younger people. I don't see a reason why this shouldn't be on the website.

I believe there is more of a concern on how the Scratch Team would deal with the website having this extensiom on the front page, but I also believe that it's fine since the scratch website itself has bad word filtering.

PR #205, opened 2023-02-25
JSON extensions

Skyhigh173:
json extensions, which have the following blocks:
- is json valid
- json contains key?
- array contains value?
- get all (keys/values/keys&values) from json
- length of array
- -
- get key from json
- set key from json
- delete key from json
- -
- get item n of array
- item # of string in array
- array concat (list 1) (list 2)
- replace item n at array
- insert item at array
- delete item at array
- delete every specstring in array
- array from index a to b
- reverse array
- flat array with depth
- -
- create array by string with delimiter
- join array as string
- -
- get list content as array (vm)
- set list content as array (vm)


GarboMuffin:
- add `const vm = Scratch.vm` at the top
 - move the extension to a use folder like Skyhigh173/json.js
 - change extension ID to something like skyhigh173json

Skyhigh173:
should be ok now

Skyhigh173:
I don't mind it, and I suggest you to add more features into your JSON extension!

True-Fantom:
@Skyhigh173

Thanks)

I already have the following ideas:

Array of keys from the object
Array of values from the object
Sort the array `alphabetically and so on`
Keys with the value ()

Skyhigh173:
My extension already have first 2.

True-Fantom:
@Skyhigh173 

What about this?)

Skyhigh173:
It is better to talk this in your own repo. (To JSON? Do it means it will return real JSON?)

True-Fantom:
@Skyhigh173 

It works like this:

Skyhigh173:
@GarboMuffin It has been some days since I updated it. Can you please check it again?

GarboMuffin:
This, and all of the other extensions in the queue, will be looked at this weekend

Issue #204, opened 2023-02-25
General style of extension images🎨


True-Fantom:
![clippingblending](https://user-images.githubusercontent.com/116278271/225658946-532e42b8-d042-4011-a590-cd6d4bb2a741.svg)
![box2d](https://user-images.githubusercontent.com/116278271/225648147-c75fcd36-da21-418c-96a2-de9dfe7280c0.svg)
![cameracontrols](https://user-images.githubusercontent.com/116278271/225648154-ff11a1a6-37a8-41bd-8042-cb8b7913e262.svg)
![couplers](https://user-images.githubusercontent.com/116278271/225648159-fec2387d-2350-4e6f-9766-7817d6578ebf.svg)
![cursor](https://user-images.githubusercontent.com/116278271/225648164-d4c6d5f6-41a8-41d3-b8bc-c25cf78ba925.svg)
![custom](https://user-images.githubusercontent.com/116278271/225648166-ddcdf2c5-5eac-48d4-af1c-086dda036851.svg)
![files](https://user-images.githubusercontent.com/116278271/225648170-f5687ff3-b18d-4bfa-94a0-059c082e878e.svg)
![gallery](https://user-images.githubusercontent.com/116278271/225648175-9ca7abe8-2c54-4219-aead-ba5ae50cfa18.svg)
![gamepad](https://user-images.githubusercontent.com/116278271/225648177-7db5d27a-5d37-4f8b-b287-90a4d33e3a0f.svg)
![math](https://user-images.githubusercontent.com/116278271/225648180-9acadd8a-628d-4464-b0be-014e9e2e7428.svg)
![pointerlock](https://user-images.githubusercontent.com/116278271/225648183-a9c4916e-781d-4973-ae2f-f113450c1be9.svg)
![stretch](https://user-images.githubusercontent.com/116278271/225648186-25041c05-9766-4879-82e8-b12489145f67.svg)
![tw](https://user-images.githubusercontent.com/116278271/225648187-55c7c1a9-e483-4f9e-9c39-a400670b3f86.svg)

GarboMuffin:
these are very intruging

GarboMuffin:
I really like these -- we're going to start using them

Are there any parts of these that you did not make that we should be crediting?

True-Fantom:
@GarboMuffin Should I specify images with license only? Or all the ones I used?

[...]

PR #207, opened 2023-02-26
Camera controls extension

DT-is-not-available:
Allows free movement of the camera along the X and Y axis, and zooming. Note that this extension automatically removes fencing to make this work correctly.
Adds the following blocks:
image
I currently do not have an image for the extension, I will make one sometime tomorrow.

### DIFF
--- a/extensions/DT/cameracontrols.js
+++ b/extensions/DT/cameracontrols.js
@@ -1,0 +1,225 @@
+(function(Scratch) {
+  
+  'use strict';
+  if (!Scratch.extensions.unsandboxed) {
+	throw new Error("AR extension must be run unsandboxed");
+  }
+
+  const vm = Scratch.vm;
+
+  let cameraX = 0;
+  let cameraY = 0;
+  let cameraZoom = 100;
+  let cameraBG = '#ffffff';
+
+  vm.runtime.runtimeOptions.fencing = false;
+  vm.renderer.offscreenTouching = true;
+
+  function updateCamera() {
+    vm.renderer.setStageSize(
+      vm.runtime.stageWidth/-2+cameraX, 
+      vm.runtime.stageWidth/2+cameraX, 
+      vm.runtime.stageHeight/-2+cameraY, 
+      vm.runtime.stageHeight/2+cameraY
+    );
+    vm.renderer._projection[15] = 100/cameraZoom;
+    /*
+    vm.renderer._backgroundColor4f = [
+      parseInt(cameraBG.substring(1,3),16)/255,
+      parseInt(cameraBG.substring(3,5),16)/255,
+      parseInt(cameraBG.substring(5,7),16)/255,
+      1
+    ]
+    */
+  }
+
+  // tell resize to update camera as well
+  vm.runtime.on("STAGE_SIZE_CHANGED", _=>updateCamera());
+
+  function doFix() {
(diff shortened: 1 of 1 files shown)


TheShovel:
This is something I wanted to see for a while now. This is cool.

True-Fantom:
@DT-is-not-available

Hello! I have slightly changed your extension so that it is in scratch style:



And also I added icon and created new image (#204)


[Here](https://github.com/True-Fantom/true-fantom.github.io/blob/main/tests/camera.js) is the finished code, just copy it

DNin01:
Would this coincide with Box2D?

True-Fantom:
@DNin01 I looked at the code, apparently yes, but it's not accurate

DT-is-not-available:
@True-Fantom 
Thanks! I will update the images and credit you later today. As for the palette layout, I had no idea you could just insert "---" as a separator, very useful.

@DNin01
I have no idea, I'll have to check but I imagine it wouldn't work perfectly if at all.

True-Fantom:
@DT-is-not-available Was happy to help)

DT-is-not-available:
I've gone ahead and added the images to the extension, as well as the modified block palette. I do have an unused "set background color" block, since it doesn't change the result of the touching-color boolean. Does anyone know how to do that? Also at some point I want to add camera rotation, but I'm not sure how to do that at this point, I'll see if I can tell webgl to apply proper transformations before drawing anything rather than messing with the projection values.

GarboMuffin:
It helps to post the code that doesn't work when asking for further help

GarboMuffin:
Ok it is in there, just a bit indirect

Look at vm.renderer.setBackgroundColor

DT-is-not-available:
Oh I see now, should look through scratch-render some more, thanks!

True-Fantom:
@DT-is-not-available Since you added a background color, I think need to add reporter for it:

+position the block a little higher) [Here](https://github.com/True-Fantom/true-fantom.github.io/blob/main/tests/camera.js) is code

DT-is-not-available:
Background color feels unrelated to camera movement, so I'll probably move it and a reporter under the movement blocks. (I may add more visual blocks since proper transformations requires custom shaders, so I might as well use them)

True-Fantom:
Ok, that's a great idea

DT-is-not-available:
I will modify the layout of the extension tonight, and probably leave it as is for a while until I learn more about how webgl works. On that note, is there a way I should go about applying custom glsl code as a vertex shader without screwing up the default?

GarboMuffin:
Why does this extension have unused code related to shaders?

ghbbeep:
I'm assuming they're not talking about this extension.

GarboMuffin:
The extension in this PR appears to contain a lot of code and shaders that are seemingly unused. Is that intentional?

DT-is-not-available:
@GarboMuffin That was not intentional, my bad. I am considering rewriting the extension to use a custom shader however, as this would allow for much easier implementation of things like camera rotation, and well, camera "shader" effects. I've removed the shader related code and updated the set background color block to be at the end of the block palette, along with a background color reporter.

DT-is-not-available:
Alright, I'm done updating this for a while.

True-Fantom:
@GarboMuffin Can you merge this, please?

Issue #177, opened 2023-01-31
Additional graphic effects


DNin01:
It might be cool if we had an extension that added more graphic effect options like color-invert, blur, and drop-shadow.

True-Fantom:
@DNin01 I think we should also add "color opacity", "color saturation" and "color brightness" effects

Because:

![Снимок экрана (588)](https://user-images.githubusercontent.com/116278271/222183287-69dab2d8-fb0b-435a-a0de-4403a958afcc.png)
![Снимок экрана (586)](https://user-images.githubusercontent.com/116278271/222183552-864bd4c6-54a3-4ba2-aacc-c9c0ca7fe93f.png)

![Снимок экрана (591)](https://user-images.githubusercontent.com/116278271/222183311-b7a534fd-f4ed-4557-bf9e-d695d263ece2.png)

![Снимок экрана (592)](https://user-images.githubusercontent.com/116278271/222185439-77cea999-3144-44fe-94e7-d99de1356acc.png)

PS: color effect works fine, but you need to multiply the number by 2 for it:

![Снимок экрана (590)](https://user-images.githubusercontent.com/116278271/222186156-381c2681-6e0c-4112-8467-383294b9d83c.png)

PPS: With saturation -100 effect, you can achieve monochrome effect :)

True-Fantom:
It would also be cool to see "color coarsening" and "color threshold" effects:
![0553](https://user-images.githubusercontent.com/116278271/222188447-3a55aff6-0238-444a-83f0-8a64cb731d48.jpg)
![0554](https://user-images.githubusercontent.com/116278271/222188441-7c5d32bc-d2ca-4bee-bfee-4a0e198207ec.jpg)

TheShovel:
#299 is adding this, for the entire stage.

PR #210, opened 2023-02-27
Add Effects extension

Xeltalliv:
image

Example of using those blocks:
image

Adds 2 effects:
- Clipping, which only shows parts of sprite inside of the clipping area.
- Additive blending

Those effects are cleared after stopping the project. They affect stamps, but don't affect pen. Touching blocks work as if those effects don't exist. They are disabled for stage.


True-Fantom:
> It might need a better name.

Perhaps just "effects" would have been better. Or "shaders"

True-Fantom:
It may be worth adding these blocks:

- `(clipping box x1)`
- `(clipping box y1)`
- `(clipping box x2)`
- `(clipping box y2)`
- `<additive blending on?>`

True-Fantom:
image prototype (#204)

True-Fantom:
@Xeltalliv By the way, maybe add #177 effects in the future?

Xeltalliv:
> @Xeltalliv By the way, maybe add #177 effects in the future?

In terms of code, it makes more sense to make that a separate extension. I might eventually make it. Or someone else might make it. And for this reason, calling this extension "Effects" may be slightly problematic.

True-Fantom:
@Xeltalliv Well, in fact, all these are the effects of displaying sprites. It would not be very convenient to divide them into "color effects" and "something effects". Perhaps there is an alternative name "Rendering" or "Render".

True-Fantom:
@Xeltalliv 
Скриншот 01-03-2023 194644

True-Fantom:
@Xeltalliv I edited the code a bit and added an icon. [Here](https://github.com/True-Fantom/true-fantom.github.io/blob/main/tests/render.js) is code

PS: I changed name to "Render"

True-Fantom:
> Also, there's no need to mention me

Ok

I changed name back

True-Fantom:
Do you agree with changes?

Xeltalliv:
> Do you agree with changes?

> * added icon

Yes
> * renamed class from `Extension` to `Effects`

Keeping class named just `Extension` allows changing name of the extension more easily. I've recently switched to doing that in every new extension I make.

> * renamed `min x`, `min y`, `max x` and `max y` to `x1`, `y1`, `x2` and `y2` for convenience

There is a reason why it was called that way.
People might not expect this:
image
But this is immediately clear:
image

> * replaced `(function(Scratch) {` with `(Scratch => {`
> * removed unnecessary empty lines
> * add some empty lines

It was formatted in the way I prefer. Double/more empty lines are used to more visibly separate distinct sections of code. [Trailing commas](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas) in "blocks" and "menus" were also intentional.

TheShovel:
Maybe rename this extension to Clipping if you are not planning on adding other effects.

True-Fantom:
Okay)

I asked and offered everything I wanted. You made excellent extension! Good luck in the future)

Xeltalliv:
Note to GarboMuffin: this extension modifies gl state, but doesn't fully restore it at the end. Despite that it seems to work. Not sure if it's better to keep it as is, or make it intercept, save and then restore everything which will add extra overhead, but potentially improve compatibility.

GarboMuffin:
This seems like another case where a documentation page could be beneficial

I'll fix the merge conflicts and merge this in a bit, such a documentation page can be added later if you want to

WojtekCodesToday:
ok

PR #214, opened 2023-03-01
More Comparisons

NOname-awa:
More Comparisons


True-Fantom:
This will largely not correspond to the logic of Scratch

True-Fantom:
Скриншот 03-03-2023 171913

NOname-awa:
"^" Is XOR, when the Boolean values are not equal to true,
B(XB6M3UG96Q3W~Y 7NR830
The second is to convert Boolean value to numbers (1 and 0) Although the result is not true and false, 
but they can be used in "if"
the first if it is useless，it can be deleted

NOname-awa:
"^" Is XOR
B(XB6M3UG96Q3W~Y 7NR830
The second is to convert Boolean value to numbers (1 and 0) Although the result is not true and false,
but they can be used in "if"
the first if it is useless，it can be deleted

True-Fantom:
> "^" Is XOR

Yes, I understand. But it doesn't work properly. It use `Boolean()`, and scratch uses its own conversion method:

    static toBoolean (value) {
        // Already a boolean?
        if (typeof value === 'boolean') {
            return value;
        }
        if (typeof value === 'string') {
            // These specific strings are treated as false in Scratch.
            if ((value === '') ||
                (value === '0') ||
                (value.toLowerCase() === 'false')) {
                return false;
            }
            // All other strings treated as true.
            return true;
        }
        // Coerce other values and numbers.
        return Boolean(value);
    }

NOname-awa:
> > 第二种是将布尔值转换为数字（1 和 0）
> 
> 可更换 Скриншот 03-03-2023 203002

OK, thanks！

True-Fantom:
You are welcome)

I would advise you to look https://github.com/TurboWarp/scratch-vm/blob/develop/src/util/cast.js

NOname-awa:
so after this modification, can you join "Turbowarp Extension Gallery"...
(My English is not very good, there may be wrong words)

NOname-awa:
> You are welcome)
> 
> I would advise you to look https://github.com/TurboWarp/scratch-vm/blob/develop/src/util/cast.js

So, after modification, can it add "Turbowarp Extension Gallery"?
(My English is not very good, there may be wrong words)

True-Fantom:
Yes! But I would suggest that you fix all the errors first

True-Fantom:
for example this:
Скриншот 03-03-2023 205710

True-Fantom:
And I think we need to add more unique blocks, because most of them have analogues.
Скриншот 03-03-2023 212307

NOname-awa:
> 我认为我们需要添加更多独特的块，因为它们中的大多数都有类似物。 Скриншот 03-03-2023 212307

ok

NOname-awa:
> And I think we need to add more unique blocks, because most of them have analogues. Скриншот 03-03-2023 212307
358~~4OFY5CIZ8EQ(}2C{TV
-images.githubusercontent.com/126544888/222806042-39072e11-3990-4f7f-8821-a66e93b09e51.png)
3~L57X5~7%2K_JXIQTWNH}S
OK
(Blue is another extension)

NOname-awa:
> And I think we need to add more unique blocks, because most of them have analogues. Скриншот 03-03-2023 212307
358~~4OFY5CIZ8EQ(}2C{TV
(DXB%1V$W0{@1PQMSKOV9}W
3~L57X5~7%2K_JXIQTWNH}S
OK（Blue is another extension）

NOname-awa:
> And I think we need to add more unique blocks, because most of them have analogues. Скриншот 03-03-2023 212307

what else needs to be changed

GarboMuffin:
- move the extension file into a user folder, such as `NOname-awa`
 - change extension ID to something that includes part of the user folder, such as `NOname-awacomparisons`
 - wrap it in the IIFE syntax as described in https://docs.turbowarp.org/development/unsandboxed-extensions
 - see if some of your code is made redundant by `Scratch.Cast.compare(a, b)` (less than 0 means a < b, 0 means a == b, greater than 0 means a > b)

PR #252, opened 2023-03-09
CameraControls: add camera rotate feature

LoserFox:


LoserFox:
> @DT-is-not-available 

yeah,original extension exist this question. I not really sure how to fixing it,If you know the spec or some manual that explains it, please to send it

True-Fantom:
I think it should look like this:
Скриншот 09-03-2023 204547
(fast photoshop)

LoserFox:
> here are the icons for turns: rotate-left2 rotate-right2

How do I use these icons? I don’t know how to insert icons in text, sorry.

True-Fantom:
@LoserFox https://docs.turbowarp.org/development/extensions/assorted-apis (towards the end)

Скриншот 09-03-2023 213142

True-Fantom:
And we could also implement these blocks as well:
Скриншот 09-03-2023 212248

DT-is-not-available:
Unless theres a way I'm unaware of, I'm pretty sure I would need a custom vertex shader for that. I'll look into this.

DT-is-not-available:
I completely missed that this was a PR, not an issue. The camera X is already situated with Scratch's coordinate system and does not need to be inverted (doing so makes it look like the camera is going backwards). I would still like to eventually implement this as a vertex shader, as doing so would increase performance (no need to 'resize' the stage each time, it is currently not very performant under 60fps and/or turbo mode).

Xeltalliv:
> I would still like to eventually implement this as a vertex shader, as doing so would increase performance (no need to 'resize' the stage each time, it is currently not very performant under 60fps and/or turbo mode).

I'm not sure, but you can try doing something like this (which doesn't use resizing and seems to work):
```
  function camera(x,y,scale,rot) {
    rot = rot / 180 * Math.PI;
    let s = Math.sin(rot) * scale;
    let c = Math.cos(rot) * scale;
    vm.renderer._projection = [        c/240,       -s/180, 0,0,
                                       s/240,        c/180, 0,0,
                                           0,            0,-1,0,
                               (c*x+s*y)/240,(c*y-s*x)/180, 0,1];
    vm.renderer.dirty = true;
  }
```

DT-is-not-available:
That works! I had to modify the function (invert the X axis and replace `240` and `180` with respective stage coordinates) but that seems to do the trick, and it runs instantly. I've added it to the code and implemented camera rotation.

LoserFox:
> I completely missed that this was a PR, not an issue. A few things:
> 
> 1. The camera X is already situated with Scratch's coordinate system and does not need to be inverted (doing so makes it look like the camera is going backwards).
> 2. When the camera rotates, it should rotate around the camera's current position, not the world center, as it currently is.
> 
> I would still like to eventually implement this as a vertex shader, as doing so would increase performance (no need to 'resize' the stage each time, it is currently not very performant under 60fps and/or turbo mode).

i think 'resize' the stage each time is meanless,why you to do this?when I deleted it during my own testing, no strange problems occurred, I don’t quite understand.

DT-is-not-available:
It is now, but at the time of implementing it, it was required since I was not using the projection to move the camera, only to zoom. I'll make a new PR with the updated version soon which uses the method previously stated, as adding an entire library to the source seems pointless for something one function can achieve.Sent from my Galaxy

LoserFox:
I feel a bit lost, but that's ok, I'll try to make a few extensions of my own

PR #254, opened 2023-03-09
ShovelUtils: Fix several bugs, including some that can corrupt projects

TheShovel:


TheShovel:
Might as well add the banner now too.

Xeltalliv:
Suggestions:
`Set target to` should do nothing instead of throwing when sprite it was meant to switch to doesn't exist.
`Get list` should return empty string instead of throwing when list is not found.
`Set list` should do nothing instead of throwing when list is not found.
`Set list` should do nothing string instead of throwing when json is invalid.
`Set list` should do nothing instead of **crashing scratch-gui** when json represents something other than 1-dimensional array of strings, numbers or booleans (and list is shown).
`Get brightness` should return empty string instead of undefined, when value is invalid.
`Import image` shouldn't do anything instead of importing gray question mark costume when download fails.
`Import sound` shouldn't do anything instead of importing silent 0 second sound when download fails.
`Import project` shoul do nothing instead of after 15 seconds **breaking the editor** when download fails (especially considering that it fails with default value)

TheShovel:
I will fix all of those instead of  `Import image` because you should see that it failed to fetch.

TheShovel:
Also `brightnessByColor` will stay the same, because what it returns makes sense.

TheShovel:
`importProject` can't be handled correctly, so sadly I can't fix that.

Xeltalliv:
> `Set list` should do nothing instead of **crashing scratch-gui** when json represents something other than 1-dimensional array of strings, numbers or booleans (and list is shown).

This can be fixed by checking parsed result in `Array.isArray(valueToCheck)`, and then looping though each element of array and using `typeof` to check each element.


> Also `brightnessByColor` will stay the same, because what it returns makes sense.

In scratch reporter blocks should never return undefined, null, arrays, objects, etc. The only valid [values scratch can work with](https://docs.turbowarp.org/development/extensions/unsandboxed#increased-power-brings-increased-responsibility) are numbers, strings and booleans.

TheShovel:
Set list is already fixed. Gonna also change the brightness block tomorrow I guess.

Xeltalliv:
> `Set list` should do nothing instead of throwing when list is not found.
> `Set list` should do nothing string instead of throwing when json is invalid.

Those 2 are fixed.

> `Set list` should do nothing instead of **crashing scratch-gui** when json represents something other than 1-dimensional array of strings, numbers or booleans (and list is shown).

This isn't because the block itself sets the value successfully without throwing errors, but after the list is shown, since the value is invalid, it crashes.
Basically this (crash trying to read the length of (not) a list):
image
And this (crash trying to display one of the list values):
image

TheShovel:
Oh, right. Gonna just check if it's valid json to fix the issue in the first image. Idk how I would go about the 2nd one tho.

Xeltalliv:
> Gonna just check if it's valid json to fix the issue in the first image.

123 from the first image is a valid json.
To fix the issues from the first image, after parsing the value with `JSON.parse`, put result into a variable. Then check that variable using `Array.isArray(valueToCheck)`. If it returns false, then the value was wrong and the block should do nothing, so `return` from it.
If it returned true, then it is a valid array, so the code can proceed. The next thing to do is to loop through every element and check it's value using `typeof(valueToCheck)`. If it returns something other than `"string"`, `"number"` or `"boolean"` then the value wasn't valid and may cause the what is shown on the second image, so once again, the block shouldn't do anything, `return` from it.
If looping finishes, it means that nothing invalid was found and it is safe to set the list.

GarboMuffin:
I will probably delist this extension from the homepage until the bugs that reportedly break scratch-gui or the editor are addressed

TheShovel:
@Xeltalliv All issues you mentioned are fixed. You can check if there is anything left, because everything seems stable to me now. The sound and image import blocks will still return default stuff, as that is intended.

extensions/TheShovel/ShovelUtils.js
```diff
@@ -216,15 +216,30 @@
...
     setedtarget({ NAME }) {
-      vm.setEditingTarget(vm.runtime.getSpriteTargetByName(NAME).id);
+      //I know this might cause sprites called "stage" to be ignored. But lets be real, who names their sprite "stage"?
+      try{if (NAME.toLowerCase() === "stage"){ 
+        vm.setEditingTarget(vm.runtime.getTargetForStage().id);}
+      else
+      {vm.setEditingTarget(vm.runtime.getSpriteTargetByName(NAME).id);}}
+      catch(error){
+        return ""
+      }
```
Xeltalliv, line 242:
```suggestion
      let target;

      //I know this might cause sprites called "stage" to be ignored. But lets be real, who names their sprite "stage"?
      if (NAME.toLowerCase() === "stage") { 
        target = vm.runtime.getTargetForStage();
      } else {
        target = vm.runtime.getSpriteTargetByName(NAME);
      }
      if (target) {
        vm.setEditingTarget(target.id);
      }
```

Xeltalliv:
The problem from the second image isn't fixed.

TheShovel:
> The issue from the second image still isn't fixed.
> 
> Also, your use of [this code](https://gist.github.com/w3core/e3d9b5b6d69a3ba8671cc84714cca8a4) may be problematic, since there doesn't seem to be any license information about it.

When I tested it, the issue in the 2nd image was fixed. I typed the same thing you did, and it worked as intended. Out of curiosity, how did you trigger the issue now?

Xeltalliv:
I created a list, showed it, and then activated the block to set it to `[{"toString":1}]`. First activating the block and then showing the list also worked.

PR #263, opened 2023-03-11
Xeltalliv/clippingblending:add custom blend

LoserFox:
let the user itself control the webgl blend mode (also keep additiveBlend)

### DIFF
--- a/extensions/Xeltalliv/clippingblending.js
+++ b/extensions/Xeltalliv/clippingblending.js
@@ -30,7 +30,20 @@
   let height = 0;
   let scratchUnitWidth = 480;
   let scratchUnitHeight = 360;
-
+  
+  var blendMode = {
+     ZERO: 0,
+     ONE: 1,
+     SRC_COLOR: 0x0300,
+     ONE_MINUS_SRC_COLOR: 0x0301,
+     SRC_ALPHA: 0x0302,
+     ONE_MINUS_SRC_ALPHA: 0x0303,
+     DST_ALPHA: 0x0304,
+     ONE_MINUS_DST_ALPHA: 0x0305,
+     DST_COLOR: 0x0306,
+     ONE_MINUS_DST_COLOR: 0x0307,
+     SRC_ALPHA_SATURATE: 0x0308,
+  }
 
   renderer._drawThese = function (drawables, drawMode, projection, opts) {
     active = true;
@@ -91,7 +104,12 @@
       if (this.additiveBlend) {
         gl.blendFunc(gl.ONE, gl.ONE);
       } else {
-        gl.blendFunc(gl.ONE, gl.ONE_MINUS_SRC_ALPHA);
+        if (this.customBlend){
+          gl.blendFunc(this.customBlend[0], this.customBlend[1])
+        }else{
+          gl.blendFunc(gl.ONE, gl.ONE_MINUS_SRC_ALPHA);
+        }
+        
       }
     }
     return gu.call(this);
@@ -101,6 +119,9 @@
   };
   DrawableProto.updateAdditiveBlend = function (enabled) {
     this.additiveBlend = enabled;
+  };
+  DrawableProto.updateCustomBlend = function (modes) {
+    this.customBlend = modes;
   };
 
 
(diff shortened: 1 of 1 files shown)


GarboMuffin:
@Xeltalliv

Xeltalliv:
I originally tried implementing this myself, but after seeing how there isn't really anything useful other than additive blending, I decided that keeping extension simple and more user-friendly was a better option.

LoserFox:
> I originally tried implementing this myself, but after seeing how there isn't really anything useful other than additive blending, I decided that keeping extension simple and more user-friendly was a better option.

I add this is because I want user have more custom selection.pro user can use it to create his self amazing effect.for normal user, just use addivite blend is okay(also because I have a project that needs to use it to implement alpha mask)

Xeltalliv:
> (also because I have a project that needs to use it to implement alpha mask)

Is it this?
image
If so, it can be added it as `multiplicative blending`.

True-Fantom:
@Xeltalliv I agree with @LoserFox. It seems to me that users should have ALL the functionality of JavaScript and Scratch components . And although this is probably not possible, but I think we should strive for it. Otherwise, what is the point of us sitting here at all?)

LoserFox:
> > (also because I have a project that needs to use it to implement alpha mask)
> 
> Is it this combination of arguments?
> image

yet,but it's a little complex,need to use many blend modes.

Xeltalliv:
> need to use many blend modes.

what other bending modes do you need and how do you plan to use them? There may be a better way to implement them.

I'm thinking of adding a premade list of known-good blending modes and giving them names, so that everyone can easily figure out how to use them.

If we still decide to go with giving users full contol (which I don't really want, since most modes are completely useless, and the extension doesn't look well with technical terms), then [gl.blendFuncSeparate](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/blendFuncSeparate) should be used instead of [gl.blendFunc](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/blendFunc) because it takes 4 arguments instead of 2 and gives more control.

LoserFox:
I think it's time to explain what I really want to do.
I have recently made a STG game. Among them, I need to let the attacks be limited to a frame, just like this
Screenshot 2023-03-12 204501
So I searched on the Internet and found that this [article ](https://stackoverflow.com/questions/5097145/opengl-mask-with-multiple-textures)shows how to achieve this mask effect. So, yes, what I want is just a simple mask effect, not a complex BlendFunc.
However, I have no idea how to pack it, so I thought, why not give all the "blendfunc" to the user and let the user realize it independently?
For myself, I only need a mask effect,This is all.

Xeltalliv:
Sorry for not replying for 2 days.

> Among them, I need to let the attacks be limited to a frame, just like this

If I understand it correctly and depending on what exactly you need, there may be 2 other simpler ways to do it without blending:
- If attacks always need to be restricted to a rectangle, it would be a lot easier to use `set clipping box x1:() y1:() x2:() y2:()` block.
- It may be possible to make a sprite that covers the entire stage and has a transparent hole. Then all attacks could be behind that sprite and as a result only be visible through the hole.

> So I searched on the Internet and found that this [article ](https://stackoverflow.com/questions/5097145/opengl-mask-with-multiple-textures)shows how to achieve this mask effect.

It wouldn't work on scratch because it relies on destination alpha channel, which [scratch's canvas doesn't have](https://github.com/LLK/scratch-render/blob/d70132f2f3ffe4ce6a3ae06ce0bf9b60084bc9fd/src/RenderWebGL.js#L114).

> So, yes, what I want is just a simple mask effect, not a complex BlendFunc. However, I have no idea how to pack it, so I thought, why not give all the "blendfunc" to the user and let the user realize it independently? For myself, I only need a mask effect,This is all.

After experimenting with this combination (multiplication blending),
image
image
image
I would like to add it, but instead of giving full access to blendFunc it can be what is shown on the image below.
Existing additive blending blocks can be deprecated and hidden from the block palette.
image

GarboMuffin:
The mock ups in the previous comment look acceptable

GarboMuffin:
@LoserFox: for your undertale game, your best option is probably to use on of the ideas proposed by Xeltalliv:

> If attacks always need to be restricted to a rectangle, it would be a lot easier to use set clipping box x1:() y1:() x2:() y2:() block.

or

> It may be possible to make a sprite that covers the entire stage and has a transparent hole. Then all attacks could be behind that sprite and as a result only be visible through the hole.

if these are insufficient then please open an issue so solutions can be discussed

@Xeltalliv: you are welcome to submit a PR adding the blocks in that mock up if you wish

My understanding is that this PR is currently stalled and unlikely to be merged in its current form as true custom blending is not useful in Scratch -- if that's wrong then let me know and I will reopen it

PR #273, opened 2023-03-13
Add fps block

TheShovel:
Code from https://www.growingwiththeweb.com/2017/12/fast-simple-js-fps-counter.html


GarboMuffin:
Would Scratch users generally expect "fps" to refer to the FPS of the entire page or just the FPS of the project? I'd expect the latter

TheShovel:
I have no idea how to do that.

PR #281, opened 2023-03-18
add graphics 2D

NOname-awa:


GarboMuffin:
see eslint warnings here: https://github.com/TurboWarp/extensions/pull/281/files

try to resolve those; they're warnings but with the new setup (implemented after you opened this), warnings will cause pull request checks to fail

GarboMuffin:
you should rename some of these to be more clear to the average user:

 - it's not obvious that "line section" means length
 - it's not obvious that "square" actually refers to "area"
 - it's not obvious that "round" refers to a circle with a given radius or diameter

PR #283, opened 2023-03-19
Extension images are now GPL 3.0 only

GarboMuffin:
and also other README updates


CST1229:
Was this caused by the text extension's image?

GarboMuffin:
It was a combination of:

 - I remembered that the text image was supposed to be GPL 3.0, so I had to go include that license text anyways
 - I decided to make all of the images GPL 3.0 as our general policy is to use permissive licenses for the project runner (scratch-vm, scratch-render, extension code) but copyleft for the editor (scratch-gui, scratch-paint, these images)

PR #284, opened 2023-03-19
Fix #268

### LINKED ISSUE #268: [JSON extension] bug with get list and set list
get list and set list can not be saved properly.

7E6F5349-511B-48D0-AE32-E9D47F717811
(get list (NOTES v) become (get list {…}) and can not work)

GarboMuffin:


GarboMuffin:
Normally these would be considered backwards-incompatible changes, but the extension was broken anyways.

PR #286, opened 2023-03-19
Simplify mouse cursor menu

Samq64:
Simplifies the mouse cursor menu by removing the labels and just using an array. I tested for backwards compatibility.

Also, if I were  to submit more extensions, would you want this one moved to my personal folder?


GarboMuffin:
> Also, if I were to submit more extensions, would you want this one moved to my personal folder?

No, we're leaving existing extensions in the same place. Just new extensions in user folders.

PR #304, opened 2023-03-25
Camera controls: bugfixes and move x steps

DT-is-not-available:
Fixed accidentally reversing the y-axis in the last pull request because I'm too used to down being positive, rearranged the block palette to more closely match the movement blocks, and added the move camera x steps block.


GarboMuffin:
seems like this would technically be a backwards-incompatible change

DT-is-not-available:
I'm pretty sure I reversed the Y axis in the last update when I implemented the new camera movement function, so I was considering this a fix to a backwards-incompatible change.

DT-is-not-available:
If I rename the current blocks dealing with camera Y to prepend negative, and make new blocks that are scratch coordinate accurate (as they are now), would that then be backwards-compatable?

gomiigg:
imagine the ability to have multiple (maybe just up to 4) cameras at once, and having certain sprites locked to certain cameras. like having a camHUD, camGame, etc. also i have no idea where I am or what im doing here or how github works i just thought it was a cool idea

gomiigg:
> imagine the ability to have multiple (maybe just up to 4) cameras at once, and having certain sprites locked to certain cameras. like having a camHUD, camGame, etc. also i have no idea where I am or what im doing here or how github works i just thought it was a cool idea

ok nevermind instead of doing all that it could just be done using a "lock to camera at (x),(y)" block or something similar

PR #314, opened 2023-03-28
Add a new extension Dark?

0832k12:
A block that monitors whether the current browser is in dark mode.


LukeManiaStudioss:
This should be added to navigator, I'm not sure if an entirely new extension just for 1 block is necessary.

PR #316, opened 2023-03-29
more handle string and math

NOname-awa:
Note: it is not the same as true-fantom's math extension, which focuses more on string handling


Skyhigh173:
I think these function/blocks already exist in other extensions

GarboMuffin:
We can certainly add this as a hidden thing if you want, but to put this on the homepage, which blocks in here are only in this extension?

PR #317, opened 2023-03-30
Regular Expression.js

NOname-awa:
Regular Expression


GarboMuffin:
even people who are extremely familiar with regular expressions are not going to be able to understand what "set () () ()" means and how to use these blocks

NOname-awa:
i modified this extension

PR #319, opened 2023-04-01
Add 0832/rxFS2

0832k12:
I eliminate the bug of "searching for files, listing files" and beautify the blocks.


GarboMuffin:
Thank you for your patience

These changes look to be backwards incompatible. So we can either:

 - Keep all of the old blocks around forever, but hide them from the block list and add new ones to replace the old ones
 - Add an "rxfs2" extension to replace the old one on the homepage and not touch the old one

0832k12:
I choose the first one
"Keep all of the old blocks around forever, but hide them from the block list and add new ones to replace the old ones"

GarboMuffin:
I suppose there is a third solution, we probably can consider a repository of not open source/proprietary extensions and move this one there

0832k12:
Can I make you a **“共同 开发者”**? This way, you can update the extension.
please E-mail to "0832k12@gmail.com to registration.

TheShovel:
I actually wasn't aware of this and now I have to remove this extension from one of my projects. This is a very sketchy practice from  0832k12 and I don't think I'll trust using their code in the future.

0832k12:
Sorry, I feel guilty for my mistake and the impact it has had on your work. I will not use this license again in the future.
@GarboMuffin @TheShovel

GarboMuffin:
"directly compatible with MIT license" is good enough for us; we will use it under that license. thank you, will get this into the next batch of merges

PR #339, opened 2023-04-14
Update Consoles.js

SIPC:


extensions/-SIPC-/consoles.js
```diff
@@ -6,18 +6,18 @@
         constructor () {}
         getInfo() {
             return {
-                id: 'sipcconsole',
+                id: 'sipcconsoles',
```
GarboMuffin, line 9: can't change this

GarboMuffin: all of your other changes seem backwards-compatible so just don't touch the extension id

SIPC: Oh! I wasn't aware that ’s‘ was added here

PR #341, opened 2023-04-15
Create global-coordinate.js

NOname-awa:


TheShovel:
When the position of the screen is changed, rotation doesn't work correctly. It shouldn't rotate around the middle of the canvas, it should rotate around the middle of the screen, using it's coordinates.

TheShovel:
[video.webm](https://user-images.githubusercontent.com/68913917/232532062-51260388-909a-489b-b27f-bbbd97fc3f18.webm)

extensions/NOname-awa/global-coordinate.js
```diff
@@ -0,0 +1,484 @@
...
+			}
+		}
+		Set_si({ s, screen }) {
+			if (!isNaN(s)) {
+				gs[screen - 1] = s;
+			}
+			else {
+				gs[screen - 1] = 0;
+			}
+			if (gs < 0) {
```
GarboMuffin, line 455: This looks wrong
```
Operator '<' cannot be applied to types 'number[]' and 'number'.ts(2365)
```

gs is an array; why is it being compared to a number? are you sure this does what you want?

GarboMuffin:
bumping  per above comment

PR #351, opened 2023-04-22
New extension: Images

CST1229:
Updated port of the plugin-based https://cst1229.github.io/files/extensions/unsandboxed-images.js. 
Lets you display images from a URL, kinda like Pen+ but costume-based.
(Sometimes called Unsandboxed Images.)

### DIFF
--- a/extensions/CST1229/images.js
+++ b/extensions/CST1229/images.js
@@ -110,6 +110,13 @@
             },
           },
           {
+            opcode: "imageID",
+            blockType: Scratch.BlockType.REPORTER,
+            text: "current image ID",
+            arguments: {},
+            disableMonitor: true,
+          },
+          {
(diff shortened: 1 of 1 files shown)


extensions/CST1229/images.js
```diff
@@ -0,0 +1,382 @@
...
+      this.render = vm.runtime.renderer;
+      [];
+
+      this.createdImages = new Set();
+      this.validImages = new Set();
+    }
+
+    getInfo() {
+      return {
+        id: "images",
```
GarboMuffin, line 26: for any observers, normally such an extension ID would be an issue, but this is port of an existing extension

CST1229: i also completely forgot about that

PR #352, opened 2023-04-22
data analysis extension

qxsck:
An extension of data analysis, including mean, minimum, maximum, median, mode, variance


GarboMuffin:
* format per https://docs.turbowarp.org/development/extensions/unsandboxed#syntax
 * consider using `Scratch.Cast.toNumber` instead of `Number`
 * consider what happens when these blocks are run with empty strings as the input. They should be set up to always return eg. 0 instead of things like `null`

PR #353, opened 2023-04-22
Add JavaScript

TylerMS887:
Adds the JavaScript extension. The purpose of it is to run JavaScript code.


TheShovel:
If I'm not wrong, GarboMuffin declined other extensions that do this because it can be a "security risk" or something like that.

TylerMS887:
Thanks. It can be used by evil hackers to create malware with extensions.
We might even create a new website for *sandboxed* extensions.

PR #355, opened 2023-04-23
OpenAI Integrations - Sandbox Compatible

kindpump:
User can use OpenAI chat completion models with their private API Key to receive AI responses


True-Fantom:
I don't think putting a link to your profile on an image is a good idea. It not only looks not nice, but also useless, since there is a description of extensions.

kindpump:
word i kinda threw the banner together last second I'll change it in a bit> I don't think putting a link to your profile on an image is a good idea. It not only looks not nice, but also useless, since there is a description of extensions.

Scratchfangs:
You copied what they said word for word 😅

GarboMuffin:
The extension needs to make it more clear that anyone who has your API key can make requests on your behalf, possibly incurring a charge on someone's credit card

kindpump:
Good idea @GarboMuffin, should that warning be a part of custom documentation or somewhere in the extension itself?

True-Fantom:
@kindpump Will you still be working on the extension? If not, I can take the development into my own hands.

kindpump:
I've been playing with it but haven't really been able to do anything I wanted to: (use other models like chat response model or dalle2 text to image model). If you want to take over development that'd be great since you probably actually know what you're doing and not just trial and error-ing like I am haha

kindpump:
@True-Fantom if you can figure out how to access dalle2 image generation that'd be amazing I've spent hours trying to figure out how to grab the images and load as dataURI's etc

True-Fantom:
@kindpump Well I'll try to do it in a while. Don't worry, almost everyone here does everything by trial and error :)

kindpump:
@True-Fantom haha well at least I'm not the only one. Hope you're able to get any of it to work. I'm currently working on a project with it but without the image generation its just another gpt api direct

Martinelplayz:
This needs to be added soon

GarboMuffin:
we will close this in favor of one of the other openai extensions submitted

Issue #302, opened 2023-03-24
Clipping And Blending Extension Doesn't Work With Pen


Mistium:
Please make this extension work with pen it would be so useful

Not much else to say, Thanks in advance

btw nobody needs to actually do this asap just get to it when you get to it lol

PR #369, opened 2023-04-29
Clipping and blending: blend modes and pen support

Xeltalliv:
This PR:
* Deprecates and hides additive blending blocks
* Adds blocks for working with blend modes (https://github.com/TurboWarp/extensions/pull/263)
image
* Makes this extension affect pen (https://github.com/TurboWarp/extensions/issues/302)
image

### DIFF
--- a/extensions/Xeltalliv/clippingblending.js
+++ b/extensions/Xeltalliv/clippingblending.js
@@ -32,6 +32,7 @@
   let height = 0;
   let scratchUnitWidth = 480;
   let scratchUnitHeight = 360;
+  let penDirty = false;
 
 
   renderer._drawThese = function (drawables, drawMode, projection, opts) {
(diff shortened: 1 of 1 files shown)


LukeManiaStudioss:
This looks incredible!!!

gomiigg:
how do i load this onto the editor? i know that the raw file url shouldn't work, but the url for the main file doesn't work either so which link do i paste into the custom extensions box?

CST1229:
> how do i load this onto the editor? i know that the raw file url shouldn't work, but the url for the main file doesn't work either so which link do i paste into the custom extensions box?

Neither, as only files on extensions.turbowarp.org and localhost:8000 are loaded unsandboxed. You'll have to wait for this PR to be merged. (However, on the packager, all extensions are loaded unsandboxed.)

PR #371, opened 2023-04-30
Add S-Grab Extension

NexusKitten:
S-Grab is a minimalistic extension powered by ScratchDB to provide information from the Scratch Website. Made by NexusKitten (known as NamelessCat on Scratch).
image (19)


extensions/NexusKitten/sgrab.js
```diff
@@ -0,0 +1,211 @@
...
+        };
+      }
+      async usergrab(args) {
+        const response = await Scratch.fetch('https://scratchdb.lefty.one/v3/user/info/'+args.WHO);
+        const jsonData = await response.json();
+        if (args.WHAT === 'follower') {
+          return jsonData.statistics.followers;
+        } else if (args.WHAT === 'following') {
+          return jsonData.statistics.following;
+        }
```
GarboMuffin, line 149: if args.WHAT is something you don't recognize, all of the blocks still needs to return *something*, probably just an empty string

True-Fantom:
I don't think putting your profile name on an image is a good idea. It not only looks not nice, but also useless, since there is a description of extensions.

extensions/NexusKitten/sgrab.js
```diff
@@ -0,0 +1,249 @@
...
+
+    class SGrab {
+      getInfo() {
+        return {
+          id: 'sgrab',
```
GarboMuffin, line 13: change the extension ID to somehow incorporate your username eg. `nexuskittensgrab`

NexusKitten: Alright, should be changed now

PR #373, opened 2023-05-01
Post-Process 

LoserFox:
used to achieve full-screen effects in TurboWarp.
image

### DIFF
--- a/extensions/LoserFox/post-process.js
+++ b/extensions/LoserFox/post-process.js
@@ -421,7 +421,7 @@ rendererDrawPrefix();
       switch (drawprogram_mode){
         case "glitch":
           gl.useProgram(drawprogram);
-          setUniform1f(gl,"_Time", Date.now() / 1000.0);
+          setUniform1f(gl,"_Time", Math.random());
       }
     }
     class postprocessing {
--- a/extensions/LoserFox/post-process.js
+++ b/extensions/LoserFox/post-process.js
@@ -114,6 +114,31 @@
 
       }
     `;
+    var PointillismShaderCode = `
+    precision mediump float;
+
+    uniform sampler2D u_texture;
+    uniform float u_size; // Point size
+    uniform float u_threshold; // Threshold for black and white dots
+    uniform vec2 u_resolution; // Resolution of the canvas
+    
+    varying vec2 v_texcoord;
+    
+    void main() {
+      // Sample the color from the texture
+      vec4 texColor = texture2D(u_texture, v_texcoord);
+      
+      // Convert the color to grayscale
+      float gray = dot(texColor.rgb, vec3(0.299, 0.587, 0.114));
+      
+      // Quantize the grayscale value to either black or white
+      float threshold = u_threshold * (1.0 + 0.2 * (sin(v_texcoord.y * u_resolution.y * 10.0) + sin(v_texcoord.x * u_resolution.x * 10.0)));
+      float quantized = step(threshold, gray);
+      
+      // Combine the dot pattern with the quantized color
+      gl_FragColor = vec4(vec3( quantized), 1.0);
+    }
+    `;
     var quadPositions = [
       -1, -1,
       -1, 1,
@@ -207,15 +232,15 @@
         glitch: null,
         dispersion: null,
         gray: null,
-        reverse: null
-
+        reverse: null,
+        pointillism: null
       };
       programs.none = createProgram(gl,createshader(gl,gl.VERTEX_SHADER,vertexShaderCode),createshader(gl,gl.FRAGMENT_SHADER,noneShaderCode));
       programs.glitch = createProgram(gl,createshader(gl,gl.VERTEX_SHADER,vertexShaderCode),createshader(gl,gl.FRAGMENT_SHADER,GlitchShaderCode));
       programs.dispersion = createProgram(gl,createshader(gl,gl.VERTEX_SHADER,vertexShaderCode),createshader(gl,gl.FRAGMENT_SHADER,dispersionShaderCode));
       programs.gray = createProgram(gl,createshader(gl,gl.VERTEX_SHADER,vertexShaderCode),createshader(gl,gl.FRAGMENT_SHADER,GrayShaderCode));
       programs.reverse = createProgram(gl,createshader(gl,gl.VERTEX_SHADER,vertexShaderCode),createshader(gl,gl.FRAGMENT_SHADER,ReverseShaderCode));
-
+      programs.pointillism = createProgram(gl,createshader(gl,gl.VERTEX_SHADER,vertexShaderCode),createshader(gl,gl.FRAGMENT_SHADER,PointillismShaderCode));
       return programs;
     }
 
@@ -354,6 +379,9 @@ rendererDrawPrefix();
         console.log('STAGE_SIZE_CHANGED. resize the post-process framebuffer.');
         gl.bindTexture(gl.TEXTURE_2D, framebuffertexture);
         gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.canvas.width, gl.canvas.height, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
+        if (drawprogram_mode ==  "pointillism"){
+          setUniform2fv(gl,"u_resolution", gl.canvas.width, gl.canvas.height);
+        }
       }
     }
     vm.runtime.on('PROJECT_RUN_START',_ => {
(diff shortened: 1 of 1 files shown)


TheShovel:
The extension doesn't seem to work.

LoserFox:
> The extension doesn't seem to work.
> image
> 

double click the green flag

LukeManiaStudioss:
image
I'm seeing similar problems.

TheShovel:
> > The extension doesn't seem to work.
> > image
> 
> double click the green flag,I can't reproduce this issue stably, can you give me more info?
I just clicked on the glitch effect block. Did nothing else.

TheShovel:
> > The extension doesn't seem to work.
> > image
> 
> double click the green flag,I can't reproduce this issue stably, can you give me more info?

I just clicked on the glitch effect block. Did nothing else.
I am on the latest version of Chromium.

TheShovel:
Does it only update when the sprite's skin gets updated?
When you move it on the X and Y, it doesn't update the effect. I think you should add a `Step effect block`, so users can update the effect without having to update the sprite's skin.

TheShovel:
Also, maybe remove some of the unnecessary logs.

LoserFox:
> Does it only update when the sprite's skin gets updated?
No, just make sure the screen is redraw. move on the position can update effect.
> Also, maybe remove some of the unnecessary logs.
yes, i will remove them

GarboMuffin:
what we really need is for the renderer to be extensible enough that you shouldn't need all these terrible internal hacks...

LoserFox:
so what should I do for next

GarboMuffin:
extensions that interact deeply with the renderer such as this one will always take a bit longer to get reviewed

LoserFox:
Is there any way to speed up the process?

LukeManiaStudioss:
How about just be patient

GarboMuffin:
Based on above comments it sounds like many people have trouble figuring out how to use this extension. It may be beneficial to make it work as people expect

NexusKitten:
@GarboMuffin I think it would be a good idea to close this PR, it's been inactive for over a month and it's not in a stable enough state to merge

LoserFox:
I've made some final changes and it's now much easier to use. I'm sorry I've had a lot to deal with in the IRL lately and I'm a little late to the party

NexusKitten:
I like the update, some things I noticed though:
- "opcodeRequestReDraw" doesn't work and instead returns an error. 

image

- When using a block incompatible with an effect, it logs in the console twice instead of once. It doesn't really bug me however thought I'd mention it.

image

- Not exactly sure how the extension works (I've never worked with post processing before) however it'd be nice if you could put the three percentage blocks in a dropdown to be more concise.

- Documentation would be nice since this extension is on the more complicated side.

This extension is working up to be pretty promising though, can't wait to see the finished product!

LoserFox:
> "opcodeRequestReDraw" doesn't work and instead returns an error. 
Restored.
> When using a block incompatible with an effect, it logs in the console twice instead of once. It doesn't really bug me however thought I'd mention it.
I wasn't able to reproduce this, and I always saw that they only hinted at it once
>Not exactly sure how the extension works (I've never worked with post processing before) however it'd be nice if you could put the three percentage blocks in a dropdown to be more concise.
The reason I separated them at first was because there could be more than one parameter. The type of parameters may also be different.
>Documentation would be nice since this extension is on the more complicated side.
write a docs maybe a better choose, but final version(now) are really simple to use..... OKAY i will try to make a example and  with docs.

NexusKitten:
New update fixes most of the issues I had, including the double console logging for some reason lol

LoserFox:
now I need to hold myself back from looking for that certain reason.

LoserFox:
Since clippingblending duck punching gl.bindFramebuffer will stop working when using custom framebuffer, no reason is known at the moment. 
Use WebGLRenderingContext.prototype.bindFramebuffer to bypass the monkey patching it does to achieve the effect. 
If Xeltalliv could explain specifically what this test is for, I think there might be a better way to do it. @Xeltalliv

GalacticCoder922:
Will this go well? This pull request is VERY controversial due to errors.

Xeltalliv:
> Since clippingblending duck punching gl.bindFramebuffer will stop working when using custom framebuffer, no reason is known at the moment. Use WebGLRenderingContext.prototype.bindFramebuffer to bypass the monkey patching it does to achieve the effect. If Xeltalliv could explain specifically what this test is for, I think there might be a better way to do it. @Xeltalliv

The check is there to only apply effects of clipping and blending when drawing directly onto canvas or pen layer. Pen layer has flipped Y. All other cases for which turbowarp may draw sprite (like touching blocks, or sprite dragging) don't need to be affected by those effects.

If you want to make our extensions compatible, you can just modify my extension instead of using complicated workarounds

extensions/LoserFox/post-process.js
```diff
@@ -0,0 +1,753 @@
+(function(Scratch) {
```
GDH4ne, line 1: While using this extension, i've come upon a huge problem. When i use the effect, it literally freezes the screen and then when i toggle the effect away or there's some sort of movement, the project unfreezes itself. i'll link under here a video.

[video_3.webm](https://github.com/TurboWarp/extensions/assets/69966423/d87d2fde-5d9c-48af-8f71-30be3348117e)

GDH4ne, line 1: While using this extension, i've come upon a huge problem. When i use the effect, it literally freezes the screen and then when i toggle the effect away or there's some sort of movement, the project unfreezes itself. i'll link under here a video.

[video_3.webm](https://github.com/TurboWarp/extensions/assets/69966423/d87d2fde-5d9c-48af-8f71-30be3348117e)

LoserFox: I think this is scratch itself problem. I also provided a block to avoid this question: redraw post-process, only need put it in any loops.

LilyMakesThings: Consider that this behaviour doesn't happen when using the extension, and this isn't a Scratch problem.

LoserFox:
> > Since clippingblending duck punching gl.bindFramebuffer will stop working when using custom framebuffer, no reason is known at the moment. Use WebGLRenderingContext.prototype.bindFramebuffer to bypass the monkey patching it does to achieve the effect. If Xeltalliv could explain specifically what this test is for, I think there might be a better way to do it. @Xeltalliv
> 
> The check is there to only apply effects of clipping and blending when drawing directly onto canvas or pen layer. Pen layer has flipped Y. All other cases for which turbowarp may draw sprite (like touching blocks, or sprite dragging) don't need to be affected by those effects.
> 
> If you want to make our extensions compatible, you can just modify my extension instead of using complicated workarounds

I can't modify your extension without your agree. but you agree now

LoserFox:
> > Since clippingblending duck punching gl.bindFramebuffer will stop working when using custom framebuffer, no reason is known at the moment. Use WebGLRenderingContext.prototype.bindFramebuffer to bypass the monkey patching it does to achieve the effect. If Xeltalliv could explain specifically what this test is for, I think there might be a better way to do it. @Xeltalliv
> 
> The check is there to only apply effects of clipping and blending when drawing directly onto canvas or pen layer. Pen layer has flipped Y. All other cases for which turbowarp may draw sprite (like touching blocks, or sprite dragging) don't need to be affected by those effects.
> 
> If you want to make our extensions compatible, you can just modify my extension instead of using complicated workarounds

I can't modify your extension without your agree. but you agree now

LilyMakesThings:
> > > Since clippingblending duck punching gl.bindFramebuffer will stop working when using custom framebuffer, no reason is known at the moment. Use WebGLRenderingContext.prototype.bindFramebuffer to bypass the monkey patching it does to achieve the effect. If Xeltalliv could explain specifically what this test is for, I think there might be a better way to do it. @Xeltalliv
> > 
> > 
> > The check is there to only apply effects of clipping and blending when drawing directly onto canvas or pen layer. Pen layer has flipped Y. All other cases for which turbowarp may draw sprite (like touching blocks, or sprite dragging) don't need to be affected by those effects.
> > If you want to make our extensions compatible, you can just modify my extension instead of using complicated workarounds
> 
> I can't modify your extension without your agree. but you agree now

For future reference, you are allowed to modify anyone's extension as you wish, regardless of permission (so long as you don't break anything)

LoserFox:
I give up, not anymore, goodbye. thanks for the everyone testers.

gamerman14:
this pull request is insane 💀

GDH4ne:
> this pull request is insane 💀

this pull has really intense lore lol

PR #380, opened 2023-05-04
Reshuffle the extension gallery's extensions slightly

LukeManiaStudioss:
Move camera controls and clipping and blending higher up alongside utilities.

### DIFF
--- a/website/index.ejs
+++ b/website/index.ejs
@@ -393,6 +393,24 @@
         </div>
 
         <div class="extension">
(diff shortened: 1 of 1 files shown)


LukeManiaStudioss:
Closing this as I'm going to propose the changes with Clones+

True-Fantom:
It seems that we need to arrange something like a vote to determine the positions of extensions. You can make a lot of changes in them, but everyone will have their own opinion about them.

PR #381, opened 2023-05-04
Update bitwise.js

True-Fantom:
svgv999

Скриншот 04-05-2023 221307


Xeltalliv:
It is possible to simplify:

```javascript
 const number2bitsSimple = number => {
    return (number >>> 0).toString(2).padStart(32, '0').substr(-32);
  };
```


Perhaps it's also worth adding blocks for this:

```javascript
intBitsAsFloat = (x) => new Float32Array(new Int32Array([x]).buffer)[0];
floatBitsAsInt = (x) => new Int32Array(new Float32Array([x]).buffer)[0];
```

They could be used for emulators, data compression, [fast inverse square root](https://en.wikipedia.org/wiki/Fast_inverse_square_root)

extensions/bitwise.js
```diff
@@ -1,113 +1,259 @@
...
           {
             opcode: 'bitwiseXor',
             blockType: Scratch.BlockType.REPORTER,
-            text: '[LEFT] ^ [RIGHT]',
+            text: '[LEFT] xor [RIGHT]',
             arguments: {
               LEFT: {
                 type: Scratch.ArgumentType.NUMBER,
-                defaultValue: '1'
+                defaultValue: '\n'
```
GarboMuffin, line 198: empty strings are supposed to work here now, is it still broken?

True-Fantom: Oh, sorry, just out of habit

True-Fantom:
> It is possible to simplify:
> 
> ```js
>  const number2bits = number => {
>     return (number >>> 0).toString(2).padStart(32, '0').substr(-32);
>   };
> ```
> 

Oh! I didn't know there was a `.padStart` function! Thanks I will use it. (The only thing is that the `.substr` function seems to be outdated, but okay, I'll just replace it with `.substring`)

> Perhaps it's also worth adding blocks for this:
> 
> ```js
> intBitsAsFloat = (x) => new Float32Array(new Int32Array([x]).buffer)[0];
> floatBitsAsInt = (x) => new Int32Array(new Float32Array([x]).buffer)[0];
> ```
> 
> They could be used for emulators, data compression, [fast inverse square root](https://en.wikipedia.org/wiki/Fast_inverse_square_root)

Hm... Probably, with this, you can come up with many more blocks. Right now I don't really understand how typed arrays work, but I'll try to figure it out soon. In any case, I suggest including this in the next PR. If you are well versed in this and you have free time, I will be glad if you help.

Thank you for the offer.

True-Fantom:
@GarboMuffin Does this have any other problems?

GarboMuffin:
okay, have taken a deeper look now.

I'm not a huge fan of the use of the "32" image in these blocks. Is there a reason we can't just write out "is [x] binary", "[b] to number", "[n] to binary"?

image

I'm also worried about these ones. I think the simple left and right shifts should still be ">>"/"<<". The new names are still just symbols, so using the same symbols that other programming languages use  makes sense. For the `>>>` block, following the same logic it might make sense to use ">>>".

image

the names of the circular blocks and logic blocks are fine

True-Fantom:
> okay, have taken a deeper look now.
> 
> I'm not a huge fan of the use of the "32" image in these blocks. Is there a reason we can't just write out "is [x] binary", "[b] to number", "[n] to binary"?
> 
> image
> 

Yes, there are several reasons for this:

- So shorter than `number bits` (We can't just call it bits or binary)
- I got the idea to use images for new (for scratch) "data types" like url, json, date, regexp and so on
Скриншот 09-05-2023 160301
Скриншот 09-05-2023 160330

I think it's convenient, fits the scratch style and more clearly emphasizes that there is some kind of unique input field.
The only thing I doubt about the image itself, if there is any way to more clearly depict the bits of the number, I use it.


> I'm also worried about these ones. I think the simple left and right shifts should still be ">>"/"<<". The new names are still just symbols, so using the same symbols that other programming languages use makes sense. For the `>>>` block, following the same logic it might make sense to use ">>>".
> 
> image
> 

I did this so that they were more like standard operators and math operators:

Скриншот 09-05-2023 161207

Also, this approach reduces blocks in size, but at the same time the main meaning is not lost, triple arrows are also triple arrows, and the cyclic shift in general becomes more visual than writing it through `<<<<` and `>>>>`

> the names of the circular blocks and logic blocks are fine

True-Fantom:
@GarboMuffin Please answer, it can already merge?

True-Fantom:
@GarboMuffin I understand all the changes except the reduction of binary values. Their whole point is that they consist of 32 characters. If you write them abbreviated, it will not only break the logic of the extension, but also, in principle, the operation of the blocks!

IMG_20230515_214911
IMG_20230515_214900

PR #382, opened 2023-05-04
Add ClonesPlus extension

LukeManiaStudioss:
ClonesPlus adds various blocks to make the cloning features on Scratch far more useful.

Extension thumbnail (image and uncropped SVG):
image

ClonesPlus


website/index.ejs
```diff
@@ -441,9 +429,15 @@
         </div>
 
         <div class="extension">
-          <%- banner('utilities') %>
-          <h3>Utilities</h3>
-          <p>A bunch of interesting blocks. Originally created by <a href="https://scratch.mit.edu/users/Sheep_maker/">Sheep_maker</a>.</p>
+          <%- banner('clonesplus') %>
```
GarboMuffin, line 432: banner('LukeManiaStudios/ClonesPlus')

LukeManiaStudioss: oops 💀

LukeManiaStudioss: I'll get used to this eventually 😭

GarboMuffin: eventually we will clean up the mess that is index.ejs. a mistake like this really should not be able to have "All checks have passed"

website/index.ejs
GarboMuffin, line 1: your formatter made more changes to this file than it should've

LukeManiaStudioss: You think? 😭

LukeManiaStudioss:
Are there any more issues with this that I should address before it gets merged?

GarboMuffin:
ended up making quite a few changes to prevent renaming variables from causing as many issues (menus now use variable IDs instead of names); consider taking a look to see if there's any regressions

LukeManiaStudioss:
First of all, I had not even considered renaming variables so I appreciate you not only pointing that out but resolving it yourself, that was impeccably helpful

I just tried it, and I have no problems with it. Try as I might I can't break it lol

David-Orangemoon:
LETS GOOOOOOO.

UndermanReal:
I think you should also add a "touching clone with (var) containing (value)?"

LukeManiaStudioss:
1. Make this an issue
2. That block already exists, with limited functionality

PR #405, opened 2023-05-13
Add Looks Plus

LukeManiaStudioss:
Looks Plus adds blocks to expand on Scratch's Looks Category
image

Extension Thumbnail:
looksPlusThumbnail


GarboMuffin:
in the futur you should try using argument names like SOUND or SPRITE instead of INPUTA INPUTB INPUTC

GarboMuffin (changes requested):
Marking this as "request changes" so that I don't accidentally merge it before implementing some changes in the renderer to prevent the snapshot block from including webcam images

Some thoughts:

 - should the "[color] effect"  block be instead "[color] effect of [Sprite1]"
 - should the "set layer to [1]" block be instead "set layer of [Sprite1] to [1]"
 - "set layer to" and "layer # of" might want to be consistent about using "layer" or "layer #"
 - should "[SVG] of costume # [1] of [this sprite]" use a typical `COSTUME` input instead of the numbers?
 - it is possible for "set SVG for [costume1] with [...]" to also support bitmap costumes?
 - similar for "reset SVG" (currently it throws an error on bitmap costumes)

LukeManiaStudioss:
I agree with all of this, except can a standard costume input accept a number value if required?

LukeManiaStudioss:
I think it's worth using try/catch on these blocks...
image
...because there are so many ways they _could_ stall.

As for letting them support bitmaps, I simply do not know how to make that work, else I would've added it before making a pull request.

LukeManiaStudioss:
> I think there should also be a costume X, costume Y, and costume directions that can affect the costume's center. Like if you wanted to use the "move steps" block while spinning a sprite, you would have to use multiple costumes for that, but if there were costume directions you could just turn the costume direction instead of the ACTUAL direction.

Will add it the second I can find a way to add it.

Martinelplayz:
i am excited for this extension.

LukeManiaStudioss:
> Will add it the second I can find a way to add it.

Will add it the second I can find a stable way to add it.

LukeManiaStudioss:
not me clicking "close with comment" by mistake lol

Martinelplayz:
block suggestion : (reporter) 
(costume # of costume named ("dango"))

LukeManiaStudioss:
Just to be clear, this version of the extension is done.

If I post anything else about it, like new blocks, I'm intending for those to go into an update on another pull request once this is merged.

PR #407, opened 2023-05-13
Block to verify if the string is Base64

TheShovel:
image

### DIFF
--- a/extensions/encoding.js
+++ b/extensions/encoding.js
@@ -257,13 +257,15 @@
         }
 
         isBase64(args) {
-            if (args.string ==='' || args.string.trim() ===''){ return false; }
-            try {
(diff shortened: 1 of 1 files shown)


TheShovel:
I am not going to bother with Eslint. Someone else make it happy.

LukeManiaStudioss:
It's 4 errors my g

TheShovel:
I don't know what the terms mean. I tried doing stuff like how Eslint says but it just caused more issues. The extension works just fine, it's complaining that it's not formated how GarboMuffin wants.

LukeManiaStudioss:
On line 260:
`if (args.string === '' || args.string.trim() === '') {`
`     return false;`
`}`

TheShovel:
Happy?

PR #408, opened 2023-05-14
Add Temporary Variables

LukeManiaStudioss:
This extension is the temporary variables from LMS Utilities but in a separate extension.
image

(thumbnail, however I doubt it'd be necessary)
tempVarsThumbnail


LukeManiaStudioss:
I scuffed this PR

PR #410, opened 2023-05-14
Add Temporary Variables extension.

LukeManiaStudioss:
Temporary Variables allows you to create variables that don't clog up the data category and are invisible.

(This is exactly the same as the temp variables in LMS Utilities, just into its own extension)
image

Thumb:
TempVariables


LukeManiaStudioss:
Brilliant changes, thank you very much

PR #411, opened 2023-05-14
Update runtime-options.js

LukeManiaStudioss:
image

### **General Changes:**
- Color of blocks changed to a grey as blue in Scratch commonly associated with "motion".
- "Framerate" on blocks re-specified as "framerate limit" to avoid confusion.

### **New blocks:**
image
get clone # : reports the amount of clones on the stage

image
set username to () : changes the runtime username

image
run green flag : simulates a green flag click

### **Newer Thumbnail Idea**
I didn't commit this but here's also a thumbnail to match the other main TW extensions
RuntimeOptions

### DIFF
--- a/extensions/runtime-options.js
+++ b/extensions/runtime-options.js
@@ -73,11 +73,6 @@
           '---',
 
           {
-            opcode: 'getCloneCount',
-            text: 'get clone #',
-            blockType: Scratch.BlockType.REPORTER
-          },
-          {
             opcode: 'getCloneLimit',
             text: 'get clone limit',
             blockType: Scratch.BlockType.REPORTER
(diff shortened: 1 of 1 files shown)


OeildeLynx31:
I don't know how, but ok.

UndermanReal:
Oh also, I don't know why, but whenever you change the stage size, the quality drops. But this also happens when you turn on fullscreen.

GarboMuffin:
set username is fine, color change is fine, framerate rename is fine, green flag is marginal but fine, but not convinced we need another clone # block

> Oh also, I don't know why, but whenever you change the stage size, the quality drops. But this also happens when you turn on fullscreen.

probably not extension's fault, report separately

True-Fantom:
@LukeManiaStudioss Maybe add `isTerminal: true` for the green flag?

LukeManiaStudioss:
Alright, that's fine, we can remove that

PR #412, opened 2023-05-14
Add Comment Blocks

LukeManiaStudioss:
Comment blocks allow you to annotate your scripts more clearly.

image

### **Example:**
image


gomiigg:
Color makes it a little hard to read the "//" text but other than that this looks good

PR #414, opened 2023-05-15
Bug fix

True-Fantom:


True-Fantom:
@GarboMuffin https://github.com/TurboWarp/extensions/pull/381#issuecomment-1548383701

GarboMuffin:
we should instead change the blocks because "00101010100" and "10100101000101010010101001010101010100001010010101010" are clearly valid binary; don't see any need to restrict it to 32 bits exactly

True-Fantom:
@GarboMuffin This is due to the structure of the number in js, exactly 32 bits are allocated to the number, so there are 2 options: either remove all the leading zeros from the values, or leave it as it is. The second option is more convenient, since the blocks by type set the letter, will work as a set bit, without any problems.

True-Fantom:
@GarboMuffin Plus, in the second option, you do not need to change the code of the returned values.

True-Fantom:
In this pr, I fixed all the main problems, returned the old bit values , replaced "binary" with "number bits", since this is much more correct, replaced the record of cyclic shifts so that they correspond to the programming languages and previous shifts. So, I think it needs to be accepted, because everything is done here naturally.

True-Fantom:
@GarboMuffin ?

PR #415, opened 2023-05-15
bitwise: Make the new blocks work on numbers that aren't 32 bits

GarboMuffin:
This is how our average user will expect the blocks to work. Regardless of what these blocks are called, someone will try to make a "binary to number" converter project and be very confused why it doesn't work if they don't enter exactly 32 bits and we will receive bug reports. If someone wants the behavior of only accepting 32-bits, they can add that on their own.

image

The circular bit shifts are fine to keep using their symbol. I am not aware of any programming language that has a dedicated operator for circular shifts, so there is no symbol that we would be expected to use.

Closes https://github.com/TurboWarp/extensions/pull/414


True-Fantom:
@GarboMuffin It shouldn't work like that, bitwise operations have their own logic. `Скриншот 15-05-2023 222400`. It has its own rules, if a number is greater than a certain value, it becomes negative, it does not support the fractional part and much more.

True-Fantom:
@GarboMuffin I'll probably make an extension to change the number's calculus system.

GarboMuffin:
You are welcome to submit another extension with the changes you desire, as I believe the current blocks, as written in this PR, will match what our users expect.

True-Fantom:
@GarboMuffin 
Скриншот 15-05-2023 223849

True-Fantom:
Plus an arbitrary input length is strange, because the output will always be only 32

GarboMuffin:
You are welcome to add a new extension or new blocks that operates with the format that you describe.

PR #418, opened 2023-05-15
Add base.js and update bitwise.js

True-Fantom:
svgviewer-output (5)

svgv999

Reasons why this PR should be merge:

- The binary representation of numbers has nothing to do with bits, so it doesn't correspond to the extension. It needs to be removed.

- An analog has been added to do this, so TurboWarp will not lose any capabilities, but on the contrary will increase them.

- Documentation has been added so that now, new users will not have misunderstandings about the functions of the blocks.


True-Fantom:
Considering that the changes were made recently, I think we can neglect backward compatibility. And also I will make documentation for bitwise so that there is no confusion.

True-Fantom:
@GarboMuffin I have already started writing documentation. When I finish it and add it to the extension, will you accept this PR? https://hackmd.io/@vJu6RuV8RCKRkItVMFOkpA/B1iL_fgH3

True-Fantom:
I made and added the finished documentation. I'll probably refine it further, but nevertheless it's already possible to merge.

True-Fantom:
I've done backward compatibility...

True-Fantom:
@GarboMuffin What do you think about this decision? It seems to me that it is quite harmonious - it does not violate logic and does not create misunderstandings.

True-Fantom:
@GarboMuffin I understand you. Let's do this for now: I will close this PR, and make a new one only for base, and I will update bitwise later when I finalize everything and think about it.

PR #419, opened 2023-05-15
encoding: Don't unnecessarily escape everything and fix other bugs

GarboMuffin:
Apparently all the base64 and md5 blocks have been doing `encodeURIComponent` on the input before encoding/hashing, so they've been returning the wrong thing for quite some time.

Base 64 decode would also throw uncaught error if input was invalid

Also replaced the md5 function with one that isn't minified


GarboMuffin:
A little concerned that this is *technically* a breaking change

PR #424, opened 2023-05-16
add extension for dictionaries

lopste:
This extension allows you to create dictionaries, parse JSON into one for usage, and then stringify them back. (`vercte/dictionaries.js`)
List of blocks in the extension, which include modifying dictionaries and converting them to and from JSON

Thumbnail:
Thumbnail
(yes I completely copied lukemaniastudioss' pull request template I'm new here)


OeildeLynx31:
What are differences between this extension and the JSON extension?

LukeManiaStudioss:
This looks really cool! 👌👌

lopste:
>
> What are differences between this extension and the JSON extension?
>
This contains support for actively writing to the object and easily
retrieving keys from it, along with incrementing keys. Eventually, I’ll add
support for nested objects and arrays. It also internally uses maps to make
it more secure (but I’m not too sure how far you can breach into it using
Scratch as a front-end.)

OeildeLynx31:
Ok, thank you!

extensions/vercte/dictionaries.js
```diff
@@ -0,0 +1,142 @@
+window.dictionaries = new Map();
```
GarboMuffin, line 1: in addition to all the other warnings, extensions cannot store things on `window`; use the IIFE from https://docs.turbowarp.org/development/extensions/unsandboxed then just define `var dictionaries = `

extensions/vercte/dictionaries.js
```diff
@@ -0,0 +1,142 @@
...
+class DictionaryExtension {
+  getInfo() {
+    return {
+      id: 'verctedictionaries',
+      name: 'Dictionaries',
```
GarboMuffin, line 6: Also, since your image has a well defined color scheme, you should set the block color accordingly

lopste:
Also, looks like the lint failed. I swear I linted it, hold on.

lopste:
I fixed all the issues mentioned and added some useful new blocks since keys couldn't be easily removed.

GarboMuffin:
do you consider this extension to be done or are you adding more changes

lopste:
I think it's good for not. I may add more changes in the feature, specifically nested objects, but it's done for now.

extensions/vercte/dictionaries.js
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  dict_delete_key({ KEY, DICT }) {
+    if (dictionaries.has(DICT)) {
+      dictionaries.get(DICT).delete(KEY);
+    }
+  }
+
+  dict_primitive_null() {
+ return null;
```
GarboMuffin, line 192: blocks can't return null. unknown things break when that happens

lopste:
Oh. Hm. I guess I should just remove that block and add an option to
nullify a property instead. Or just bring null out of the equation.

LukeManiaStudioss:
> Does it need an icon? Screenshot of "Dictionaries" extension in editor

It's not required, no

PR #425, opened 2023-05-17
big changes to JSON extension

Skyhigh173:
v2 update, which changes the following :
- fixed some grammar mistakes
- new thumbnail (looks way better)
- rearranged blocks so it looks better
- new "filter key" block
- changed color of the blocks
- fixed a "JSON bug" where you can't save NaN, Infinity value

Hopefully it doesnt break compatibility!


Skyhigh173:
heres the thumbnail
json

Skyhigh173:
Some people say the old thumbnail is better than new one, so I am gonna switch it back

CST1229:
> about the `Comparing to itself is potentially pointless no-self-compare` error, I don’t know why `isNaN()` doesn’t work, but `x != x` worked

`isNaN()` also returns true if the value is not NaN but is not a number (for example, a string which contains letters). Use [Number.isNaN()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN) instead.

PR #426, opened 2023-05-17
Add NOname-awa/graphics2d image and update some attribution

GarboMuffin:


Pizzalover4783:
Missed opportunity to add the RixxyX Thumbnail

PR #429, opened 2023-05-17
Update LMS Utilities

LukeManiaStudioss:
- Remove 26 blocks that are made redundant by other extensions (EDIT FROM MAINTAINERS: "Remove" here means "Hide")
- Update block colour to be far more readable
- Remove it from the gallery to discourage people from using it

### DIFF
--- a/extensions/LukeManiaStudios/lmsutils.js
+++ b/extensions/LukeManiaStudios/lmsutils.js
@@ -742,6 +742,7 @@
             opcode: 'goToLink',
             blockType: Scratch.BlockType.COMMAND,
             text: 'open link [INPUT] in new tab',
+            hideFromPalette: true,
             arguments: {
               INPUT: {
                 type: Scratch.ArgumentType.STRING,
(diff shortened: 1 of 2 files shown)


Scratchfangs:
Some of my projects that are in .html format right now uses a lot of features in the LMS extension. Removing them would completely break my project and for everyone who is using it.

LukeManiaStudioss:
> Some of my projects that are in .html format right now uses a lot of features in the LMS extension. Removing them would completely break my project and for everyone who is using it.

There are several problems with this message:
- "Removing" blocks has no impact on existing projects. It hides them from the palette and they are permanently accessible.
- Projects in HTML format can be unpackaged with the Turbowarp Unpackager

Scratchfangs:
There is a Turbowarp Unpackager??

CST1229:
> There is a Turbowarp Unpackager??

yes, https://turbowarp.github.io/unpackager/

GarboMuffin:
Just to clarify for anyone who is worried about removing blocks breaking their projects: projects that use them still work, the blocks just aren't listed in the palette

Martinelplayz:
why are you trying to remove it?

SurvExe1Pc:
"to discourage people from using it", and because 26 of its blocks are "redundant"

PR #437, opened 2023-05-19
clippingblending: Add more blending options

CST1229:
Adds 2 more blending options to the Clipping and Blending extension:

- `subtract`
Your standard subtract option.

- `invert`
Somewhat hard to explain; inverts the sprite when on a white background, and makes it normal on a black one (I would have reversed this behavior so white is normal, but I couldn't find a way to do that). Useful for making a "black-and-white contrast thing" effect.

https://github.com/TurboWarp/extensions/assets/68464103/587271ae-1f2e-44c8-9354-74a9b1f32491

### DIFF
--- a/extensions/Xeltalliv/clippingblending.js
+++ b/extensions/Xeltalliv/clippingblending.js
@@ -92,12 +92,23 @@
     }
     switch (blendMode) {
       case 'additive':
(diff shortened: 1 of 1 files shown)


GarboMuffin:
@Xeltalliv

PR #449, opened 2023-05-21
Fix IOS touch events and add more QOL features for touch events.

David-Orangemoon:
Most IOS browsers do not support simultaneous touch events. So you have to do a pretty hacky solution to it.

Made it so devices with more available touches also support those.

### DIFF
--- a/extensions/obviousAlexC/SensingPlus.js
+++ b/extensions/obviousAlexC/SensingPlus.js
@@ -25,7 +25,9 @@
     initializedSpeechRecognition = true;
 
     if (!SpeechRecognition) {
-      console.warn('Speech recognotion blocks are not supported in this browser');
+      console.warn(
+        "Speech recognotion blocks are not supported in this browser"
+      );
       return;
     }
 
@@ -37,11 +39,11 @@
         .join("");
     });
     speechRecognition.addEventListener("error", (event) => {
-      if (event.error === 'no-speech') {
+      if (event.error === "no-speech") {
         // this is fine, just ignore it
         // it will be restarted in the end handler
       } else {
-        console.error('Speech recognition error', event.error);
+        console.error("Speech recognition error", event.error);
         recording = false;
         speechRecognition.stop();
       }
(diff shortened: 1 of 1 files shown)


David-Orangemoon:
Dear garbo. Before you merge this please take into account that by removing VM surfer you accidentally broke the item [number] in the list block (actually all list blocks but still) so I will write a minor repair to that before you merge this.

David-Orangemoon:
Okay its fine now.

extensions/obviousAlexC/SensingPlus.js
```diff
@@ -128,6 +134,19 @@
         MATRIX: "matrix",
         NOTE: "note",
       },
+      Cast: {
```
GarboMuffin, line 137: FYI: this is fine for now, but with the added ability to properly run extensions from files, everything like this is probably going to be removed soon

extensions/obviousAlexC/SensingPlus.js
```diff
@@ -178,6 +200,9 @@
 
   /** @param {TouchEvent} event */
   function handleTouchMove(event) {
+    if (hackyTouchFix) {
```
GarboMuffin, line 203: is there a reason to not just *always* do this, and have you actually found a way to test this

GarboMuffin:
I am pretty sure the reason this doesn't work is because iOS doesn't support changedTouches

https://caniuse.com/?search=changedtouches

It does not seem that this would fix that

David-Orangemoon:
> I am pretty sure the reason this doesn't work is because iOS doesn't support changedTouches
> 
> https://caniuse.com/?search=changedtouches
> 
> It does not seem that this would fix that

but also [IOS docs](https://developer.apple.com/documentation/webkitjs/touchevent/1629351-changedtouches) say that I can use it.

extensions/obviousAlexC/SensingPlus.js
David-Orangemoon: > FYI: this is fine for now, but with the added ability to properly run extensions from files, everything like this is probably going to be removed soon

Yeah, the only reason I did add that was to test on other devices since I'm only at home for a small portion of the day.

GarboMuffin:
Okay we will merge this for now because it claims to fix some rather significant bugs. A more permanent patch may be written later

PR #458, opened 2023-05-24
added http request ext

RedMan13:


package-lock.json
GarboMuffin, line 1: don't commit changes to this file

extensions/godslayerakp/http.js
```diff
@@ -0,0 +1,445 @@
...
+
+    // inject comments bc yes
+    // @ts-ignore
+    const oldConvert = runtime._convertForScratchBlocks.bind(runtime);
+    const createComment = blockInfo => ({
+        info: blockInfo,
+        xml: `<label text="${blockInfo.slice(3)}"></label>`
+    });
+    // @ts-ignore
+    runtime._convertForScratchBlocks = (blockInfo, categoryInfo) => {
```
GarboMuffin, line 16: use

blockType: Scratch.BlockType.LABEL,
text: "text"

instead of that insecure mess

development/server.js
```diff
@@ -65,6 +65,6 @@ app.use((req, res) => {
 
 // The port the server runs on matters. The editor only treats port 8000 as unsandboxed.
 const PORT = 8000;
-app.listen(8000, () => {
+app.listen(PORT, () => {
```
GarboMuffin, line 68: move to a separate pr

extensions/godslayerakp/http.js
```diff
@@ -0,0 +1,445 @@
...
+
+        /**
+         * no need to install runtime as it comes with Scratch var
+         */
+        constructor() {
+            this.clearAll();
+        }
+        getInfo() {
+            return {
+                id: 'extawebrequestsHTTP',
```
GarboMuffin, line 123: prefix extension ids with something related to username

extensions/godslayerakp/http.js
```diff
@@ -0,0 +1,445 @@
...
+                    request.events.activate('reqFail');
+                    request.events.activate('reqEnd');
+                });
+        }
+    }
+
+    const instance = new WebRequests();
+    Scratch.extensions.register(instance);
+    // @ts-ignore
+    runtime.ext_http = instance;
```
GarboMuffin, line 444: you shouldn't need to do that

GarboMuffin: is there a good reason those extra blocks can't just be in the extension? there is no precedent for an extension to require another one to be used first

extensions/godslayerakp/http.js
```diff
@@ -0,0 +1,445 @@
...
+                            "video/mp4",
+                            "video/x-ms-wmv",
+                            "video/x-msvideo",
+                            "video/x-flv",
+                            "video/webm"
+                        ],
+                        acceptReporters: true
+                    },
+                    jsTypes: {
+                        items: [
```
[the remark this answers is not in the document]
RedMan13: wdym

extensions/godslayerakp/http.js
RedMan13: > is there a good reason those extra blocks can't just be in the extension? there is no precedent for an extension to require another one to be used first

yes and no, i put them separately becuase they look ugly and dont rlly fit well into the main extension imo

extensions/godslayerakp/http.js
RedMan13: oh

development/server.js
RedMan13: alr

extensions/godslayerakp/http.js
RedMan13: its not insecure? and this merge was pre-labels but yes it needs changed

RedMan13: fixed...

development/server.js
RedMan13: fixed...

extensions/godslayerakp/http.js
RedMan13: fixed...

package-lock.json
RedMan13: idk how to fix

RedMan13:
and now its broken :/

package-lock.json
RedMan13: nvm figured it out i think

extensions/godslayerakp/http.js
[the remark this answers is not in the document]
RedMan13: ima close this since there isnt any reason for it i can see

extensions/godslayerakp/http.js
RedMan13: reading it again i noticed you specified "required", they are not meant to be required and for the most part are not

GarboMuffin: it seems to me that every block in httpExtra uses ext_http, so yes it is a required dependency

GarboMuffin:
thanks for your patience

does httpExtra do anything that isn't possible with other JSON extensions here?

if it does, is there a good reason we can't just put them in the main extension? perhaps using a button to hide them by default, but it doesn't look like there's too many there anyways

RedMan13:
> thanks for your patience
> 
> does httpExtra do anything that isn't possible with other JSON extensions here?
> 
> if it does, is there a good reason we can't just put them in the main extension? perhaps using a button to hide them by default, but it doesn't look like there's too many there anyways

yes it does do more then what json can as it manualy edits the whole request options object 
a button to toggle it is actualy a rlly good idea

extensions/godslayerakp/http.js
RedMan13: > it seems to me that every block in httpExtra uses ext_http, so yes it is a required dependency

oh you meant the other way
i was reading it as http requiring httpExtra

GarboMuffin:
thank you for your patience

PR #460, opened 2023-05-25
update JSON extension - new blocks, labels

Skyhigh173:
## new blocks
image
## labels
- General Utils
- JSON Strings
- Array
- Advanced
- VM
## bug fixes
- when you put reporter blocks into vm_setlist, vm_getlist, it wont work. (fixed)

### DIFF
--- a/extensions/Skyhigh173/json.js
+++ b/extensions/Skyhigh173/json.js
@@ -789,7 +789,7 @@
     }
     json_vm_setlist({ list, json }, util) {
       try {
-        const listVariable = util.target.lookupVariableById(list);
+        let listVariable = util.target.lookupVariableById(list);
 
         if (listVariable == undefined) {
           listVariable = this.getListsID(util).filter(x => x.text === list)[0].value;
--- a/extensions/Skyhigh173/json.js
+++ b/extensions/Skyhigh173/json.js
@@ -1,7 +1,14 @@
 (function(Scratch) {
   'use strict';
   const vm = Scratch.vm;
+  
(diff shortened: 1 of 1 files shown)


LukeManiaStudioss:
"VM" is ambiguous to people who aren't familiar with how TurboWarp works. Consider renaming it to "TurboWarp"

Skyhigh173:
> "VM" is ambiguous to people who aren't familiar with how TurboWarp works. Consider renaming it to "TurboWarp"

thanks, I will rename it

Skyhigh173:
I will make some new blocks (isArray, Array.from) etc

GarboMuffin:
Some quick thoughts:

 - I replaced blocks like "new JSON" with "new Object" to reduce ambiguity; an array is valid JSON
 - Is there a good reason for not equal to be a dedicated block? At least right now the symbol is very easy to confuse at small screen resolutions.

Skyhigh173:
> Some quick thoughts:
> 
> * I replaced blocks like "new JSON" with "new Object" to reduce ambiguity; an array is valid JSON but is not an object
> * Is there a good reason for not equal to be a dedicated block? At least right now the symbol is very easy to confuse at small screen resolutions.

how about < () [= v] () > ? it i becomes menu and you can select equal or not equal

GarboMuffin:
sure

Skyhigh173:
> how about < (A) [= v] (B) > ? it i becomes menu and you can select equal or not equal
IMG_3942
(both are same block)

Skyhigh173:
I use my phone to edit file.
Please check if there is any mistakes, thanks!

PR #471, opened 2023-05-27
Add translations to NOname-awa/graphics2d

NOname-awa:
I have added Chinese language


GarboMuffin:
due to some bugs in turbowarp desktop v1.7.2-beta.1, we will have to wait a bit for this

GarboMuffin:
We will tolerate these manual translations from native speakers for now as a way to test the l10n system

Other contributors: please do not start writing translations by hand for languages you are not fluent in!

PR #479, opened 2023-05-29
obviousAlexC/SensingPlus: Ask for permission before recording audio or reading from clipboard

GarboMuffin:


David-Orangemoon:
👍

Issue #483, opened 2023-05-29
Sensing + speech recognition isn't detecting if recording is on even though it is


Boxxedboi:
![image](https://github.com/TurboWarp/extensions/assets/86323360/ba175bd5-0cda-4e4e-9737-67a005caed28)

LukeManiaStudioss:
The speech recognition feature is only available for Chrome and Safari. And no, chromium browsers like Opera GX don't count.

Boxxedboi:
oh, thanks

Boxxedboi:
> The speech recognition feature is only available for Chrome and Safari. And no, chromium browsers like Opera GX don't count.

what about the tw app?

David-Orangemoon:
> > The speech recognition feature is only available for Chrome and Safari. And no, chromium browsers like Opera GX don't count.
> 
> what about the tw app?

Nope.

PR #484, opened 2023-05-29
obviousAlexC/SensingPlus: Hide speech blocks

GarboMuffin:
hopefully temporary, but not optimistic

related to https://github.com/TurboWarp/extensions/issues/483

they don't work:

 - In the desktop app
 - In packaged electron projects
 - In Firefox
 - In most Chromium forks
 - In most WebKit forks
 - When offline

and the privacy aspect of sending audio to a remote server for transcription is slightly scary.

### DIFF
--- a/extensions/obviousAlexC/SensingPlus.js
+++ b/extensions/obviousAlexC/SensingPlus.js
@@ -573,18 +573,21 @@
                 menu: "toggleMenu",
               },
             },
+            hideFromPalette: true,
           },
           {
             opcode: "returnWords",
             blockType: Scratch.BlockType.REPORTER,
             text: "Recognized Words",
             blockIconURI: speechIco,
+            hideFromPalette: true,
           },
           {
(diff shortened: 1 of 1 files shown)


David-Orangemoon:
I understand.

LukeManiaStudioss:
I think it's worth noting Scratch developed their own speech2text extension, so at some point they must've thought it would've been a good idea.

Furthermore, the translate extension sends requests to Google. While that is not quite the same as sending recorded snippets, Scratch seems to trust Google for providing APIs in their extensions.

I agree with a point made before that these should be separated into their own extension, perhaps listed on the gallery but marked as only compatible with certain browsers.

GarboMuffin:
Scratch's speech2text worked by uploading audio data to a Scratch-controlled server who probably then uploaded them to Google. At some point they seem to have concluded that this is not a good thing to add to Scratch, perhaps for privacy or cost concerns.

PR #447, opened 2023-05-21
Open source port of Scratch Labs Animated Text

GarboMuffin:
## Help us test this: https://github.com/TurboWarp/extensions/pull/447#issuecomment-1556334233

Fully open source clone of https://lab.scratch.mit.edu/text/. Projects should be mostly compatible. This extension was written by testing Scratch Labs and seeing what the blocks actually do, not by reading the source code. Unlike some other mods, this version probably doesn't violate the GPL.

closes https://github.com/TurboWarp/scratch-vm/issues/127

closes https://github.com/TurboWarp/packager/issues/209

closes https://github.com/TurboWarp/scratch-vm/issues/47

related to https://github.com/TurboWarp/extensions/issues/357

PR #485, opened 2023-05-30
Animated Text: Add Reporter and "text active?" boolean

LukeManiaStudioss:
### **Let's fill in some gaps.**

image


Dy-tech:
This extension is designed to be compatible with the [Scratch Lab](lab.scratch.mit.edu) equivalent, so we can't add any extra blocks.

LukeManiaStudioss:
> This extension is designed to be compatible with the [Scratch Lab](lab.scratch.mit.edu) equivalent, so we can't add any extra blocks.

The goal of this extension is to be compatible with Scratch Lab projects, however GarboMuffin has never turned down the idea of adding new blocks, even stating ["more blocks may be added later"](https://github.com/TurboWarp/extensions/pull/447#issuecomment-1557840433).

Dy-tech:
@GarboMuffin, Can @LukeManiaStudioss add more blocks?

LukeManiaStudioss:
> @GarboMuffin, Can @LukeManiaStudioss add more blocks?

image

PR #488, opened 2023-05-30
Box 2D: Add 12 New Blocks

LukeManiaStudioss:
- Boolean to check the static mode of an object.
image

- Blocks to set/get the Angular Velocity of an object.
image

- The long property block has been split into 3 separate blocks. 3 more blocks have been added to set a custom value to each property, as well as to get the property's values afterwards. This has been highly requested.
image

### DIFF
--- a/extensions/box2d.js
+++ b/extensions/box2d.js
@@ -13345,7 +13345,7 @@
       if (!body) {
         body = this.setPhysicsFor(util.target);
       }
-      body.SetAngularVelocity(args.force);
+      body.SetAngularVelocity(-Cast.toNumber(args.force));
     }
 
     angVelocity(args, util) {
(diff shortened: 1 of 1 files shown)


Martinelplayz:
ooh. looks cool

extensions/box2d.js
```diff
@@ -13147,6 +13340,23 @@
       body.ApplyTorque(-Cast.toNumber(args.force));
     }
 
+    setAngVelocity(args, util) {
+      let body = bodies[util.target.id];
+      if (!body) {
+        body = this.setPhysicsFor(util.target);
+      }
+      body.SetAngularVelocity(args.force);
```
GarboMuffin, line 13348: Using the pre-existing spin with force block, a positive value causes a clockwise rotation, but here a positive value means counterclockwise. The new blocks should match the original.

LukeManiaStudioss: How should I go about fixing this? Do I just invert the value of the argument input?

PR #490, opened 2023-05-30
Update Library Website

NexusKitten:
Adds multiple quality of life additions to the website, such as:
- Tags for extensions (easy to add more if desired)
- Added "credit" section for each extension
- Made all extensions a uniform height
- Removed LMS Utils (as requested from LMS)
- Rearrange a few extensions to group together like extensions.

chrome_BV4cWUsZoc


True-Fantom:
@NexusKitten There is one idea that I have been wanting to write about for a long time: To make it possible to add a new or update label to the extension. I don't have the opportunity to make a high-quality example right now, but nevertheless:
IMG_20230602_001022

True-Fantom:
@NexusKitten Thanks!)

True-Fantom:
@NexusKitten Since the merge of this PR is delayed, I can offer a few more ideas:

- Add extension sorting, but according to capabilities.
- Add a square for the introductory text, because it's more beautiful :)
- Change label text color to white, to match the main colors of the banners.

Here's a little crooked photoshop:

Без имени

NexusKitten:
> @NexusKitten Since the merge of this PR is delayed, I can offer a few more ideas...

I may implement some of these ideas in the future (most of them should be relatively simple), however, GarboMuffin himself stated that alone this PR is going to have to take a lot of consideration (it makes sense, considering it affects extensions as a whole rather than just adding one or modifying a few). That being said, I've considered the idea of implementing some of these ideas into a UserScript, so it's not more that Garbo has to work around but still gives us some quality of life features in the library.

NexusKitten:
This isn't really going anywhere.

PR #491, opened 2023-05-30
Camera Controls: Add 2 blocks

LukeManiaStudioss:
These blocks were commented away with 0 functionality, so I programmed them in.

"move camera to (target)" will snap the camera to the position of a sprite target on the stage.
image

"point towards (target)" will point the camera in the direction of the specified target.
image


TheBombyMaster:
This will be extremely useful for things like overworlds and RPG's! Can't wait to see what people make with these when they get added to the extension gallery!

PR #495, opened 2023-06-01
Clones Plus: Allow block colors to be (mostly) compatible with the Scratch Addons "Customizable block colors" addon.

LilyMakesThings:
This slightly unorthodox patch to Clones Plus puts the block colours as the control category colour, therefore allowing it to be in sync with the Scratch Addons "Customizable block colors" addon.

image


Martinelplayz:
very cool

LilyMakesThings:
I imagine there will be a better way to do this soon, so I'm closing this for now.

PR #496, opened 2023-06-01
[new extension] BigInt extension

Skyhigh173:
add a new extension **BigInt**.

about BigInt: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/BigInt

Blocks : 
image

### DIFF
--- a/extensions/Skyhigh173/bigint.js
+++ b/extensions/Skyhigh173/bigint.js
@@ -1,6 +1,5 @@
 (function(Scratch){
   'use strict';
-  /* global Scratch */
 
   const bi = x => {
     try {
--- a/extensions/Skyhigh173/bigint.js
+++ b/extensions/Skyhigh173/bigint.js
@@ -58,17 +58,6 @@
             }
           },
           {
-            opcode: 'as',
-            blockType: Scratch.BlockType.REPORTER,
-            text: 'As BigInt [text]',
-            arguments: {
-              text: {
-                type: Scratch.ArgumentType.STRING,
-                defaultValue: ''
-              }
-            }
-          },
-          {
             opcode: 'to',
             blockType: Scratch.BlockType.REPORTER,
             text: 'To Number [text]',
@@ -379,9 +368,6 @@
     }
     from({ text }) {
       return bi(text);
-    }
-    as({ text }) {
-      return bi(text) + 'n';
     }
     to({ text }) {
       return Number(bi(text));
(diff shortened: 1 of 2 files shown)


Martinelplayz:
whats the difference between the math here and the math already in scratch?

Skyhigh173:
@Martinelplayz 
It's BigInt, you can check the doc link

CST1229:
> whats the difference between the math here and the math already in scratch?

Scratch's number casting (used in the Operator blocks) casts the number to a normal number (instead of a BigInt).

Skyhigh173:
Remember, It's based on **BigInt** ! **Not** original JS math functions!!!

extensions/Skyhigh173/bigint.js
```diff
@@ -0,0 +1,456 @@
...
+
+  class BigIntExtension {
+    getInfo() {
+      return {
+        id: 'skyhigh173BigInt',
+        name: 'BigInt',
+        color1: '#59C093',
+        blocks: [
+          {
+            opcode: 'bigint_from',
```
GarboMuffin, line 33: you can leave it as-is, but you don't *need* to put bigint_ in the opcodes; they are already prefixed with the extension ID so names like "from" are already guaranteed to not conflict with other extensions

extensions/Skyhigh173/bigint.js
```diff
@@ -0,0 +1,456 @@
...
+            text: 'To BigInt [text]',
+            arguments: {
+              text: {
+                type: Scratch.ArgumentType.STRING,
+                defaultValue: ''
+              }
+            }
+          },
+          {
+            opcode: 'bigint_as',
```
GarboMuffin, line 44: from and as should probably just be merged

also, it's worth thinking about what "6.5 to bigint" should do. cutting off the decimals seems like a reasonable solution

GarboMuffin: turns out the other blocks already do truncate decimals, just these don't

extensions/Skyhigh173/bigint.js
```diff
@@ -0,0 +1,456 @@
...
+
+  const bi = x => {
+    try {
+      if (x.slice(-1) == 'n') {
+        return BigInt(x.slice(0, -1));
+      }
+      return BigInt(x);
+    } catch {
+      try {
+        return BigInt(Math.trunc(Number(x)));
```
GarboMuffin, line 12: Might be a better idea to do Math.trunc by using string operations to remove the decimals, otherwise by converting to Number you can lose precision

GarboMuffin:
this is nice, left some comments, can probably get through tomorrow

GarboMuffin:
build errors unrelated

Issue #493, opened 2023-05-31
blocks we need


Drago-Cuven:
#this is just a fair few blocks from snap that would be really cool to have on turbowarp.
![Screenshot 2023-05-31 2 59 49 PM](https://github.com/TurboWarp/extensions/assets/57764393/449f51a2-ecce-4c1e-a6c2-e832bdf89abe)
![Screenshot 2023-05-31 2 48 59 PM](https://github.com/TurboWarp/extensions/assets/57764393/da4189df-7f62-4434-a19f-3047ac31857e)
![Screenshot 2023-05-31 2 42 43 PM](https://github.com/TurboWarp/extensions/assets/57764393/9f32c62b-6198-4e20-8836-1d94b16cdd10)
![Screenshot 2023-05-31 2 41 21 PM](https://github.com/TurboWarp/extensions/assets/57764393/1dda49f8-ffcf-47f9-86d6-0e604cf6ea74)
![Screenshot 2023-05-31 2 40 30 PM](https://github.com/TurboWarp/extensions/assets/57764393/8ad7c7ae-4214-4ba7-8642-6567c7234cc0)

the change tab block could probably be used in the lms utils extension.

LukeManiaStudioss:
"run JS" blocks or anything of that nature has been obviously banned from any extension on the gallery. I don't have much else to say lol

GarboMuffin:
If you think you want to use cookies in a scratch project, you actually probably want our local storage extension instead.

Random strings might exist in the encoding extension, strangely enough

Martinelplayz:
> If you think you want to use cookies in a scratch project, you actually probably want our local storage extension instead.
> 
> Random strings might exist in the encoding extension, strangely enough

ok. well there is one more block we surely need
![Uploading Screenshot 2023-05-31 4.15.21 PM.png…]()

Martinelplayz:
and i saw a pull request for an extension that lets you make better custom blocks at some point. THIS WOULD BE PERFECT FOR IT!!!!!!
![Screenshot 2023-05-31 4 41 28 PM](https://github.com/TurboWarp/extensions/assets/57764393/87daa11d-0e1f-411d-8ab1-bda993a036b1)
like seriously! CUSTOMIZED COLORS?!

Martinelplayz:
@GarboMuffin  can this be done?

LukeManiaStudioss:
short and sweet answer is not for a while

LIZARD-OFFICIAL:
[...]

PR #504, opened 2023-06-02
Add battery extension

GarboMuffin:
https://github.com/TurboWarp/extensions/issues/493#issuecomment-1570967970


LukeManiaStudioss:
The fact this works on laptops too makes me happier than it has any right to.

GarboMuffin:
need to add a note to the description that it "may not work on all devices and browser"

Martinelplayz:
NICE! when is this getting merged?

Martinelplayz:
Battery Extension
Battery Extension

PNG and SVG. My own Thumbnail for this ext

PR #510, opened 2023-06-03
All Menus: Add experimental "All Menus" extension

LilyMakesThings:
'All Menus' creates a category in the block palette with every menu of every Scratch category as well as loaded extensions.
image

Some notes about this extension:
- If the user has no other extensions loaded, the blocks from this extension are 100% compatible with Scratch.
- The extension won't save to projects because you can't technically drag extension-created blocks from the extension, as there aren't any.
- As a result of the previous statement, changes made to this extension won't be "breaking" changes. This can be deleted at any time.


SurvExe1Pc:
i like this idea

LilyMakesThings:
In theory the "refresh blocks" button could be axed, this refreshes automatically whenever an extension is loaded anyway.

ClaytonTDM:
image
im having too much fun with this

LilyMakesThings:
This extension has had a full rewrite, it no longer uses icky injection code.

ClaytonTDM:
HELL YEAH

horizontal-shading:
Thank you Lily for adding this

PR #512, opened 2023-06-03
More Timers: Add More Timers

LukeManiaStudioss:
More Timers adds blocks to manipulate and control several Scratch timers at once.
image


images/README.md
```diff
@@ -195,3 +195,7 @@ All images in this folder are licensed under the [GNU General Public License ver
 
 ## battery.svg
  - Created by [@Martinelplayz](https://github.com/Martinelplayz) in https://github.com/TurboWarp/extensions/pull/504#issuecomment-1574243161
+
+## LukeManiaStudios/MoreTimers.svg
+- Created by
```
GarboMuffin, line 200: by who

website/index.ejs
```diff
@@ -667,6 +667,12 @@
           <h3>McUtils</h3>
           <p>Helpful utilities for any fast food employee. Created by <a href="https://scratch.mit.edu/users/LukeManiaStudios/">LukeManiaStudios</a>.</p>
         </div>
+
+        <div class="extension">
```
GarboMuffin, line 671: put it before Cloudlink

GarboMuffin:
One thing I like from #552 is the lack of a distinction between start timer and reset timer. Similarly, in my mind, running "set timer to 10", regardless of whether that timer was started previously, should just set that timer to 10, etc.

LukeManiaStudioss:
> One thing I like from #552 is the lack of a distinction between start timer and reset timer.

This is something I 100% agree with, because they both almost work the same anyway. However, I do not know what a merged version of the block would be called aside from "start/reset timer", which seems clunky.

> Similarly, in my mind, running "set timer to 10", regardless of whether that timer was started previously, should just set that timer to 10, etc.

Also agree with this.

LukeManiaStudioss:
I don't have a better name for the block than this. We can always change it later.
image

PR #519, opened 2023-06-04
LMS Utilities: Adjustments

LukeManiaStudioss:
- Add toggle to show legacy blocks
- Optimise the code to make it more readable

### DIFF
--- a/extensions/LukeManiaStudios/lmsutils.js
+++ b/extensions/LukeManiaStudios/lmsutils.js
@@ -1143,6 +1143,7 @@
       if (dropdown === 'number') return !isNaN(input);
       if (dropdown === 'uppercase') return (input == input.toUpperCase());
       if (dropdown === 'lowercase') return (input == input.toLowerCase());
+      return false;
     }
 
     encodeToBlock(args) {
@@ -1153,6 +1154,7 @@
           return char.charCodeAt(0).toString(2);
         }).join(' ');
       }
+      return '';
     }
 
     decodeFromBlock(args) {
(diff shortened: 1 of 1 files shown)


extensions/LukeManiaStudios/lmsutils.js
```diff
@@ -1399,65 +1280,33 @@
...
 
     screenReporter(args) {
-      if (args.DROPDOWN === 'width') {
-        return screen.width;
-      }
-      if (args.DROPDOWN === 'height') {
-        return screen.height;
-      }
+      if (args.DROPDOWN === 'width') return screen.width;
+      if (args.DROPDOWN === 'height') return screen.height;
```
GarboMuffin, line 1284: a detail that should've been caught earlier: this and the block below need to return *something* when args.DROPDOWN is none of those

extensions/LukeManiaStudios/lmsutils.js
```diff
@@ -1384,6 +1264,7 @@
 
     readClipboard(args) {
       if (navigator.clipboard && navigator.clipboard.readText) {
+        // @ts-expect-error - not typed yet
```
GarboMuffin, line 1267:
```suggestion

```

extensions/LukeManiaStudios/lmsutils.js
```diff
@@ -2,6 +2,8 @@
...
 
+  let hideLegacyBlocks = true;
+
   var vars = {};
   vars['variables'] = {};
```
GarboMuffin, line 8:
```suggestion
  vars['variables'] = Object.create(null);

```

PR #523, opened 2023-06-04
Add all extension images

Dy-tech:
@GarboMuffin Here are the images COMPLETED.


LukeManiaStudioss:
Extension thumbnails need to be 600x300 SVGs. Furthermore it's preferred that you don't put the name of the extension in the thumbnail, since you only need to look an inch down to see it.

images/rixxyx.svg
GrahamSH-LLK, line 1: idk what this is but it's ugly

LukeManiaStudioss: It's meant to be, that's the joke

website/index.ejs
```diff
@@ -573,7 +573,7 @@
         </div>
 
         <div class="extension">
-          <%- banner('Longboost/color_channels') %>
+          <%- banner('color_channels') %>
```
GarboMuffin, line 576: Doing this breaks the links on the website, by the way

GarboMuffin:
I will do this on my own

PR #525, opened 2023-06-04
Temporary Variables: Overhaul [Add 2 Variable Types]

LukeManiaStudioss:
Added 2 variable types, Thread Variables and Runtime Variables:
- Runtime variables last for as long as the project is running before being cleared.
- Thread variables are attached to the branch they're in.

image

(new thumbnail needed)


GarboMuffin:
Since it seems you have hidden *all* of the old blocks, it would probably make more sense to just make a new extension, new extension id, etc.

LukeManiaStudioss:
Alright, I'll do that now

PR #537, opened 2023-06-06
Update box2d.js to add gravity reporters

David-Orangemoon:


GarboMuffin:
you should try to find a way to make whatever git client you use not do this if possible

image

PR #543, opened 2023-06-07
Add color extension

samuellouf:
I created an extension named "Colors +".
I will probably add some new blocks soon.
Screenshot of the Colors + extension


LukeManiaStudioss:
I don't understand the utility behind the RGB and HSL blocks; they return a value that isn't recognised by other extensions and can't easily have its values extracted other than just having the values on its own, which you probably would have anyway if you're able to input it into the extension.

I understand they're used by the final block, but then the output of that block is also sometimes arbitrary for the same reason (it also has an error case when you input HSL to RGB with a hex default value, I've not tested for other error cases though).

Furthermore, 2 of the other blocks are already in other extensions.

GarboMuffin:
considering this is an extension specifically about colors, having some overlap with other extensions is completely reasonable and probably a good thing honestly

>  I will probably add some new blocks soon.

please do

LukeManiaStudioss:
> considering this is an extension specifically about colors, having some overlap with other extensions is completely reasonable and probably a good thing honestly
The overlap isn't a problem on paper, I only see it as an issue because the extension in its current state doesn't serve much utility. I do like the idea of these blocks being in one place.

LukeManiaStudioss:
Also the name should just be "Colors" or something, as opposed to "Color +", which implies that it's an expansion of Scratch category called "color"

Skyhigh173:
I can make my own version of Color Extension too

LukeManiaStudioss:
> I can make my own version of Color Extension too

I would like to see that, there's a color extension on PenguinMod, perhaps use that as a reference

Skyhigh173:
> > I can make my own version of Color Extension too
> 
> 
> 
> I would like to see that, there's a color extension on PenguinMod, perhaps use that as a reference

yep m2

samuellouf:
Ok, I'm going to edit my extension.

GarboMuffin:
we can merge this in its current state and add more blocks later, if you want

PR #552, opened 2023-06-09
'More timer' extension

MA-master-sec:
Yes,just like the code.


GarboMuffin:
this is likely made redundant by https://github.com/TurboWarp/extensions/pull/512

other review comments, although I don't anticipate this being merged as its functionality will exist elsewhere:

 - the file is in the wrong spot -- it should be in a folder in extensions
 - extension ID needs to be prefixed with eg. your username
 - would need to use Object.create(null) instead of {} for the timer dictionary to avoid names like toString causing problems

GarboMuffin:
made redundant by https://github.com/TurboWarp/extensions/pull/512

PR #553, opened 2023-06-09
Discord API: Add Discord API extension

LilyMakesThings:
Discord API is an extension with 2 parts at the moment. 1 part of it is the 1 block you'll see when opening the extension for the first time, the other is the experimental mode which shows exactly what I wanted the extension to be.
image
image
image

I did this because I'm currently trying to find a way to save Discord Webhooks/Embeds that are created in this extension like variables. In various testing environments I have gotten as far as reverse engineering how blockly generates random UIDs and have created my own, functional, custom variables that saves to projects. However, the VM currently wipes the IDs back to being Scalar, making it useless.

I've also tried using comments, however they do not work packaged for obvious reasons.

I will move the experimental blocks out of their current area of the extension when there is a stable way of saving extension data to a project. The ideal method would be custom variable types, though I'll doubt we'll see those any time soon.


LucasXD5:
hi

NexusKitten:
hi

LilyMakesThings:
This needs far more work, as I don't imagine custom variables types/a way for extensions to save project data will ever be available within the next 6 months, and that's fine

GarboMuffin:
Is the lack of easy persistent storage the only issue?

LilyMakesThings:
> Is the lack of easy persistent storage the only issue?

100% yes

GarboMuffin:
let's leave it open then

IgosProjects:
HOW DO I USE IT

PR #558, opened 2023-06-10
FIX NGIO (pls don't break it this time)

David-Orangemoon:


GarboMuffin:
Do you know which part was actually broken? because the diff is not readable

David-Orangemoon:
> Do you know which part was actually broken? because the diff is not readable

The login portion had the userdat.login removed so it would say the user was not logged in.

PR #561, opened 2023-06-11
mdwalters/notifications: Partially fix Android support

GarboMuffin:
Chrome-based browsers on Android, for unknown reasons, can only display notifications through service workers, which does not make any sense.

Luckily our online editor does have a service worker, so we are able to piggy-back off of it. Projects generated by the packager are not quite as lucky.


GarboMuffin:
Linting errors unrelated

PR #563, opened 2023-06-11
Search Parameter API

ZXMushroom63:
Search Parameter API is an extension I made to allow modification of search parameters in the url,
eg: x = 15 would be
https://example.com/?x=15
So far the extension has 7 blocks, each modifying or reading search params. I want to use this extension because when a project is packaged, the search params can be used as a config, eg:
https://example.com/packaged.html?mode=40
image


GarboMuffin:
Would it make sense to use history.replaceState instead of location.search = to avoid refreshing the page?

ZXMushroom63:
Yep! I did not think of that. Will update.

SurvExe1Pc:
also will there be support for multiple duplicate tags like the extension one in turbowarp (can be used multiple times)?

ZXMushroom63:
I am not quite sure what you mean, but if you are talking about how you can go to
https://turbowarp.org/editor?extension=some_extension_url_here
and it will load the extension from the URL, no, search parameters only let you have one value per key. It is part of their core design.

CST1229:
> no, search parameters only let you have one value per key. It is part of their core design.

False; this works: https://turbowarp.org/editor?extension=https://extensions.turbowarp.org/lab/text.js&extension=https://extensions.turbowarp.org/stretch.js

SurvExe1Pc:
> no, search parameters only let you have one value per key. It is part of their core design.


> > no, search parameters only let you have one value per key. It is part of their core design.
> 
> False; this works: https://turbowarp.org/editor?extension=https://extensions.turbowarp.org/lab/text.js&extension=https://extensions.turbowarp.org/stretch.js

yeah it does support multiple, but i can depend on how to do it, you could use regex(s), the URLSearchParams api and some others, 
for URLSearchParams you could just loop over all the keys+values as using the `get()` method only returns the first occurrence,
example from [MDN](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams):
`
for (const [key, value] of mySearchParams.entries()) {
}
`

ZXMushroom63:
Oops!
I will implement this as soon as possible.
Thanks for bringing this up!

ZXMushroom63:
While I'm at it, do you think this extension is a good place for adding protocol handlers?
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler/Web-based_protocol_handlers

GarboMuffin:
separate extension

ZXMushroom63:
got it!

PR #565, opened 2023-06-11
Create timer.js

MA-master-sec:
A easier way to use more timer in your project.


GarboMuffin:
you already opened https://github.com/TurboWarp/extensions/pull/552; you can edit PRs you don't need to make a new one

PR #570, opened 2023-06-12
JavaScript: Add JavaScript extension

LilyMakesThings:
# Before you comment/reject, read the full pull request
I understand the documentation states that this extension should not exist. I also understand _why_ it shouldn't exist. But please hear me out with this extension.

It doesn't take a rocket scientist to understand just how dangerous giving people eval blocks could be. It's for that reason why PenguinMod recently sandboxed their JavaScript blocks, and why they've never been allowed on TurboWarp (unless you are aware of the existence of the now obsolete TurboLoader). Despite this, it's indisputable that people who make games with TurboWarp, like myself, could easily benefit from having this at our disposal. It's for that reason I created this extension, however it has several precautions in place to avoid it being used for anything malicious:

- In order for the blocks to run, you must enable them by clicking a button within the Toolbox. Clicking the button displays a warning over how dangerous the blocks are. 
imageimage

- As a way to not confuse users who run the extension and are confused why the blocks don't work, the GUI will tell you when JavaScript is disabled.
image

- JavaScript can be disabled again at any time.
image

This is the best compromise for a JavaScript extension. It's impossible to enable JavaScript in the editor using the existing blocks, and the user is encouraged _not_ to use it as long as they're aware of what the blocks do.

### Why does this extension need to exist?

As I said, developers would benefit from having this. The packager allows you to run custom JS anyway, so you have just as much of a chance getting phished by a packaged TurboWarp project with or without this extension. (Packaged projects can also run custom, proprietary extensions unsandboxed. This is a trivial extension for someone to make). It's for this reason why the blocks are enabled by default in the packager.

This extension would allow for developers to not only use their own JS, as the packager already allows them to, but they'd also be able to test the code's functionality on the fly, which is a major bonus for the workflow of the project.

### The Blocks:
image

(it goes without saying, this should not be on the gallery)


Martinelplayz:
should i keep this one out of the collection?

Martinelplayz:
i can already see this being really useful. i'm testing the extension now, and having fun with it. it does solve my tab title changing issue too.

GarboMuffin:
> it does solve my tab title changing issue too.

a far better solution is for someone to submit an extension with a "set title" block

Martinelplayz:
> > it does solve my tab title changing issue too.
> 
> a far better solution is for someone to submit an extension with a "set title" block

very true, and someone is already making one.

Martinelplayz:
JsThumbnail
JsThumbnail
i know that its not going to the gallery but if there is ever a "hidden gallery", it makes it somehow, or that "show all extensions" button suggestion comes through this could possibly be of use.

TheShovel:
> > it does solve my tab title changing issue, too.
> 
> A far better solution is for someone to submit an extension with a "set title" block

It's quite annoying to have to load 50 billion extensions through. Having a block to run a quick script instead of fetching an entite extension to do it is much more convenient.

LilyMakesThings:
> > > it does solve my tab title changing issue, too.
> > 
> > 
> > A far better solution is for someone to submit an extension with a "set title" block
> 
> It's quite annoying to have to load 50 billion extensions through. Having a block to run a quick script instead of fetching an entite extension to do it is much more convenient.

You're all discussing this as a solution to small extensions, I just see this as a development tool for making them

Martinelplayz:
i'm still loving this. it's fun to experiment with js.  hope this gets merged soon.

CST1229:
Maybe JavaScript could only automatically be enabled if unsafe cloud variable behaviors are enabled (`window.scaffolding && window.scaffolding.cloudManager.overrides.has("☁ eval")`), since the eval cloud variable already exists there. (Not when the project is packaged with the option off, though.)
Also, `runtime.isPackaged` only returns true if the packaged runtime mode ("Remove raw asset data after loading to save RAM") is enabled, so you probably shouldn't use it to detect the packager.

extensions/LukeManiaStudios/JavaScript.js
```diff
@@ -0,0 +1,129 @@
...
+
+  let ineditor = false;
+
+  // Credit to TheShovel for the GUI element.
+  const para = document.createElement("p");
+  if (ineditor == false) {
+    try {
+      scaffolding;
+    } catch (error) {
+      para.innerText = "JavaScript: JavaScript is currently disabled.";
```
GarboMuffin, line 21: eslint is just being silly on this one, ignore it

BlueDome77:
Could it get into the gallery if it had a request permission popup like notifications and network?

tigercoding56:
> # Before you comment/reject, read the full pull request
> 
> I understand the documentation states that this extension should not exist. I also understand _why_ it shouldn't exist. But please hear me out with this extension.
> 
it should exist 
we need it 
, as this can be used
to implement 
image decoding

GarboMuffin:
I would consider this if I heard a good reason for it to be added now that the extension library continues to grow (if slowly) and custom extensions being not hard to use

Some other mods have JS evaluation blocks -- how are they being used there?

LilyMakesThings:
> Some other mods have JS evaluation blocks -- how are they being used there?

To do stuff like math evaluation and edge-case things that extensions can't do

PR #585, opened 2023-06-15
Part 2 of LilyMakesThings rename

GarboMuffin:
Renaming extension files creates permanent bloat -- it's not something we're going to do often.


LilyMakesThings:
> @LilyMakesThings I assume you want LilyMakesThings instead of LukeManiaStudios for your URLs, yes? This is a one-time thing, so try to get it right, I don't want to do it again (every rename creates a lot of bloat)

Yes, that would be wonderful

LilyMakesThings:
Changing it to just "lily" would almost future proof it, but part of me wonders if that's too generic

GarboMuffin:
first come first  serve, why not

LilyMakesThings:
> first come first serve, why not
> 
> do you want it?

yes please!

LilyMakesThings:
Will this hold up on desktop/desktop beta?

SurvExe1Pc:
wouldn't it be easier to just move all the request to the "/LukeManiaStudios" path to the new path instead of manually defining every extension, basically like Forwarding with masking?

LilyMakesThings:
> wouldn't it be easier to just move all the request to the "/LukeManiaStudios" path to the new path instead of manually defining every extension, basically like Forwarding with masking?

this option allows the folder to be renamed, which is honestly more comfortable

PR #605, opened 2023-06-18
Cast: Add Cast

LilyMakesThings:
**Requires VM changes: https://github.com/TurboWarp/scratch-vm/pull/145**

Simple extension to use Scratch.Cast to convert inputs to strings, numbers, or booleans.
image

You can also get what type of value an input is.
image

Bonus: it works as a coupler for values ༼ つ ◕_◕ ༽つ
image


SurvExe1Pc:
how did you get the reporter in the boolean input???
also this seems very useful

LilyMakesThings:
Ok I'm done now :)

SurvExe1Pc:
> Ok I'm done now :)

lol, also what does the "default" option do?

> Bonus: it works as a coupler for values ༼ つ ◕_◕ ༽つ
HOWW, its a reporter thoe. I am so confused

LilyMakesThings:
>lol, also what does the "default" option do?


default just doesn't cast the value, which allows it to function as a coupler :)

SurvExe1Pc:
> > lol, also what does the "default" option do?
> 
> default just doesn't cast the value, which allows it to function as a coupler :)

BUT HOW???????

SurvExe1Pc:
i am going to have to lookup what the different between switch and if else does.

CST1229:
> how did you get the reporter in the boolean input???

Set the `output` of the block to null in getInfo.

LilyMakesThings:
> > > lol, also what does the "default" option do?
> > 
> > 
> > default just doesn't cast the value, which allows it to function as a coupler :)
> 
> But howw???????

When defining the block, set "isDynamic" to true, and "output" to null (in that order)!

Keep in mind, this method of achieving this does have the side effect of making any menu with "acceptReporters: false" not show up on the block, but it is the simplest method.

LilyMakesThings:
Optional Thumbnail:
image

LilyMakesThings:
Note: Just found this would also be ideal

SurvExe1Pc:
> Note: Just found this would also be ideal

oml, so no more trickery lol

GarboMuffin:
thanks for your patience

GarboMuffin:
allowDropAnywhere is not live yet, but when it is, that part of the extension will just work

PR #606, opened 2023-06-18
More Sensors

G1nX-01:
okay so
this extension is able to get the user's language code, as well as if they have headphones.
it has 2 time-related blocks. the first one will transform a date into a relative timestamp ("1 day ago" or "in 77 days")
and the other will make a real-time countdown based on your timezone for any future date.
that's all. i plan to add an icon if i can.

### DIFF
--- a/extensions/Lily/LooksPlus.js
+++ b/extensions/Lily/LooksPlus.js
@@ -527,17 +527,20 @@
       const targets = Scratch.vm.runtime.targets;
       const myself = Scratch.vm.runtime.getEditingTarget().getName();
       for (let index = 1; index < targets.length; index++) {
-        const targetName = targets[index].getName();
-        if (targetName === myself) {
-          spriteNames.unshift({
-            text: 'this sprite',
-            value: targetName
-          });
-        } else {
-          spriteNames.push({
-            text: targetName,
-            value: targetName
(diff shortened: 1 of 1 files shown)


LilyMakesThings:
I like the utility of this extension, especially being able to detect language codes as well as if the user has headphones! However, there is an existing extension called Sensing+, is there any reason why these blocks can't just go into that extension?

SurvExe1Pc:
a year is 365.256363004 days or shortened to 365.24 days because leap year, does it account for that or?

G1nX-01:
> I like the utility of this extension, especially being able to detect language codes as well as if the user has headphones! However, there is an existing extension called Sensing+, is there any reason why these blocks can't just go into that extension?

ok

G1nX-01:
> a year is 365.256363004 days or shortened to 365.24 days because leap year, does it account for that or?

Oh yea i don't really know. I'll have to check.

SurvExe1Pc:
> > a year is 365.256363004 days or shortened to 365.24 days because leap year, does it account for that or?
> 
> Oh yea i don't really know. I'll have to check.

some math to show it, https://www.wolframalpha.com/input?i=floor%28%28365.24*4%29%2F365%29

PR #614, opened 2023-06-19
Add Browser Utilities Extension

NexusKitten:
A handful of blocks for receiving and setting browser information.

image
image

It should go without saying that most if not all of these blocks will not function when the project is packaged into an exe file.


LilyMakesThings:
4 of these blocks exist within other extensions; in network (dark green) they're far more precise. Consider revising the existing blocks to at least match the functionality.

image

LilyMakesThings:
It's nice to finally see alert, confirm and prompt have a home outside of LMS Utils :') only a few more blocks until that extension is finally obsolete! woohoo!

GarboMuffin:
I think I will make isonline just check navigator.online. It's not great, but it doesn't involve making requests to a website that many schools block for one reason or another

GarboMuffin:
this would probably be not very affected by #674

GarboMuffin:
i wonder if we just rename navigator to browser utilities and put everything from this in there

NexusKitten:
> i wonder if we just rename navigator to browser utilities and put everything from this in there

Possibly, I feel like some of the blocks in network would fit under the umbrella of "Browser Utilities" (specifically the ones below), however I feel like the get/post blocks and blocks about network specs are out there enough to warrant their own extension. It's ultimately up to you, however I'd rather just have them next to each other and move some blocks over.
image

NexusKitten:
> 

If this message makes no sense, I'm sorry. I have trouble remembering the differences between Network and Navigator; they have very similar names and functions. You can merge Browser Utilities and Navigator

NexusKitten:
@GarboMuffin I've made several changes, most notably merging this extension with Navigator as you suggested.

This extension is now ready to merge.

Eliaspierce (approved):

NexusKitten:
Updated to include new blocks from #759

NexusKitten:
@GarboMuffin any reason lint isn't running for me?

LilyMakesThings:
> @GarboMuffin any reason lint isn't running for me?

You have conflicts

Issue #611, opened 2023-06-19
Extensions : "Looks+", "Sensing+", "Files" update suggestion


XeroName:
My environment : TurboWarp Desktop v1.8.0 beta-1.

Hello.
The custom extensions recently added in TurboWarp seems very powerful and having high-level potential.

According to "Scratch built-in timer" benchmark result, it seems quite a bit slow rather than basic blocks, but also allows tons of useful functions developing Scratch-based game just like my case.

But using some of the extensions, I got a few requirements/dissatisfaction about lack of functional process.
So here's the few suggestions about the extensions : **"Looks+", "Sensing+", "Files".**

**In Detail :** 

**1. Functional Error at "Looks+" - The "[effect] effect of (sprite)" block doesn't work with clones.**
When I set the (sprite) tab of block with "this sprite" and use it on Clones, it gets the original sprite's effect values and doesn't recognize parent of itself(clone). Opening the (sprite) tab when clone got birth, just increasing of "this sprite" button selection.
![image](https://github.com/TurboWarp/extensions/assets/63216951/cb4e1564-1661-437d-926d-d1ba16db8b0c)
What a mirror.

**2. EXACTLY Mouse pointer detector at "Sensing+"**
Sensing+ is very useful to add support of mobile touchscreen. But in some of the cases like situation have to support both PC and Mobile Screen, detecting which is exactly mouse pointer goes too difficult. Yeah I know how to detect exactly mouse pointer, not a touchscreen touch, but it is too abstruse and get mess the codes. Basic Scratch's mouse pointer detection is very annoying in this situation.
Adding blocks such as <is Exactly Mouse pointer?> will be very useful to recognizing at computer mouse pointer.

**3 . "Files" : Disable showing "Drag and Drop" box screen when open files / Set a title of file saving window**
Although there's already have an block "set open file selector mode to (open selector immediately)", after using that block, it pops up both file opening window and D&D box at the same time! I hope there be an option disable showing D&D box, to only show file opening window.

Also, when I trying to save any file using "download (string) as (file.txt)" block, the title of file saving window displays raw blob names : such as "blob:file:///12blah9123-blah5345-bl75675ah".
![image](https://github.com/TurboWarp/extensions/assets/63216951/17e6296a-e9e5-4f68-b502-3f18bd257474)
Giving an option to edit it, it will be really cool.

these above are all of my suggestions.
Thank you for reading, and always huge thanks and cheers for GarboMuffin and TurboWarp's developers.

XeroName

[...]

PR #618, opened 2023-06-20
Lily/LooksPlus: fix "this sprite" in clones

GarboMuffin:
https://github.com/TurboWarp/extensions/issues/611


GarboMuffin:
in hindsight it probably would've made sense to keep the menus working the same but add another option for eg. "_myself_", but oh well, we're stuck with this now

PR #622, opened 2023-06-22
Add Deltatime.js

XeroName:
Hello, I would like to ask if I can add Unsandboxed Extension at TurboWarp.

Extension description : A soft, precision Delta Timing blocks with adjustable filter.
Credit : whole code and cover/icon image made by myself.

Thank you.


XeroName:
DT_ExtensionPreview
This is the preview image of Deltatime.
Have no accessibility to /images/ folder, so I have no idea how to add cover image.

XeroName:
Sorry I forgot uploading screenshot of blocks.
Here's the screenshot when Deltatime extension loaded :
image

LilyMakesThings:
God-tier extension, not gonna lie, this solves a lot of my problems.

XeroName:
> 

So much appreciate it.

Recently, I started using quite of extensions to my game, and most of that are yours such as "Temp Vars" and "Comment blocks".
I'm really happy with those cool functions and would like to say thank you!

XeroName:
Oops, seems I forgot some of process before to merge extension.

I've just edited images/README.md and website/index.ejs files.
Sorry for about that and would be thankful to point out if I still missed something. 😀

GarboMuffin:
thank you for your patience, will try to get this in tonight

XeroName:
Much appreciated it! 🎉

GarboMuffin:
What is the point of the getTimeRes block?

XeroName:
> What is the point of the getTimeRes block?

Based on my experience, "Deltatime" and "FPS" block used at many cases of with operator blocks, so I thought those blocks are needs to be short length and easy draggable to user's convenience.

That's why Dt and FPS blocks are separated, while getTimeRes block is for users who want to change target resource easy and instantly.
Functionally works same with Dt and FPS blocks.

Seems that getTimeRes block is unnecessary, so willing to hide or remove that block if anyone suggests.

XeroName:
image
This is the screenshot of Deltatime v1.2.2

True-Fantom:
Hello! I have recently returned and I want to share my impressions about the new extensions :)

I really like the idea of expansion, considering how laggy TW can be, it will be very useful, especially with the box 2d extension. But I also found the cons. I have several suggestions and one bug found.

I'll start with a bug: dt cannot be greater than 1 (or immediately equal to infinity). if the FPS of the project is equal to 1, given the TW prolags, the project will still run slower than at high fps. And since you can set FPS less than 1, it's really important. Also, at low values, dt clearly loses accuracy, which is also a problem. I do not know if it is possible to change everything, because I still do not have time to dig into the code, but in theory it is definitely possible.

Next, the main offer! Add a dt limit. This mechanics is in Construct 2, where the maximum value of dt is 0.2. This is done so that objects dependent on it do not teleport through walls and so on. Setting a fixed limit would be unpleasant, but adding the ability to set it at will would be nice. There will be 2 blocks in total: set dt limit & dt limit. By default, it can be set to infinity, 1 or 0.2 (you will need to check which will be more convenient).

This suggestion and bugfix will also solve the problem with the "FPS > 0" block, which will no longer be needed.

Well, and more: I really don't see the need for separate blocks for the filter and calculator. Yes, they help to use dt, but in my opinion they still clog up the palette more, because they are VERY easy to recreate. If I were you, I would still remove them.

True-Fantom:
I realized what the problem was! This line is to blame: `vmFPS = Math.round(1000 / frameTime);`
If you remove the rounding, dt will work stably. In this case, the FPS will also not be whole, but I think that this was originally supposed to be the case, if you need to round them, it's better to do it manually.

XeroName:
> I realized what the problem was! This line is to blame: `vmFPS = Math.round(1000 / frameTime);` If you remove the rounding, dt will work stably. In this case, the FPS will also not be whole, but I think that this was originally supposed to be the case, if you need to round them, it's better to do it manually, instead of additional extension editing.

The reason I round vmFPS was because of the constantly fluctuating noise.
That noise is not a problem that can be easily solved with a filter. It occurs problem that delays update time as the FPS stability obtained through the filter arises, although it doesn't always perfectly flatten the FPS.

This can be a big problem for my original purpose - multiplayer game, which requires FPS accuracy and stability with fast update times at the same time.

So the easiest solution was to round up the FPS.
This allows FPS and dt to be stable even at low filter values.

But the main problem, as you mentioned, it can't simulate Ultralow FPS like FPS<1 situation.
Agree with your point but I think this is an extreme case example.
Even data loading at "The Mast" was simulated at 3~4FPS without hardware acceleration at my computer.

Would not mind if you want to recreate the extension, but kindly note this in your mind : 
This FPS simulator is not perfectly accurate even raw value and I choose stability at norminal purpose instead of loosing extra-low FPS simulation.

True-Fantom:
Is this only about FPS? If so, we could use a similar hybrid:


    vm.runtime._step = function () {
      oldStep.call(this);
      const now = performance.now();
      let thisFrameTime = now - lastLoop;
      frameTime += (thisFrameTime - frameTime) / filtStren;
  
      vmFPS = Math.round(1000 / frameTime);
      vmDt = 1 / (1000 / frameTime);
  
      lastLoop = now;
    };

XeroName:
Not only for fps, also dt.
If you want to get raw value of resources, it seems good to reprogram it to provide both round and raw value.

True-Fantom:
Okay, then I'll try to do it today!

IonMakesGarbage:
let deltatiming work with turbomode. that would let things like intensive 3d raycasters (like minecraft ones that usually require turbomode) run at a reasonable speed without the player zooming around

XeroName:
> let deltatiming work with turbomode. that would let things like intensive 3d raycasters (like minecraft ones that usually require turbomode) run at a reasonable speed without the player zooming around

Turbo Mode is a function that allows blocks to operate without being limited by ScratchVM's steps. It has no relationship with FPS of ScratchVM.

XeroName:
image
This is a screenshot of v2.2.0.
Now it can simulates FPS as minimum >= 0.005.

Also easier use at Dark mode.
image

Azulcat:
Hello, how can I test this live-saving block?

XeroName:
> Hello, how can I test this live-saving block?

Hello.
You can download raw file under below instruction :
1. Move to "Files Changed" tab.
2. Click "..." button located upper right on code.
3. Click "View file".
4. Hit the "Download raw file" button and download it.
5. Try to load it on TurboWarp with unsandboxing and it will work.

XeroName:
This would be a final commit of this extension.
Also functional explanation at Deltatime doc is finished, which can be updated without changing code.

Some of the functions in documenation such as **Stability Evaluator** and **Autopilot** are just a concept, would not be updated nearby future because those requires more complicated algorithm.

Azulcat:
that's what i did but no work ;(

Azulcat:
nvm it worked!

Azulcat:
@XeroName I might be using it the wrong way but a lot of people who told me how to use delta time they show me a pic and they did not even try it.. can you please explain this wonderful block in a picture? that will help me a lott!

Azulcat:
and make sure that u tested it

extensions/XeroName/Deltatime.js
```diff
@@ -0,0 +1,223 @@
...
+    }
+    last = now;
+  };
+
+  class Dt {
+    getInfo() {
+      return {
+        id: 'dtbyxeroname',
+        name: 'Deltatime',
+        docsURI: 'https://xeroname.github.io/dtbyxeroname/',
```
GarboMuffin, line 59: we want all documentation to be on extensions.turbowarp.org now. for now just put your HTML file in `website` and use a URL like extensions.turbowarp.org/XeroName/deltatime.html. see the other files in there for example

extensions/XeroName/Deltatime.js
```diff
@@ -0,0 +1,223 @@
...
+
+  if (!Scratch.extensions.unsandboxed) {
+    throw new Error('Deltatime by XeroName : This extension must run unsandboxed');
+  }
+
+  const vm = Scratch.vm;
+  const cast = Scratch.Cast;
+
+  let maxFiltStren = 907; // Max value of Filter Strength
```
GarboMuffin, line 25: why 907?

extensions/XeroName/Deltatime.js
```diff
@@ -0,0 +1,223 @@
...
+                type: Scratch.ArgumentType.STRING,
+                menu: 'type_menu'
+              },
+
+              TARGET: {
+                type: Scratch.ArgumentType.STRING,
+                menu: 'target_menu'
+              }
+            },
+            hideFromPalette: true
```
GarboMuffin, line 90: if no projects use this then we can remove it, same for next block

extensions/XeroName/Deltatime.js
```diff
@@ -0,0 +1,223 @@
...
+          {
+            opcode: 'getCalculatorStandard',
+            blockType: Scratch.BlockType.REPORTER,
+            text: 'Standard FPS of Calculator',
+            disableMonitor: true
+          },
+          {
+            opcode: 'calcMultiplyValue',
+            blockType: Scratch.BlockType.REPORTER,
+            text: 'Calc [DISTANCE]',
```
GarboMuffin, line 142: is there a better way to name these?

extensions/XeroName/Deltatime.js
XeroName: Sorry, that was a temporary value to test extension and I forget about it.
Now the max value will be Number.MAX_VALUE.

extensions/XeroName/Deltatime.js
```diff
@@ -0,0 +1,223 @@
...
+
+  let maxFiltStren = 907; // Max value of Filter Strength
+  let filtStren = 8; // Default value of Filter Strength = 8
+  let calcStandFPS = 30; // Standard value of Calculator
+
+  let fT = 0, last = performance.now(), rawFPS, vmFPS, vmDt;
+  const oldStep = vm.runtime._step;
+
+  vm.runtime._step = function () {
+    oldStep.call(this);
```
GarboMuffin, line 33: would it make sense to calculate delta time *before* the frame, not after?

extensions/XeroName/Deltatime.js
XeroName: I will remove all of the hidefrompalette blocks.

extensions/XeroName/Deltatime.js
XeroName: Now that block will be named as "**Value to Move [DISTANCE] with ΔT**".

extensions/XeroName/Deltatime.js
XeroName: Would it be only allowed as HTML file for now?
Or can I wait for Markdown supports?

XeroName:
> @XeroName I might be using it the wrong way but a lot of people who told me how to use delta time they show me a pic and they did not even try it.. can you please explain this wonderful block in a picture? that will help me a lott!

Deltatime is commonly used to fairly compensate for differences in computer performance.

As FPS increases, the ΔT value will be decreases as opposite and same at vice-versa.
You can use it just multiplying ΔT value to originally purposed value to move/rotate sprite.

This video may be helpful : https://youtu.be/rWtfClpWSb8

extensions/XeroName/Deltatime.js
XeroName: May I have a time? I'll find a way to fix it.

XeroName:
```javascript
let fT = 0, last = performance.now(), rawFPS, vmFPS, vmDt;
  const oldStep = vm.runtime._step;

  vm.runtime._step = function () {
    const now = performance.now();
    let thisFT = now - last;
    fT += (thisFT - fT) / filtStren;

    rawFPS = 1000 / fT;

    if (rawFPS < 0.5) {
      vmFPS = rawFPS.toFixed(2);
    } else {
      vmFPS = Math.round(rawFPS);
    }

    if (vmFPS > 0) {
      vmDt = 1 / vmFPS; // At normal situation
    } else {
      vmDt = 0; // Prevent situation of "DIV/0"
    }
    last = now;

    oldStep.call(this);
  };
```
now would it be make sense? or am I still misleading your point?
always thanks for your kindness.

Azulcat:
thx

True-Fantom:
@XeroName Hello again! I hope you haven't forgotten about me in all this time :)

Unfortunately, I did not manage to solve that very problem in one day, and then many "distracting" events followed that did not allow me to write earlier. Anyway, during all this time I gradually developed the extension and as a result I got a completely new version unlike either yours or my old one. Here is a screenshot:

Снимок экрана (1318)

But despite all this, I could not solve the original problem. I have seen your solution, but to be honest, it seems to me very crutchy, and can cause a lot of misunderstandings among users who do not read the documentation. I thought and experimented for a long time and eventually came to the conclusion that it is impossible to make a filter that will carefully preserve the original values by simply feeding them in a different form. Moreover, I don't see the point in it. After conducting some experiments, I realized that the "noise" for which the filter is needed perfectly fits with the actual frame changes, giving failures only due to the limitations of the length of the number in JS. And when using the filter itself, on the contrary, strong errors appear, which largely depend on randomness.

I'm writing about this because I think I'm missing something. Does the filter have any really useful properties that make it necessary to include it in the extension? Because so far I only find bugs.

PS: When I tried to fix everything, I got this version, but it also did not bring any benefit and was not completely stable:

Снимок экрана (1319)

In any case, although you were not against adding my alternative version, only one will get into the gallery, so I suggest discussing all our solutions and coming to one common extension.

PS: I have also started making extremely detailed documentation, but it has not yet been completed and is a bit outdated. I will return to it when there is a ready version of the extension. In the meantime, here is this sketch: https://true-fantom.github.io/docs/deltatime.html (It's really very broken, but I'll fix it!)

PPS: It is also possible to say in advance about the expansion design (block colors, text, block layout, etc). In my version, all this corresponds to the Scratch 3 style as much as possible, and therefore, despite the fact that I am quite open to changes, I would like to keep it. I picked up this color for half an hour ._.

XeroName:
@True-Fantom 
Hello Fantom.

The filter topology I've used is for smooth phase-changing of FPS and Dt, not for precision.
As you talked about, for precision, there should be no filtering at that values.
Although the standing reason of the filtering blocks are there may be a people who wanted.

For example, there may be some situation to prefer stability of value instead of precision.
Simply thought by graph or text display like FPS meter, which requires visibility first.
In these cases, fluctuating values even with superb precision, is likely less prefered than the stability of value because it should be easy to read, compare and recognized by human.

Yet except upon situation mentioned, the user who using extension still have the option to compelety turn off the filtering by setting Filter Strength by 1, which will focus on updating time and precision hang on with noise.

I do not big brain at mathmatics or calculating, etc and have not much knowledge about programming.
Most of user who decided to use Deltatime extension may be same with me.
They do not need ultra precision FPS or messy other options causing headace. (Otherwise would not use Scratch or TurboWarp!)
The functions who wanted by people like me may simple, easy to use and flexible at many applications.

Conclusion is that filter gives an option to choose precision or smoothing of deltatime values.

And I'm almost new at Github, have no idea why this is still in not approval(maybe there are something I've missed).
So maintaining this extension honestly makes me tired.
If you're ready to open your own deltatime extension, I'm ready to close this PR.
Do not mind me, even you don't have to mention me at any credits.

I don't know if I would being help you that much., but always feel free to ask something to me.

Thanks

True-Fantom:
@XeroName Thank you very much for your reply!

I understood why the filter is needed, and I support the fact that users should be able to use it! But I think the filter should not work by default. And it would also be cool to be able to create filters on any user variables, not just on fps from the extension. So, I think the best option would be to create a separate filter extension, and write about it in the DT documentation!

And well, yes, you can not worry at all that this PR has not merged for so long. Most likely this is due to the fact that Garbo is either very busy at a normal job, or just really likes to relax ^_^

So if you like this solution, let's do as you said, I will make a new PR with a new version of this extension and a filtering extension. The only point: I insist that you be mentioned as a co-author, because without you this extension would not exist at all :)

Thanks!

True-Fantom:
@XeroName Hm... I experimented a little and found out that the code of your filter is extremely easy to recreate, and it may even be useless to do a filter extension:
Снимок экрана (1320)
Also, all the disadvantages of manual re-creation can be eliminated using custom blocks:
Снимок экрана (1321)
Now I do not know what to do... Maybe just show this example in the DT documentation and that's it? What do you think?

XeroName:
@True-Fantom 
Removing the filtering blocks and _showing an example how to make it_ makes sense to me.

GarboMuffin:
Is the consensus to completely remove the filter for both delta time and FPS?

XeroName:
> Is the consensus to completely remove the filter for both delta time and FPS?

Yes it is.
I don't know that much detail but maybe @True-Fantom will recreate the PR about new deltatime.

GarboMuffin:
Hey, thank you very much for your patience. I assure you, the delay is all my fault, not yours

I have slightly massacred this extension. More blocks can be added later

XeroName:
> 

You don't have to sorry to me. I've been some time to rest or other things to do waiting for this.
I can understand your situation managing whole TurboWarp and others.

What I concerned about is something mistake I made because I don't know how to use GitHub :)

Appreciate that this PR has been merged!
Thank you!

True-Fantom:
@XeroName Hello! After so much time, I finally finished my PR!

As I see a lot of things have changed while I was away. I hope I did everything right with the new format.

PR #633, opened 2023-06-23
Add a new extension LazyAudio

0832k12:
This extension can play audio from website.


0832k12:
I just found an extension with the same functionality in the extension library.

PR #639, opened 2023-06-25
More Events: Add More Events extension

LilyMakesThings:
I've wanted to create this extension for a long while, this is about as many blocks as I can think of.

I needed this extension for specifically the "broadcast to target" block. (In hindsight, it may need some work).

image


LilyMakesThings:
This block now exists thanks to @CST1229 
image

LilyMakesThings:
Thumbnail:
image

BlueDome77:
"when extension added" will pair nicely with ShovelUtils "load extension" block!

BlueDome77:
Are you going to add this block?

scratchblocks

LilyMakesThings:
> Are you going to add this block?
> 
> scratchblocks
> 
> It only makes sense cause the original "broadcast ()" block has a STRING input

That's impossible to add.

BlueDome77:
> > Are you going to add this block?
> > scratchblocks
> > It only makes sense cause the original "broadcast ()" block has a STRING input
> 
> That's impossible to add.

why?

LilyMakesThings:
I forgot to say but I made the changes to this block I eluded to:
- It now supports sending messages to the stage target
- The menu isn't broken 
- It sends to clones as well
image

LilyMakesThings:
A few questions:
- Are merge conflicts something I'll need to resolve?
- Should these blocks be removed?
image

OeildeLynx31:
> A few questions:
> - Are merge conflicts something I'll need to resolve?
> - Should these blocks be removed?
> image
> 

For the second question, I think you can add a menu instead of the string input, but in which we can drag reporters. The menu could have all the basics events, like 'green flag', or 'stop' ...

SurvExe1Pc:
> > A few questions:
> > 
> > * Are merge conflicts something I'll need to resolve?
> > * Should these blocks be removed?
> >   image
> 
> For the second question, I think you can add a menu instead of the string input, but in which we can drag reporters. The menu could have all the basics events, like 'green flag', or 'stop' ...

i think its like that so you can start extension hats also

NexusKitten:
looking forward to this extension!

LilyMakesThings:
> I'm confused...
> 
> Screenshot 2023-07-11 at 8 59 18 AM Screenshot 2023-07-11 at 9 01 55 AM Source: [Penguin Mod](https://studio.penguinmod.site/editor.html)

I still don't quite know why I was credited, I think it was either because of the idea or because of the one block they share with this extension. Don't pay attention to penguinmod, this is the turbowarp repository

extensions/Lily/MoreEvents.js
```diff
@@ -0,0 +1,424 @@
...
+    { text: '3', value: '3' },
+    { text: '4', value: '4' },
+    { text: '5', value: '5' },
+    { text: '6', value: '6' },
+    { text: '7', value: '7' },
+    { text: '8', value: '8' },
+    { text: '9', value: '9' },
+  ];
+
+  setInterval(() => {
```
GarboMuffin, line 73: setInterval like this significantly impacts battery life. use vm.on('BEFORE_STEP', () => {}) instead (from memory, it's something like that, animated text uses it)

extensions/Lily/MoreEvents.js
```diff
@@ -0,0 +1,424 @@
...
+        runtime.shouldExecuteStopClicked = true;
+      });
+      const originalGreenFlag = vm.greenFlag;
+      vm.greenFlag = function() {
+        runtime.shouldExecuteStopClicked = false;
+        originalGreenFlag.call(this);
+      };
+
+      runtime.on('STAGE_SIZE_CHANGED', () => {
+        runtime.startHats('lmsMoreEvents_whenRuntimeOptionChanged', {
```
GarboMuffin, line 106: should these not be in runtime options

extensions/Lily/MoreEvents.js
GarboMuffin: its BEFORE_EXECUTE, the stop sign code below already uses it

extensions/Lily/MoreEvents.js
LilyMakesThings: I can remove them and put them into a separate PR later, if that's preferred. I'm inclined to agree

LilyMakesThings:
I removed "when runtime options changed" and "when extension added", but kept the turbo mode block.

LilyMakesThings:
Man I don't even know how to resolve conflicts myself 😭

LilyMakesThings:
Added data sending through broadcast messages
image

LilyMakesThings:
Using XML blocktype to get the broadcast menu into the blocks. Doing this until there's a way to do it without XML
image

GarboMuffin:
There is a VM change that is merged but not deployed yet that will make whenValueChanged work properly

LilyMakesThings:
should "forever" only run while the project is active?

LilyMakesThings:
All of these have now been implemented. I'm not adding anymore than this.

image

LilyMakesThings:
...except that.

GarboMuffin:
is this done?

LilyMakesThings:
yep

LilyMakesThings:
OH DAMN

PR #651, opened 2023-06-28
Longman Dictionary extension

veggiecan0419:
Ever needed to get the definitions of words in your projects? 
If so, here it is.


GarboMuffin:
Icon works poorly in dark mode

image

You will need to handle fetch() errors and return a readable error message instead of letting it return undefined

veggiecan0419:
**Update details**
- icon now works in dark and light mode
- fixed the _fetch()_ errors
- renamed 'definition of...' to 'primary definition of...'
- added 'all definitions of [word] in the Longman Dictionary '
- added credits to veggiecan0419

PR #656, opened 2023-06-30
LilyMakesThings: Update Extension Thumbnails

LilyMakesThings:
I've gone back through some of my extensions to make their thumbnails more consistent.
image
image
image
image


BlueDome77:
I think Clones and Looks Plus looked much better before and matched the default Turbowarp Extension Style made by @True-Fantom as opposed to the new ones.

LilyMakesThings:
> I think Clones and Looks Plus looked much better before and matched the default Turbowarp Extension Style made by @True-Fantom as opposed to the new ones.

The "TurboWarp extension style" was actually supposed to match Scratch's extension thumbnails. These mimic the Scratch extension thumbnail style.

PR #664, opened 2023-07-01
2 New Thumbnails (Notifications and Numerical Encoding)

NexusKitten:
If the original owners want their thumbnails back, feel free to revert this pull request, however I felt as if these two extensions' thumbnails stuck out according to the rest.

Uncropped versions:
notifications
numericalencoding

Cropped:
notifications (1)
numericalencoding (1)


Pizzalover4783:
As the creator of the original Numerical Encoding thumbnail, I'm so glad that eyesore of an image is finally getting replaced

BlueDome77:
Please submit your extension thumbnails here: #90

BlueDome77:
Numerical Encoding is fire!

PR #632, opened 2023-06-23
Add clipboard blocs

tomyo-code:
A litlle extension for interact with the clipboard

PR #665, opened 2023-07-01
Add Clipboard Extension

AdamMady:
### Description: A few blocks that interact with the user's clipboard. Created by [AdamMady2](https://scratch.mit.edu/users/AdamMady2/)


LilyMakesThings:
Another one?

AdamMady:
> Another one?

one already exists?

AdamMady:
dang I should've done my research but I'll leave this here since it has more features than the other one

0832k12:
Yes, you are right.
**But** I think your newly added block can be achieved through variables, an extension should only do its job.

LilyMakesThings:
> Yes, you are right. **But** I think your newly added block can be achieved through variables, an extension should only do its job.

And the other extension has blocks that are not in this extension but can't be replicated easily.

NexusKitten:
If you don't mind, I'll probably submit another pull request that merges this and #632 into one, since we don't exactly need 2 clipboard extensions. With that being said, both of you will still receive credit for your work.

GarboMuffin:
closing in favor of https://github.com/TurboWarp/extensions/pull/667

PR #672, opened 2023-07-02
Sound Expanded: Add Sound Expanded extension

LilyMakesThings:
Sound Expanded is an extension that adds a few missing blocks from the Sound category.

This serves as a precursor to my next extension, Audio Engine.
image


NexusKitten:
sound-category booleans are cursed

AdamMady:
this reminds me of when I tried to make a "while playing note {VAR} for {VAR} beats in the {VAR}th beat fade into note {VAR} for {VAR} beats" block (it went badly)

LilyMakesThings:
> isnt start sound on loop the same as start looping sound?

"start looping ()" only works for sounds that are already playing.

AdamMady:
I have a good idea @LilyMakesThings how about you make a block where you could change the global volume of the project instead of changing each volume of each sprite one by one

AdamMady:
I think I'll have a go for both I have nothing better to do

BlueDome77:
Could you add these blocks?

scratchblocks-2

LilyMakesThings:
Please look at the block palette before suggesting blocks

AdamMady:
I know this is a bit much for a simple extension but is it possible to do a record block? where it records using the microphone and saves it to be used

BlueDome77:
> I know this is a bit much for a simple extension but is it possible to do a record block? where it records using the microphone and saves it to be used

There is an extension exactly for that! Find "Recording" in the list: https://github.com/BlueDome77/Turbowarp-Extension-List/blob/main/README.md

BlueDome77:
I think you should merge the Sound extension with this

BlueDome77:
Could you add these?

scratchblocks-2

LilyMakesThings:
> Could you add these?
> 
> scratchblocks-2

No

Martinelplayz:
> Could you add these?
> 
> scratchblocks-2

:c

NexusKitten:
can't wait!

extensions/Lily/SoundExpanded.js
```diff
@@ -0,0 +1,355 @@
...
+      return -1;
+    }
+
+    _wrapClamp(n, min, max) {
+      const range = (max - min) + 1;
+      return n - (Math.floor((n - min) / range) * range);
+    }
+
+    _getTargets() {
+      const spriteNames = [];
```
GarboMuffin, line 323: let's avoid the same mistake as looks+. have the top value always be "myself" with value `_myself_` and just list every sprite under it, and also add the stage (`_stage_`)

LilyMakesThings: I made a script that does that for skins, I'll add that now

GarboMuffin:
the combination of "start sound [ ] on loop" and "start looping [ ]" is going to confuse many people

LilyMakesThings:
> the combination of "start sound [ ] on loop" and "start looping [ ]" is going to confuse many people

If I add "while it's playing" or something to the end, would that work? or could I literally just remove it

CST1229:
> If I add "while it's playing" or something to the end, would that work? or could I literally just remove it

maybe `begin/end looping for []`

LilyMakesThings:
> > If I add "while it's playing" or something to the end, would that work? or could I literally just remove it
> 
> maybe `begin`/`end looping for []`

That's the one I think

LilyMakesThings:
Update: added project volume blocks. Works (mostly) with addons' volume slider.
image

LilyMakesThings:
I added something an attribute for dataURIs so you can export the sound as a data uri.

I pretty much only did this for the Sound extension, as well as the Video extension.

extensions/Lily/SoundExpanded.js
```diff
@@ -0,0 +1,365 @@
...
+                type: Scratch.ArgumentType.SOUND,
+              },
+              START: {
+                type: Scratch.ArgumentType.NUMBER,
+                defaultValue: 0,
+              },
+            },
+          },
+          {
+            opcode: "stopLooping",
```
GarboMuffin, line 45: is it intended that this doesn't actually stop the sound immediately?

extensions/Lily/SoundExpanded.js
```diff
@@ -0,0 +1,365 @@
...
+    setProjectVolume(args) {
+      const value = Scratch.Cast.toNumber(args.VALUE);
+      const newVolume = this._wrapClamp(value / 100, 0, 1);
+      runtime.audioEngine.inputNode.gain.value = newVolume;
+    }
+
+    changeProjectVolume(args) {
+      const value = Scratch.Cast.toNumber(args.VALUE) / 100;
+      const volume = runtime.audioEngine.inputNode.gain.value;
+      const newVolume = this._wrapClamp(volume + value, 0, 1);
```
GarboMuffin, line 309: when volume is 0 and someone changes it by -1, it probably shouldn't wrap around to 199%

GarboMuffin:
Also how does this compare to https://github.com/TurboWarp/extensions/pull/756?

extensions/Lily/SoundExpanded.js
LilyMakesThings: yes, that's why there's a "stop sound" block.

LilyMakesThings:
> Also how does this compare to #756?

ultrasound appears to not work in line with the default sound category, rather on its own use of the web audio api - think audio engine, audiostream or tuneshark. the main purpose of this is no different to clones plus, looks plus, etc.

extensions/Lily/SoundExpanded.js
LilyMakesThings: Good point, will change that

GarboMuffin:
done?

LilyMakesThings:
yes

PR #676, opened 2023-07-02
Move blocks from 0tils to corrosponding extensions

SurvExe1Pc:
Moves my networking blocks to network.js,
Moves my encryption to encodingv2.js as it break compatibility, (adds more hashes+encoding options and also AES)
Moves my raw download block to files.js
Moves delete costume block and delete sprite+no confirmation to ShovelUtils


SurvExe1Pc:
I need to find a home for 
image
now

PR #683, opened 2023-07-03
added websocket extension

RedMan13:


RedMan13:
ima mess around with github to try and get rid of the http ext's commits

PR #685, opened 2023-07-03
Add upload extension

clay-rip:
Pretty simple extension that can't be replicated with pre-existing extensions


BlueDome77:
What would this be useful for?

ClaytonTDM:
I used it in my cloud chat project to have an upload file button that posts a link to the file in chat

GarboMuffin:
Is this extension already possible through a combination of files and the HTTP/fetch/Network extensions?

ClaytonTDM:
Nope, because of how multipart/form-data uploads are handled, it isn't yet possible.

extensions/Clay/upload.js
```diff
@@ -0,0 +1,69 @@
+(function (Scratch) {
+class Upload {
+    getInfo() {
+      return {
+        id: "uploadfile",
```
GarboMuffin, line 5: extension id must be somehow based on your username, eg. clayuploadfile

GarboMuffin:
Would it be possible to turn this into a more general file uploading extension?

At least a block like "set file host to [ ]",  even if it only has one option for now

ClaytonTDM:
> Would it be possible to turn this into a more general file uploading extension?
> 
> At least a block like "set file host to [ ]", even if it only has one option for now

It depends on the case. For a service like Gofile, they have multiple servers (store1, store2, etc) so there would have to be custom logic to grab the correct server from their API.
Although, it could work better with both the current upload file to URL, and "upload file to [dropdown]" so there aren't any unnecessary blocks. That's one thing I hate about extensions such as numerical encoding, why have a block that just sets a custom variable when you could have a single reporter?

Anyways, I'll get started on the new block now.

ClaytonTDM:
..?

SaFireGalaxy:
Hey. something that's annoying with gofile and file.io is they dont provide direct links to the raw image. This really adds tons of limitations so I can't have the image show in my turbowarp project. Is there another hosting exactly like gofile that actually gives you the link to the raw hosted image, so i can have a plugin display the image in the project?

ClaytonTDM:
As far as I know, no. If I find one, I'll be sure to add support for it though.

SaFireGalaxy:
Thanks @ClaytonTDM!

Prizmin:
Is this getting merged or what?

LilyMakesThings:
> Is this getting merged or what?

Please have patience, we're all volunteers here. As stated in CONTRIBUTING.md, "we'll get around to you eventually".

GarboMuffin:
hi, sorry for the delay, currently whittling away at a review queue that pilled up during months of long days at work

here's the deal for this extension -

 - The HTTP extension does support multipart now, so this is not as necessary (but still happy to merge it, if you resolve the issues below)
 - It would be great if the extension was not limited to just files on your computer already, but if it also allowed any text, such as data: URI snapshots of the stage from looks+
 - Your scripts need to handle what happens when someone cancels the file dialog. At a bare minimum you can listen for the `cancel` event, which is a new thing for dealing with this, but support is still very new. This is why the file extension insists on showing a modal.
 - Errors in getGofileServer (for example, if someone denies the fetch) need to be handled

ClaytonTDM:
Sure! I've added all the changes and fixes you mentioned.

ClaytonTDM:
its been like a year

PR #689, opened 2023-07-04
Fix some image pixelation issues

GarboMuffin:


horizontal-shading:
I don't notice any difference.

BlueDome77:
Finally!

GarboMuffin:
It turns out Safari simply does not provide any way to make images inside an SVG rendering pixelated at this time. I think I will convert those to PNGs

Related keywords: image-rendering: pixelated, optimizeSpeed, crisp-edges not working in Safari SVG image

PR #690, opened 2023-07-04
Block colors for var and list extension

Faris90:
Block colors credit to @TheShovel

PR #692, opened 2023-07-04
Update var-and-list.js

qxsck:
#690 Adjustments have been made based on this.

### DIFF
--- a/extensions/qxsck/var-and-list.js
+++ b/extensions/qxsck/var-and-list.js
@@ -378,11 +378,9 @@
       const value = args.VALUE;
       if (variable) {
         for (var i = 0;i < variable.value.length;i++) {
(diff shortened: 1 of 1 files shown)


GarboMuffin:
we are not going to be coloring blocks like this at this time

qxsck:
> 我们目前不会像这样着色块

sorry

Faris90:
> we are not going to be coloring blocks like this at this time

@GarboMuffin and why not?

LilyMakesThings:
> > we are not going to be coloring blocks like this at this time
> 
> @GarboMuffin and why not?

I'm not GarboMuffin but I assume it's because at the moment, it's really hacky and untested. 

There's also no benefit for using this. Just colour your blocks like normal.

PR #700, opened 2023-07-06
files: Add download URL block

CST1229:
image
Useful for downloading binary data that downloading as text would mess up, like PNG or zip (*hint hint*) files.

### DIFF
--- a/extensions/files.js
+++ b/extensions/files.js
@@ -165,19 +165,7 @@
   });
 
   /**
-   * @param {string} data Text to download
-   * @param {string} file Name of the file
-   */
-  const downloadText = (data, file) => {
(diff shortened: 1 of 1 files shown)


LilyMakesThings:
While you're there, maybe also put a spacer between the modal selector and download blocks?

GarboMuffin:
you know, I suppose we may as well make this just support all URLs?

LilyMakesThings:
> you know, I suppose we may as well make this just support all URLs?

This would be more than ideal for some developers if they want to have an efficient way to download files from a repo. I can already think of a project I'd make with this.

Any concerns anyone would have about malicious files is made redundant by the fact fetch exists.

PR #727, opened 2023-07-12
Add new blocks to tweening extension

asherlr23:
This adds 3 new blocks that all allow you to run an eased animation in one block rather that having to make a script yourself. There are 3 new blocks:

(linear v) ease (in v) from x: (-100) y: (-100) to x: (100) y:(100) at speed (1) :: extension

(linear v) ease (in v) to x: (100) y:(100) at speed (1) :: extension

(linear v) ease (in v) from size: (0)% to size: (100)% at speed (1) :: extension

Yes, that is in scratchblocks.
https://scratchblocks.github.io/#?style=scratch3&script=(linear%20v)%20ease%20(in%20v)%20from%20x%3A%20(-100)%20y%3A%20(-100)%20to%20x%3A%20(100)%20y%3A(100)%20at%20speed%20(1)%20%3A%3A%20extension%0A%0A(linear%20v)%20ease%20(in%20v)%20to%20x%3A%20(100)%20y%3A(100)%20at%20speed%20(1)%20%3A%3A%20extension%0A%0A(linear%20v)%20ease%20(in%20v)%20from%20size%3A%20(0)%25%20to%20size%3A%20(100)%25%20at%20speed%20(1)%20%3A%3A%20extension

image


LilyMakesThings:
Woohoo!

extensions/JeremyGamer13/tween.js
```diff
@@ -81,224 +128,295 @@
...
+            const endy = Cast.toNumber(args.ENDY);
+            const speed = Cast.toNumber(args.SPEED);
+
+            if (typeof util.stackFrame.loopCounter === 'undefined') {
+                util.stackFrame.loopCounter = 100;
+            }
+            const progress = Cast.toNumber(Math.round(this.clamp(util.stackFrame.loopCounter, 0, 100))) / 100;
+            const tweened = this[easeMethod](progress, easeDirection);
+            const x = Cast.toNumber(this.multiplierToNormalNumber(tweened, endx, startx));
+            const y = Cast.toNumber(this.multiplierToNormalNumber(tweened, endx, startx));
```
DNin01, line 147: References to `endx` and `startx` on the right-hand side of `const y = ...` should be updated to `endy` and `starty`. Otherwise, the y position will be the same as the x position.

extensions/JeremyGamer13/tween.js
```diff
@@ -81,224 +128,295 @@
             const tweened = this[easeMethod](progress, easeDirection);
             return this.multiplierToNormalNumber(tweened, start, end);
         }
+        tweenPlayAnimationPosition(args, util) {
```
DNin01, line 131: For some reason, the sprite doesn't glide from point A to point B properly when using this block. Try it for yourself - setting mode to "linear", it should glide to the ending point at a fixed rate, but it instead eases in and eases out, and seems to stop before the block finishes.

This is probably because some starting position constant is constantly getting updated with the sprite's current x position whenever it moves forward a step, but it should stay the same throughout the whole animation. Haven't studied where that happens, though.

This does not happen with any of the other blocks.

.github/workflows/deploy.yml
DNin01, line 1: Please don't remove workflow files from this pull request, or they will get removed from the upstream repository when the PR is merged. Instead, you can disable Actions on your fork from Settings → Actions → General.

DNin01 (changes requested):

extensions/JeremyGamer13/tween.js
asherlr23: > References to `endx` and `startx` on the right-hand side of `const y = ...` should be changed to `endy` and `starty`. Otherwise, the y position will be the same as the x position. There are 2 occurrences of this issue.

Ah. Not sure how I didn’t catch that, I tested all of my blocks and they work fine.
As for the workflows, I’ll try to reverse that. I originally tried to add a action myself, and later realized that I didn’t want or need to use GitHub.io, just a local server, not knowing there was already workflows inside.

GarboMuffin:
Ignore the current "All checks have failed" -- that's unrelated

Would it make sense to replace the idea of "speed" with just the total time of the tween instead?

asherlr23:
> Ignore the current "All checks have failed" -- that's unrelated
> 
> Would it make sense to replace the idea of "speed" with just the total time of the tween instead?

Probably. Ill try that.

GarboMuffin:
It looks like someone else has opened something similar https://github.com/TurboWarp/extensions/pull/766

JeremyGamer13:
could these use seconds instead of the vague "speed" number?
it also feels like the first block should just act like the second block but with the first numbers being the current sprite position

asherlr23:
ok i've been neglecting this entire pr for a while now so im gonna draft it for now until i feel like working on it

GarboMuffin:
Superceded by https://github.com/TurboWarp/extensions/pull/851

PR #723, opened 2023-07-10
Add sorting blocks to JSON.

SurvExe1Pc:
Blocks added:
image

image
allows you to sort some numbers in ascending or descending order using a algorithm in the dropdown[f1]

image
allows you to sort letters / words in ascending or descending order.

Sorting option for numbers [f1]:
image

image

Difference from list utils version:
(video from @LilyMakesThings)
https://github.com/TurboWarp/extensions/assets/135030944/bf13697f-d459-4b25-8d34-2b4b3822a7dd

PR #728, opened 2023-07-12
Add Sorting Blocks to JSON Extension

WeirdBoi12408:
# New Blocks

image

## Convert all values in ( ) to ( )
This block will convert all the values in an object or items in an array to a specified type.

If the input is an array, it will simply convert all the items to the specified type.
image

If the input is an object, it will convert all the values to the specified type.
image

## Sort JSON ( ) in ( ) order
This block sorts a JSON in numerical or alphanumerical order depending on its items / values.

If the input is an array, it will be sorted. If the input is an object, it will be sorted and only the keys will be returned.
image

If the JSON input only contains numbers, it will be sorted numerically.
image

IF the JSON input contains anything other than a number, it will be sorted alphanumerically.
image

## Grade JSON ( ) in ( ) order
This block will return the indices of the array / object as if it were sorted.

If the input is an array, it will grade it normally.
image

If the input is an object, it will discard the keys and grade based on the values.
image

*note: This block is equivalent to the monadic* `⍋` *and* `⍒` *functions in APL.*

# Useful block combinations

If you want to sort a list of stringified numbers and want to sort them numerically, you can convert them to a Number type before sorting.
image

If you want to do the opposite of a grade (for each index return where it moves) then simply put a grade inside of a grade.
image

If you want to sort a list numerically then mix some list blocks with the sorting blocks.
image

### DIFF
--- a/extensions/Skyhigh173/json.js
+++ b/extensions/Skyhigh173/json.js
@@ -12,6 +12,42 @@
     blockType: 'label',
     text: text
   });
+  
+  // sorting
+  
+  function sortArray(array, order) {
+    if (array.every(item => typeof item === 'number')) {
+      if (order === 'ascending') {
+        array.sort((a, b) => a - b);
+      } else if (order === 'descending') {
+        array.sort((a, b) => b - a);
+      }
+    } else {
+      if (order === 'ascending') {
+        array.sort();
+      } else if (order === 'descending') {
+        array.sort().reverse();
+      }
+    }
+    
+    return array;
+  }
+  
+  function gradeArray(array, order) {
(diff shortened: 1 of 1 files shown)


Skyhigh173:
great! however there's two JSON extension poll about sorting.

WeirdBoi12408:
yeah ik i just decided to create a pull request here because lily thought that i should also submit it

WeirdBoi12408:
oops i messed up one of the pictures please ignore the nested grades one

GarboMuffin:
The sorting part of this was made redundant by #723 

We can add the other blocks if you rebase this on master

PR #732, opened 2023-07-12
Add More Motion

NexusKitten:
More Motion is an extension to add several utility blocks pertaining to the "Motion" category, such as blocks for distance, stepping, and checking position. Although it's blocks are somewhat limited as of currently, I wanted to set the groundwork for a motion extension since we don't currently have one yet.

image (30)


idoinfactexist:
call it motion plus, its just too easy

horizontal-shading:
YESSSS MOTION+ I WAS WAITING FOR THIS

Martinelplayz:
(distance between (x1) (y1) and (x2) (y2))

NexusKitten:
> (distance between (x1) (y1) and (x2) (y2))

This block is already in Graphics 2D.

CubesterYT:
@NexusKitten are you ok lol?

GarboMuffin:
please keep the pull request names reasonable if you want them to be looked at

NexusKitten:
> please keep the pull request names reasonable if you want them to be looked at.

Sorry about that!

extensions/NexusKitten/moremotion.js
```diff
@@ -0,0 +1,294 @@
...
+      } else {
+        util.target.setDirection(((180 / Math.PI) * Math.atan((x - util.target.x) / (y - util.target.y))));
+      }
+    }
+
+    rotationStyle(args, util) {
+    return util.target.rotationStyle;
+    }
+
+    fence(util) {
```
GarboMuffin, line 220: - first argument is always args (an empty object in this case). second argument is always util.
 - you should make the fencing match Scratch. here's a hint: https://github.com/TurboWarp/scratch-vm/blob/9329427fb57de83cf5d8c2774b55013c60f39ed4/src/sprites/rendered-target.js#L269

extensions/NexusKitten/moremotion.js
```diff
@@ -0,0 +1,294 @@
...
+      // Shoutout to Pythagoras!
+      return Math.sqrt(((x - util.target.x) ** 2) + ((y - util.target.y) ** 2));
+    }
+
+    steptowards(args, util) {
+      const x = Scratch.Cast.toNumber(args.X);
+      const y = Scratch.Cast.toNumber(args.Y);
+      const steps = Scratch.Cast.toNumber(args.STEPS);
+      const val = steps / (Math.sqrt(((x - util.target.x) ** 2) + ((y - util.target.y) ** 2)));
+      util.target.setXY(((x - util.target.x) * (val)) + util.target.x, ((y - util.target.y) * (val)) + util.target.y);
```
GarboMuffin, line 246: If the distance to the point is less than the distance to step, wouldn't it make sense to not step too far?

extensions/NexusKitten/moremotion.js
```diff
@@ -0,0 +1,294 @@
...
+                type: Scratch.ArgumentType.NUMBER,
+                defaultValue: '0'
+              }
+            }
+          },
+          {
+            filter: [Scratch.TargetType.SPRITE],
+            opcode: 'inbetween',
+            blockType: Scratch.BlockType.BOOLEAN,
+            text: 'in-between x1: [X1] y1: [Y1] x2: [X2] y2: [Y2]?',
```
GarboMuffin, line 169: a name like "in rectangle" or something might be less ambiguous

extensions/NexusKitten/moremotion.js
```diff
@@ -0,0 +1,294 @@
...
+              Y2: {
+                type: Scratch.ArgumentType.NUMBER,
+                defaultValue: '100'
+              }
+            }
+          },
+        ],
+        menus: {
+          WHAT: {
+            acceptreporters: false,
```
GarboMuffin, line 192: why not true?

extensions/NexusKitten/moremotion.js
NexusKitten: All issues you presented should be resolved in the new commit.

BlueDome77:
It should just be called **Motion Plus** to match the other category expansion extensions

GarboMuffin:
I made some changes

 - distinction between costume size and actual sprite size after rotation, etc.
 - touching point works with rotation etc. and checks if the costume actually exists at that point
 - rectangle touching block now works with rotation etc. and checks if any part of the costume actually intersects the rectangle (it also fixes the argument order if someone gets it backwards)

the last 2 are significant breaking changes, so let me know if you need the old behavior back

LilyMakesThings:
"touching" as a concept is a sensing thing. Wouldn't it make more sense within sensing+?

NexusKitten:
> "touching" as a concept is a sensing thing. Wouldn't it make more sense within sensing+?
> 
> "Motion" describes movement, and the reporters in the motion category typically get an axis of movement, not a condition. Otherwise, distance would've been in motion.

this is true, I mostly added it into a motion because:
1. both touching blocks are focused on the sprites position, and I figured from a logical standpoint people would associate position booleans with motion instead of sensing
2. sensing+ is kind of a mess of blocks right now, I'd rather not throw it further into "utilities" category until we can find a solution to its current problems

PR #736, opened 2023-07-13
Update couplers.js

BlueDome77:
Added COLOR, ANGLE, MATRIX, and NOTE reporters from LMS Utils


LilyMakesThings:
you can't add a c-shape because it doesn't compile

BlueDome77:
> you can't add a c-shape because it doesn't compile

Someday...

PR #747, opened 2023-07-16
A bunch of new things for Canvas Effects

TheShovel:
Added:
Sepia
Transparency
Scale
SkewX
SkewY
Offset X
Offset Y
Rotation
Border radius


BlueDome77:
Lets go!!

GarboMuffin:
Have you tested this in the packager?

TheShovel:
> Have you tested this in the packager?

Yes I did. There are no issues.

gomiigg:
It would be cool if there were a `fisheye` effect that could be combined with the `border radius` effect to simulate a CRT effect

TheShovel:
> It would be cool if there were a `fisheye` effect that could be combined with the `border radius` effect to simulate a CRT effect

No.

PR #756, opened 2023-07-17
Create ultrasound.js extension

DaTripleBlock:
Adds grouping and allows load from urls

### DIFF
--- a/extensions/DaTripleBlock/ultrasound.js
+++ b/extensions/DaTripleBlock/ultrasound.js
@@ -1,0 +1,268 @@
+/**
+ * Created By Da TripleBlock (TTT)
+ * ♥Please Dont Remove this Comment ♥
+ */
+(function (Scratch) {
+    'use strict';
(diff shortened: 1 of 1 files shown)


SharkPool-SP:
Not much use for this extension when the same can be accomplished with the upcoming Sound+ extension. There’s also not many blocks for messing with audio…

GarboMuffin:
How does this compare to https://github.com/TurboWarp/extensions/pull/672?

SharkPool-SP:
> How does this compare to #672?

Outdated comment, I meant Audio Engine from lily

LilyMakesThings:
> > How does this compare to #672?
> 
> Outdated comment, I meant Audio Engine from lily

Yeah I have 2 sound extensions, don't get em mixed up lol

Brackets-Coder:
> > The comment thread in #2271 demonstrates the opposing idea that we shouldn't impulsively close duplicate PRs just because they're duplicates. We evaluate each PR for their strengths and weaknesses and prioritize them based on which one is "better" (though "better" is not objective and each one may have different benefits), which one came first, etc. etc. The goal is to find which PR will most benefit the Turbowarp community and if both are advantageous encourage collaboration.
> 
> Well, let's compare the two.
> 
> ### Sound Expanded
> * Merged to the extensions repository
> * Predates the introduction of ultrasound.js
> * Most comprehensive option at the moment
> * Doesn't have URL loading, so not redundant with URL Playback
> 
> ### ultrasound.js
> * Has not been merged yet
> * Contains URL loading blocks that basically duplicate the functionality of URL Playback
> * Uses a third-party CORS proxy that we cannot guarantee will continue to exist in the future
> * Does not use the `Scratch.translate` API
> * Mostly redundant with Sound Expanded
> 
> Sure, you can try to add this if you'd like, but it will take a lot of work to modernize the codebase for little gain.
> 
> I think that a better approach would be porting the audio grouping blocks to Sound Expanded.

Interesting analysis, I'll take a look for myself to deduce and validate that information 👍

PR #759, opened 2023-07-17
Add user preferences blocks to Navigator

DNin01:
Adds blocks to Navigator for getting a few user preferences.

Originally, this was a new extension, but it fit better in the existing Navigator extension.

Partially tested sandboxed and unsandboxed.


BlueDome77:
I think this extension should be merged with navigator

Dy-tech:
> I think this extension should be merged with navigator

Me too.

NexusKitten:
As other commenters have pointed out, these functions are miniscule enough to likely not require it's own extension. As well, specifically the dark mode block from your extension is also in Pull Request #614.
The best course of action would probably to be to merge this extension with another. As @BlueDome77 and @Dy-tech mentioned, several of the functions would go well in Navigator, however I'm also willing to add the missing blocks to #614 if you'd like.

PR #764, opened 2023-07-19
Skins: Add Skins Extension

LilyMakesThings:
Skins aims to be an expanded version of the SVG block in Looks Plus, but with support for bitmap images as well as separate clone skins.

image

I urge you to give it a try, and suggest ideas for blocks!

(Please note, I will not be adding the replace color block, or similar blocks to this extension. That will be left for a separate extension I am working on)


LilyMakesThings:
I am not fixing those errors

LilyMakesThings:
What is the cause of that error? I've tried fixing it before but to no avail

CST1229:
> What is the cause of that error?

It's because you're using `Image()`, and setting the src of an Image sends a network request. I think the linter isn't (and probably won't be) smart enough to detect that you've used `Scratch.fetch()` with that same URL before, so you'll have to do what the message says:
> then add `// eslint-disable-next-line no-restricted-syntax`

LilyMakesThings:
> > What is the cause of that error?
> 
> It's because you're using `Image()`, and setting the src of an Image sends a network request without the user's permission. I think the linter isn't (and probably won't be) smart enough to detect that you've used `Scratch.fetch()` with that same URL before and there is no `Scratch.Image()`, so you'll have to do what the message says:
> 
> > then add `// eslint-disable-next-line no-restricted-syntax`

Thank you so much

LilyMakesThings:
Some problems have been noted regarding the rotation center of skins, which will require a new block and a patch

LilyMakesThings:
We do not need to add any new blocks for now.

PR #787, opened 2023-07-27
Add LZ-String.js

TheShovel:
It's LZ-String, but implemented in TurboWarp. Nothing else to say. It just works.

Screenshot_20230727_183605

### DIFF
--- a/extensions/TheShovel/LZ-String.js
+++ b/extensions/TheShovel/LZ-String.js
@@ -1,0 +1,273 @@
+(function(Scratch) {
+    'use strict';
+
+    //Code from https://github.com/pieroxy/lz-string/
+    var LZString = function() {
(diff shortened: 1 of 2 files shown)


extensions/TheShovel/LZ-String.js
```diff
@@ -0,0 +1,275 @@
+(function(Scratch) {
+    'use strict';
+
+    //Code from https://github.com/pieroxy/lz-string/
```
GarboMuffin, line 4: copy and paste the `LICENSE` file contents into a comment by after this - https://github.com/pieroxy/lz-string/blob/master/LICENSE

GarboMuffin:
If your project was broken by me changing the extension ID, send it to me and I will fix it for you

PR #799, opened 2023-07-29
CST1229/zip: use exported JSZip from VM

GarboMuffin:


GarboMuffin:
(this will wait until after packager update for placeholder & desktop app)

PR #761, opened 2023-07-18
S-Grab V2

SharkPool-SP:
Added 7 NEW BLOCKS to S-Grab that fetch useful stats about users and projects! 

Changes:
-Added Credits
-Added new blocks
-Fixed minor bugs
-Changed some og code to include commas and such (just quality of life stuff, doesn't break anything)

Screenshot 2023-07-17 223103

PR #803, opened 2023-07-29
Add Trampoline Extension (S-Grab Replacement)

NexusKitten:
Partially resolves #761 
Adds Trampoline, an extension for retrieving information from Scratch users and projects, using a combination of the TurboWarp Trampoline API and Scratch API, and removes/fixes S-Grab (#371) due to copyright, as well as several major bugs.
image


NexusKitten:
What would be a good workaround for:
`error  Do not access Object.prototype method 'hasOwnProperty' from target object  no-prototype-builtins`

callumjt:
why though?

callumjt:
> why though?

cant you just update s-grab?

NexusKitten:
> > why though?
> 
> cant you just update s-grab?

The two extensions use different APIs; S-Grab runs on Scratch DB, which is the same thing ScratchStats uses. Trampoline uses the Turbowarp Trampoline API, and the base Scratch API. Although the two extensions have some overlapping blocks, they're internally very different. However, S-Grab still has some blocks exclusive to Scratch DB, so I'm keeping it as a hidden extension to be thorough.

callumjt:
> > > why though?
> 
> > 
> 
> > cant you just update s-grab?
> 
> 
> 
> The two extensions use different APIs; S-Grab runs on Scratch DB, which is the same thing ScratchStats uses. Trampoline uses the Turbowarp Trampoline API, and the base Scratch API. Although the two extensions have some overlapping blocks, they're internally very different. However, S-Grab still has some blocks exclusive to Scratch DB, so I'm keeping it as a hidden extension to be thorough. 

I see

LilyMakesThings:
It's also standard for us to move our extension rewrites to a new file, especially when having them in the same file would mean hiding all of the old blocks.

NexusKitten:
@GarboMuffin I probably should've mentioned this sooner, however this pull request also removes the Scratch Logo from the MenuURI (which we forgot to do when removing Scratch copyright). It's not exactly urgent, however it's something to consider.

NexusKitten:
I'll draft this until we get more information on the [API updates](https://staging.turbowarp.org/august-2023-scratch-maintenance.html).

NexusKitten:
The API updates weren't as significant as anticipated-- updated to fit new syntax and reopened.

Issue #806, opened 2023-07-30
LZstring bug


Skyhigh173:
It doesn't work when input is number.![image](https://github.com/TurboWarp/extensions/assets/100467674/72b1a369-d4c0-4c71-8aac-c1b06bf9beff)

PR #808, opened 2023-07-30
TheShovel/LZ-String: fix a couple errors

GarboMuffin:
- fix "compress [11] as [...]" returning broken output - https://github.com/TurboWarp/extensions/issues/806
 - put more error handling in decompress


SurvExe1Pc:
ah ok!

PR #813, opened 2023-07-30
Prettier

GarboMuffin:
instead of eslint screaming at you and me fixing all of your style issues somewhat manually, we'll just make the computer do it for you


LilyMakesThings:
I think I have an idea of what this will do, but just to clarify, what _does_ this do?

Flappy25:
Wow

CubesterYT:
The use of prettier is going to be godly, were finally going to have a way to have a unified style.

DNin01:
The code blocks in the docs may also need to be formatted, depending on the configuration.

David-Orangemoon:
I already use prettier

PR #818, opened 2023-07-31
Better Input Extension

SharkPool-SP:
**VERSION 4.0!!** 
Adds various blocks for custom CSS inputs. Focused on remodeling the ask () block to be better.
image

Screenshot 2023-10-09 at 9 04 42 PM


SharkPool-SP:
Here is the thumbnail that I forgot to include: 
Better Input

qxsck:
> Adds various blocks for custom CSS inputs. Focused on remodeling the ask () block to be better. Screen Shot 2023-07-30 at 11 49 39 PM

Should UserInput return null instead of 'canceled' without input?

SharkPool-SP:
> > Adds various blocks for custom CSS inputs. Focused on remodeling the ask () block to be better. Screen Shot 2023-07-30 at 11 49 39 PM
> 
> Should UserInput return null instead of 'canceled' without input?

Doesn’t really matter. If there’s no response it will return empty; just like the current ask and wait block

LilyMakesThings:
> > > Adds various blocks for custom CSS inputs. Focused on remodeling the ask () block to be better. Screen Shot 2023-07-30 at 11 49 39 PM
> > 
> > 
> > Should UserInput return null instead of 'canceled' without input?
> 
> Doesn’t really matter. If there’s no response it will return empty; just like the current ask and wait block

It does matter, actually. For parity with the actual ask block, you should return an empty string.

SharkPool-SP:
> > > > Adds various blocks for custom CSS inputs. Focused on remodeling the ask () block to be better. Screen Shot 2023-07-30 at 11 49 39 PM
> > > 
> > > 
> > > Should UserInput return null instead of 'canceled' without input?
> > 
> > 
> > Doesn’t really matter. If there’s no response it will return empty; just like the current ask and wait block
> 
> It does matter, actually. For parity with the actual ask block, you should return an empty string.

No that’s what I mean, it will return an empty string when there’s no input

LilyMakesThings:
> > > > > Adds various blocks for custom CSS inputs. Focused on remodeling the ask () block to be better. Screen Shot 2023-07-30 at 11 49 39 PM
> > > > 
> > > > 
> > > > Should UserInput return null instead of 'canceled' without input?
> > > 
> > > 
> > > Doesn’t really matter. If there’s no response it will return empty; just like the current ask and wait block
> > 
> > 
> > It does matter, actually. For parity with the actual ask block, you should return an empty string.
> 
> No that’s what I mean, it will return an empty string when there’s no input

It should also do that when it's cancelled. Make a boolean that is toggled when an input is cancelled or not.

SharkPool-SP:
> > > > > > Adds various blocks for custom CSS inputs. Focused on remodeling the ask () block to be better. Screen Shot 2023-07-30 at 11 49 39 PM
> > > > > 
> > > > > 
> > > > > Should UserInput return null instead of 'canceled' without input?
> > > > 
> > > > 
> > > > Doesn’t really matter. If there’s no response it will return empty; just like the current ask and wait block
> > > 
> > > 
> > > It does matter, actually. For parity with the actual ask block, you should return an empty string.
> > 
> > 
> > No that’s what I mean, it will return an empty string when there’s no input
> 
> It should also do that when it's cancelled. Make a boolean that is toggled when an input is cancelled or not.

I might add the Boolean for that in a later update. I like the cancel button the way it is thought since the button makes the input report cancelled. That way people can use better input as a button selection thing rather than an input and press OK. That’s also why the Input and cancel button has an option to be enabled or disabled.

SharkPool-SP:
Extension was updated. Merged some blocks, changed some names and functions, added the ability to enable/disable 2 more buttons and fixed some bugs.

SaFireGalaxy:
Hey sharkpool? Can you possibly make it so you can force-submit the input? This would be handy for having the "enter" key be another option to submit your inputs.

SurvExe1Pc:
> Hey sharkpool? Can you possibly make it so you can force-submit the input? This would be handy for having the "enter" key be another option to submit your inputs.

its ask and wait which means it should wait, and force submitting the input without the users permission could be annoying due to it having the off chance to force submit a unfinished sentence word, or anything that was typed, i say this because it could be multi-line. if possible it should just be "shift-enter" to submit, or like in the console set it to "enter" and to add a new line set it to "shift-enter"

SaFireGalaxy:
That's a good suggestion. Maybe also, Shift enter could create a new line and enter could submit it. It will act as the submit button, but not a replacement of it.

SharkPool-SP:
> That's a good suggestion. Maybe also, Shift enter could create a new line and enter could submit it. It will act as the submit button, but not a replacement of it.

Hi, so I read the past couple comments. I’ll consider the shift enter thing. I’ll either add it or add a block that allows/doesn’t allow enter to submit.

SaFireGalaxy:
Thanks so much! This beats the ask and wait block in every possible way now, I won't have to use that block anymore.

SharkPool-SP:
> Thanks so much! This beats the ask and wait block in every possible way now, I won't have to use that block anymore.

Thanks :)

SharkPool-SP:
> > > > > > Adds various blocks for custom CSS inputs. Focused on remodeling the ask () block to be better. Screen Shot 2023-07-30 at 11 49 39 PM
> > > > > 
> > > > > 
> > > > > Should UserInput return null instead of 'canceled' without input?
> > > > 
> > > > 
> > > > Doesn’t really matter. If there’s no response it will return empty; just like the current ask and wait block
> > > 
> > > 
> > > It does matter, actually. For parity with the actual ask block, you should return an empty string.
> > 
> > 
> > No that’s what I mean, it will return an empty string when there’s no input
> 
> It should also do that when it's cancelled. Make a boolean that is toggled when an input is cancelled or not.

No need for a boolean

SharkPool-SP:
> Thanks so much! This beats the ask and wait block in every possible way now, I won't have to use that block anymore.

Updated! Added that block and some more 👀

extensions/SharkPool/BetterInput.js
```diff
@@ -0,0 +1,1305 @@
...
+            blockType: Scratch.BlockType.REPORTER,
+            text: "box count",
+          },
+          {
+            opcode: "getMaxCount",
+            blockType: Scratch.BlockType.REPORTER,
+            text: "box limit",
+          },
+        ],
+        menus: {
```
GarboMuffin, line 457: all of the menus should be acceptReporters: true unless they have a good reason not to

extensions/SharkPool/BetterInput.js
```diff
@@ -0,0 +1,1305 @@
...
+          alignmentMenu: ["left", "right", "center"],
+          fontMenu: [
```
GarboMuffin, line 459: A vast majority of these fonts will not work for a *lot* of our users (anyone on ChromeOS, macOS, Linux)

Do what animated text does - stick to the standard scratch fonts. When custom fonts are merge then you can use those too. As long as it's acceptReporters: true, people will be able to manually drop in whatever they want anyways

extensions/SharkPool/BetterInput.js
```diff
@@ -0,0 +1,1305 @@
...
+    setPrePosition(args) {
+      this.textBoxX = args.X;
+      this.textBoxY = args.Y * -1;
+    }
+
+    setPosition(args) {
+      this.textBoxX = args.X;
+      this.textBoxY = args.Y * -1;
+
+      const overlays = document.querySelectorAll(".ask-box");
```
GarboMuffin, line 664: querySelectorAll is slow and prone to collisions. it would be best if you could track those elements yourself

extensions/SharkPool/BetterInput.js
```diff
@@ -0,0 +1,1305 @@
...
+        case "Dropdown":
+          this.DropdownText = text;
+          break;
+      }
+    }
+
+    askAndWaitForInput(args) {
+      if (this.askBoxCount < this.maxBoxCount) {
+        return new Promise((resolve) => {
+          this.askAndWait(args).then(() => {
```
GarboMuffin, line 757: pretty sure the whole new Promise thing isn't required since the thing you're wrapping is, itself, a promise

extensions/SharkPool/BetterInput.js
```diff
@@ -0,0 +1,1305 @@
...
+          ) {
+            requestAnimationFrame(step);
+          }
+        };
+
+        if (this.enterEffect === "growIn" || this.enterEffect === "Both") {
+          overlay.style.transition = "opacity 0.3s";
+        }
+
+        requestAnimationFrame(step);
```
GarboMuffin, line 826: are you able to do this with CSS transitions and animations instead?

extensions/SharkPool/BetterInput.js
```diff
@@ -0,0 +1,1305 @@
...
+        };
+
+        requestAnimationFrame(step);
+      } else {
+        document.body.removeChild(overlay);
+      }
+    }
+
+    askAndWait(args) {
+      if (this.askBoxCount < this.maxBoxCount) {
```
GarboMuffin, line 879: the normal Scratch ask blocks will wait for the old prompt to disappear, then show the new one. would it make sense to do that here too?

extensions/SharkPool/BetterInput.js
```diff
@@ -0,0 +1,1305 @@
...
+          overlay.style.backgroundColor = this.textBoxColor;
+          overlay.style.boxShadow = "0 0 5px rgba(0, 0, 0, 0.3)";
+          overlay.style.borderRadius = this.textBoxBorderRadius + "px";
+          overlay.style.padding = "15px";
+          overlay.style.fontSize = this.fontSize;
+          overlay.style.textAlign = this.textAlign;
+          overlay.style.fontFamily = this.fontFamily;
+
+          const overlayImageContainer = document.createElement("div");
+          overlayImageContainer.style.background = `url(${this.overlayImage})`;
```
GarboMuffin, line 907: Scratch.canFetch, wrap the URL with quotes and pass it into encodeURI()

GarboMuffin (changes requested):
i left some comments

extensions/SharkPool/BetterInput.js
SharkPool-SP: I can remove the custom fonts stuff. Ill also try to have accept reporters in the next update whenever it comes. If you are merging this file right now, are you modifying that stuff or am I?

extensions/SharkPool/BetterInput.js
SharkPool-SP: I need it to be that way so it can properly handle the various inputs. That including the ask and wait reporter block... If Im wrong, what can I do?

extensions/SharkPool/BetterInput.js
SharkPool-SP: No need for it with the new block in this update. Since I want to allow people to have more than 1 ask box, I just added a <is waiting for input?> boolean instead. If users really wanted to wait for the old ask input to disappear before asking for the new one, they can make it themselves with the ask block. Its better this way since this extension is supposed to be a fully customizable expansion + remodel of the Ask block.

extensions/SharkPool/BetterInput.js
SharkPool-SP: Yeah sorta. Not exactly exit effects... I was planning this for a later update though since its complex to make for exiting. In my opinion its best to leave this as is until the next update rather than rushing and pushing it out... right?

extensions/SharkPool/BetterInput.js
SharkPool-SP: Saving this for last...

SharkPool-SP:
> Thanks so much! This beats the ask and wait block in every possible way now, I won't have to use that block anymore.

Thanks!!! Updates coming soon :)))

extensions/SharkPool/BetterInput.js
SharkPool-SP: Fixed

extensions/SharkPool/BetterInput.js
SharkPool-SP: Screw it, Im just gonna do it now.

SharkPool-SP:
Fixed all issues and added new stuff!
Screen Shot 2023-08-18 at 11 56 56 PM

extensions/SharkPool/BetterInput.js
GarboMuffin: This was not resolved. You still do not use Scratch.canFetch()

extensions/SharkPool/BetterInput.js
GarboMuffin: This was not resolved

extensions/SharkPool/BetterInput.js
GarboMuffin: You still use querySelectorAll in removeAskBoxes

extensions/SharkPool/BetterInput.js
GarboMuffin: You can do custom fonts now. See animated text

SharkPool-SP:
**3.0 is RELEASED!!!**

In this Update:
-Bug Fixes
-Infinite Buttons
-Image Setting for More Elements
-**ALL** Elements update live
-Vertical Sliders
-Updated Palette

SharkPool-SP:
Small Update:
image

SharkPool-SP:
image
Another Update

horizontal-shading:
That #1029 issue wasn't a real issue. It was me complaining...

Thebloxers998:
How do I change the background colour in the extension ask ui

Thebloxers998:
I have a suggestion, you should make it so we can have more than one input typing thing

SharkPool-SP:
> I have a suggestion, you should make it so we can have more than one input typing thing

With the way I built the extension, no. As a alternative just make multiple boxes

SharkPool-SP:
marking as draft as im rewriting this from scratch

SharkPool-SP:
https://github.com/TurboWarp/extensions/pull/1940

Issue #674, opened 2023-07-02
Request for comment: On utilities extensions


GarboMuffin:
I am considering a requirement that new extensions (ie ones that are not already merged) must have a single unified theme that all the blocks adhere to. This would affect most broad utilities extensions.

Motivations:

 - Discoverability is zero if the extension does not actually describe what it does.
 - Utilities extensions are frequently bloated with blocks that are completely unrelated to what most people will actually use.
 - In an attempt to reduce that bloat, utilities extensions tend to be very shallow and can only afford a small handful of blocks per category.
 - It results in duplication of functionality across several extensions, sometimes in subtly different or broken ways.
 - It allows bugs to remain hidden for longer, especially in duplicated functionality.
 - Making unique images for extensions is more difficult when the extension doesn't have a unified idea.
 - The time available to review new extensions and maintain the existing ones is finite. I would like that time to be spent on extensions that do something new.
 - They take longer to review as the entire thing needs to be reviewed at the same time instead of in chunks.

NexusKitten:
Support, I don't believe utilities extension are inherently bad and there's a few decent ones, however it's far more convenient both from a reviewer and user side to have more specific extensions such as Math, Couplers (#219), and More Comparisons (#214) that you can add as you need, rather than adding a bunch of utilities extensions and hoping it covers all your bases.

SurvExe1Pc:
so they need a base of what they actually do to make it easier, so split them into different extensions?

SurvExe1Pc:
i understand the issue, as it can be hard to make an image based on an extension that has really no base.

SurvExe1Pc:
should I close my PR?

GarboMuffin:
Other mitigating factors that make this not a large problem:

 - Our extension library has grown *massively*. There are fewer gaps to fill than ever.
 - You can just load extensions from files now. They get stored in the project.

> so they need a base of what they actually do to make it easier, so split them into different extensions?

Yes, split them into separate extensions. If another extension in the same category exists, consider adding to that one instead of making a new one.

> i understand the issue, as it can be hard to make an image based on an extension that has really no base.

Yes, see the RixxyX for an example that don't really represent the functionality because it's just too broad

> should I close my PR?

It's only been 30 minutes so this discussion is not closed. However I would encourage you to pick specific parts of 0tils to focus on and make extensions that go deep in each specific area.

SurvExe1Pc:
should I just add my extra hashes, and the AES function into encoding?

[...]

PR #822, opened 2023-07-31
Big miscalleneous reporter extension

BunkerStriker:


LilyMakesThings:
This seems reminiscent of a Utilities extension, there's too much variety which ends up hindering a lot of its features.

BunkerStriker:
I've at first made it for myself but wanted to share it, so i did.

BunkerStriker:
I couldn't really just choose a theme so i just made everything i thought of as useful with help from ChatGPT on the more complex blocks

NexusKitten:
There's a few things I noticed about this extension:

1. The theme of "reporters" is vague enough to be considered a "utilities" extension. As @LilyMakesThings pointed out, broad "utilities" extensions aren't very likely to get accepted to the gallery. For a full list of reasons why, see #674.
2. Several blocks from your extension already exist in other extensions. Although there are exceptions, we generally try to add a block once to the extension gallery. Some of the duplicate blocks include:
- Both direction + distance blocks will soon be in Graphics 2D.
- Almost all blocks from Couplers appear in the extension
- Several static blocks, the clamp block, and the "if <> () else ()" block appear in Utilities.
- Almost all JSON blocks appear in the JSON extension.
3. The only block I found with particularly confusing behavior was the block listed below, which seems to strangely prioritize whether it reports as a boolean or a reporter, and can return an empty output if the second setup below is used (empty outputs can break projects).
image
image
4. The extension color is very bright, which may be difficult for visually impaired people to read. Consider darkening the extension colors if possible.
5. Pertaining to the extension ID (Line 32), generally it's better to include your name or initials somewhere in there. For example, "bsreportersbucket" would work.
When it comes to Utilities extensions, I generally suggest splitting up the major functions not already in other extensions to their own extensions. I'm not very experienced with gaussian, however I saw you had multiple blocks pertaining to that, and I'm relatively certain there's not already a gaussian extension on the gallery, so that would make a good extension of it's own. If there's any functions that fit under the umbrella of an extension, however are not currently added, feel free to submit a pull request to add those blocks. The extension gallery is open source, so you're free to submit pull requests to anything as you see fit.

GarboMuffin:
Are there any unique blocks in this extension that are not found in other ones?

NexusKitten:
> Are there any unique blocks in this extension that are not found in other ones?

image

GarboMuffin:
- map can go in utilities (Arduino has a similar function, it is quite useful)
 - PRNG should be expanded into a dedicated PRNG extension
 - Probability stuff should be expanded into a dedicated probability extension. There is a *lot* that can be done here.
 - i do not anticipate matrix at or wait then to be useful anywhere

PR #826, opened 2023-08-01
added websocket extension

RedMan13:


RedMan13:
to be clear this is just to get rid of the junk commits inside of the original pr

GarboMuffin:
they all get squashed down anyways but ok

RedMan13:
> they all get squashed down anyways but ok

oh.

extensions/godslayerakp/ws.js
```diff
@@ -0,0 +1,283 @@
...
+  class WebRequests {
+    /**
+     * no need to install runtime as it comes with Scratch var
+     */
+    constructor() {
+      this.instances = {};
+    }
+    getInfo() {
+      return {
+        id: "websocket",
```
GarboMuffin, line 22: do you have a good reason why you want `websocket` instead of something based on your username like other extensions?

GarboMuffin:
This immediately results in an error

image

extensions/godslayerakp/ws.js
```diff
@@ -0,0 +1,283 @@
...
+            this.isMessage = false;
+            return oldState;
+          },
+          state: "opening",
+          websocket,
+        };
+        websocket.onopen = (e) => {
+          instance.state = "opened";
+          instance.isOpen = true;
+          utils.startHats("websocket_onOpen", null, utils.target);
```
GarboMuffin, line 194: The only safe time to use the block utility is immediately, in a syncronous script. You should use runtime.startHats instead

extensions/godslayerakp/ws.js
```diff
@@ -0,0 +1,283 @@
...
+              },
+            },
+            text: "close connection with reason [REASON] and code [CODE]",
+          },
+        ],
+      };
+    }
+    async newInstance(args, utils) {
+      let url = Cast.toString(args.URL);
+      if (!url.startsWith("ws")) {
```
GarboMuffin, line 153: This will fail on inputs like `wstestexample.com`

extensions/godslayerakp/ws.js
GarboMuffin: Similarly, you need to store `utils.target` in a local variable somewhere because the object gets reused for every script

GarboMuffin:
I would also love to see some attempt to make it so that multiple messages received between frames are all given an opportunity to be processed

extensions/godslayerakp/ws.js
```diff
@@ -0,0 +1,283 @@
...
+        color1: "#307eff",
+        color2: "#2c5eb0",
+        blocks: [
+          {
+            opcode: "newInstance",
+            blockType: BlockType.COMMAND,
+            arguments: {
+              URL: {
+                type: ArgumentType.STRING,
+                defaultValue: "wss://example.com",
```
GarboMuffin, line 35: also, if you happen to know of a public WebSocket echo server of some sort, consider putting that here instead

RedMan13: > also, if you happen to know of a public WebSocket echo server of some sort, consider putting that here instead

i dont but if you have any suggestions i would be fine with using them

extensions/godslayerakp/ws.js
RedMan13: > This will fail on inputs like `wstestexample.com`

...
yeah i didnt think about that

extensions/godslayerakp/ws.js
RedMan13: > Similarly, you need to store `utils.target` in a local variable somewhere because the object gets reused for every script, so right now none of these events work if any other sprites have running scripts

oh...

RedMan13:
> This throws an error:
> 
> image
> 
> image
> 
> You can argue I'm using it wrong (which I would consider poor block design), but regardless it must never be throwing errors

fair on the errors
should i make it so that the send block waits on the connect event for the script then?

extensions/godslayerakp/ws.js
RedMan13: > do you have a good reason why you want `websocket` instead of something based on your username like other extensions?

no... just didnt think about it

GarboMuffin:
yeah probably

extensions/godslayerakp/ws.js
GarboMuffin: (utils.startHats is actually probably safe to use, but strictly you shouldnt really be touching it at all at this point, so try not to use it)

RedMan13:
> > should i make it so that the send block waits on the connect event for the script then?
> 
> yeah probably

kk

extensions/godslayerakp/ws.js
GarboMuffin: ok, thanks, just wanted to make sure compatibility with another Scratch mod was not one of the goals

RedMan13:
> > should i make it so that the send block waits on the connect event for the script then?
> 
> yeah probably

done

extensions/godslayerakp/ws.js
RedMan13: for now i think ima just make my own with replit for now :/

extensions/godslayerakp/ws.js
```diff
@@ -148,72 +180,124 @@
...
+        if (!/^(ws|wss):/is.test(url)) {
+          // url doesnt start with a valid connection type
+          // so we jsut assume its formated without it
+          if (/^(?!(ws|http)s?:\/\/).*$/is.test(url)) {
+            url = `wss://${url}`;
+          } else if (/^(http|https):/is.test(url)) {
+            url = url
+              .split(":")
+              // @ts-ignore
+              .with(0, url.startsWith("https") ? "wss" : "ws")
```
GarboMuffin, line 209: for browser support reasons don't use with()

RedMan13:
dead chat :trollface:

GarboMuffin:
not dead!

RedMan13:
> not dead!

image

GarboMuffin:
if you have no complaints about the rewrite I'll merge it this week

RedMan13:
> if you have no complaints about the rewrite I'll merge it this week

alr

RedMan13:
yeah i see nothing wrong

RedMan13:
a

PR #831, opened 2023-08-04
Just added my "Weather" extension

samuellouf:
I added a "Weather" extension.
It's inspired from the ScratchX "Weather" extension. ([Docs here](https://khanning.github.io/scratch-weather-extension/)).
Blocks


samuellouf:
Ok

Issue #431, opened 2023-05-18
Please make a movie/video extention


Zigotonlover08:
Please make a movie/video extention, it would been very useful to add and play video files, you can add cutscenesfor your games without using 9999999 costumes just to add a video. This is my request/suggestion

idoinfactexist:
FINALLY!!

gomiigg:
how would this work though

TheShovel:
> > This is possible since it is possible to render videos to canvas. You could just start a hiden video player and take the current frame as a sprite costume skin. The audio would naturally come from the video player.
> 
> Ok great, are you going to make it?

Sadly, I don't have time.

Zigotonlover08:
> > > This is possible since it is possible to render videos to canvas. You could just start a hiden video player and take the current frame as a sprite costume skin. The audio would naturally come from the video player.
> > 
> > 
> > Ok great, are you going to make it?
> 
> Sadly, I don't have time.

Maybe next time?

lopste:
This seems fun to take a crack at. I'll try to actually read documentation on Scratch internals this time. What would be better? Replacing the sprite or putting a separate video in?

Zigotonlover08:
> This seems fun to take a crack at. I'll attempt to read the documentation on Scratch internals this time. What would be better? Replacing the sprite or putting a separate video in? Additionally, would you load the video from a link? That seems the most plausible but poses a few additional questions.

Idk maybe replacing sprite

LukeManiaStudioss:
Replacing the sprite would allow for more flexibility with how it's rendered on the stage, because you could just apply any motion blocks onto the sprite as opposed to having arbitrary blocks to move the video. That's how ScratchLab's "Animated Text" worked.

Zigotonlover08:
> Replacing the sprite would allow for more flexibility with how it's rendered on the stage, because you could just apply any motion blocks onto the sprite as opposed to having arbitrary blocks to move the video. That's how ScratchLab's "Animated Text" worked.

Sounds cool, i can't wait for someone to make it a reality

Martinelplayz:
this needs to happen so bad! i wish there was a way to store mp4 files in projects! even better, to load straight from yt videos. i could upload a game cutscene as unlisted and have it play with the extension. or even better i bet i could use it with an active online multiplayer game whenever there is some sort of special event going on. i could use it with the cloudlink extension to update the video being played for everyone to see at the same time.

Zigotonlover08:
[...]

PR #840, opened 2023-08-06
Video: Add Video extension

LilyMakesThings:
## Video Extension
Simple extension to play video files from URLs. Wip.

Closes https://github.com/TurboWarp/extensions/issues/431

image

## Things I WILL NOT add:
- YouTube support


penta-quark-neutro:
it is extremely impressive.
I loaded my videos and with 350MB I couldn't give it.
1.18GB broke Turbowarp.
It's great.

CST1229:
> Can you explain how this works?

From reading the code, it just seems to put a HTML `<video>` element on top of the stage canvas.

samuellouf:
Your extension is awesome. I think I'm going to edit it to make a iFrame extension.

samuellouf:
But it only works on TurboWarp.

CubesterYT:
Damn this is insane, I completely forgot the video html tag exists. I thought it wouldn't be possible without an iframe or something.

LilyMakesThings:
If anyone has any idea why this doesn't work packaged, let me know

CubesterYT:
> If anyone has any idea why this doesn't work packaged, let me know

I have a theory, A TURBOWARP THEORY. I think the packager only packages the project part of the html only. It omits any other part of the html code when packaging, because all you're doing is making the video element sit on top of the project itself, if the video is going to be packaged, then it needs to be a part of the stage.

CST1229:
> If anyone has any idea why this doesn't work packaged, let me know

I do know that the packager's project player works somewhat differently from vanilla Scratch, so that might be why.

GarboMuffin:
Something to consider is that you should be able to use a video element in renderer.updateBitmap to upload the current frame (which you could then repeat each frame). It'll be less efficient but you would be able to integrate with the rest of Scratch, same as Skins

LilyMakesThings:
> Something to consider is that you should be able to use a video element in renderer.updateBitmap to upload the current frame (which you could then repeat each frame). It'll be less efficient but you would be able to integrate with the rest of Scratch, same as Skins

I did try this, but I couldn't quite figure it out. And in this case, how would the block palette even look?

CubesterYT:
> This is the only extension I've worked on recently where I've had no idea what the palette was going to look like, so feedback/suggestions from anyone would be greatly appreciated!

Some suggestions, rename "resume video" to "play video", it makes more sense in terms of its functionality. And in terms of new blocks, it would be great if there are blocks that allow you to set a size to the video player and a location to the video player, like "resize video to width height" and "change video x y" something like that.

BlueToadMakerr:
[0.sb3.zip](https://github.com/TurboWarp/extensions/files/12314983/0.sb3.zip)

samuellouf:
> If anyone has any idea why this doesn't work packaged, let me know

I know a way to make your extension work on single html files, .exe files, .app files ...
`this.video.style.zIndex=1;`

I reality, the video is loaded on the page but it's 𝑥 layers behind the canvas.
I had the same problem with an iframe extension I'm currently coding.

So you can resolve the problem by adding:
`this.video.style.zIndex=1;`
in the `constructor(){}` after the video item is created.
It will change the video element's layer.

So, I think it should do the trick.

LilyMakesThings:
> > If anyone has any idea why this doesn't work packaged, let me know
> 
> I know a way to make your extension work on single html files, .exe files, .app files ... `this.video.style.zIndex=1;`
> 
> I reality, the video is loaded on the page but it's 𝑥 layers behind the canvas. I had the same problem with an iframe extension I'm currently coding.
> 
> So you can resolve the problem by adding: `this.video.style.zIndex=1;` in the `constructor(){}` after the video item is created. It will change the video element's layer.
> 
> So, I think it should do the trick.

I've opted to just rendering it onto sprites now

LilyMakesThings:
Changes:
- Video now renders with skin API onto the stage (temporary)
- Volume is dictated by the audioEngine's gain node (Scratch Addons)
- Control blocks removed
- Show/hide blocks removed

To do:
- Think of a logical block palette for this extension
  - How are width/height going to be defined?
  - Should multiple videos be allowed to load?
  - How are videos going to render on targets?
  [The obvious answer to all of this is to copy Skins' blocks]
- Somehow clamp the video width/height

LilyMakesThings:
## Changes:
- Updated block palette
- Allows multiple videos to be loaded
- Allows videos to render onto multiple targets

## To do:
- Find out why `updateAllDrawableProperties` doesn't work
- Get the width and height of the video and make a reporter for it
- Update drawables when a video is deleted
- Add more QOL blocks
  - Reporter for loaded videos
  - Boolean for if a video is loaded

image

LilyMakesThings:
I'm not entirely sure how to get the width/height of the video reliably; using the texture width/height isn't entirely reliable.

I'm thinking of making a fake skin the moment the video is loaded, grabbing the texture width/height, then destroying the skin. It's hacky, but at least it'd work. Thoughts?

LilyMakesThings:
## Changes (Ready for review)
This extension now creates brand new bitmap skins for videos, which makes getting width/height easier and preserves the skin data for existing skins as opposed to overwriting them. 

## Final Palette
image

gomiigg:
This is a really complex but versatile extension! One question though. How would I be able to use this for local files? Let's say I want an in-game cutscene or preloader intro. Where could I place the file in the packaged project and what would I need to enter in the URL input?

LilyMakesThings:
> This is a really complex but versatile extension! One question though. How would I be able to use this for local files? Let's say I want an in-game cutscene or preloader intro. Where could I place the file in the packaged project and what would I need to enter in the URL input?

If it's an mp4, ironically the best solution is to change the file extension .mp4 to .wav, then upload it as a sound (it doesn't cut any of the video from it). Then, use an extension like Asset Manager or Sound Expanded (somewhere on the PRs) to get the sound as a data URI; they both have a block that can do that. Then use that block inside of the URL input.

Alternatively, use the files extension's "upload file as dataURL" block (something like that) and add the result to a list or something, then use that item from the list.

gomiigg:
> > This is a really complex but versatile extension! One question though. How would I be able to use this for local files? Let's say I want an in-game cutscene or preloader intro. Where could I place the file in the packaged project and what would I need to enter in the URL input?
> 
> If it's an mp4, ironically the best solution is to change the file extension .mp4 to .wav, then upload it as a sound (it doesn't cut any of the video from it). Then, use an extension like Asset Manager or Sound Expanded (somewhere on the PRs) to get the sound as a data URI; they both have a block that can do that. Then use that block inside of the URL input.
> 
> Alternatively, use the files extension's "upload file as dataURL" block (something like that) and add the result to a list or something, then use that item from the list.

Hey that second one works better than I thought. The data shows as an INSANELY long text string. Does that affect how much memory or storage the project uses?

LilyMakesThings:
> > > This is a really complex but versatile extension! One question though. How would I be able to use this for local files? Let's say I want an in-game cutscene or preloader intro. Where could I place the file in the packaged project and what would I need to enter in the URL input?
> > 
> > 
> > If it's an mp4, ironically the best solution is to change the file extension .mp4 to .wav, then upload it as a sound (it doesn't cut any of the video from it). Then, use an extension like Asset Manager or Sound Expanded (somewhere on the PRs) to get the sound as a data URI; they both have a block that can do that. Then use that block inside of the URL input.
> > Alternatively, use the files extension's "upload file as dataURL" block (something like that) and add the result to a list or something, then use that item from the list.
> 
> Hey that second one works better than I thought. The data shows as an INSANELY long text string. Does that affect how much memory or storage the project uses? Is it essentially saving the video but in text form or is it compressed?

It's essentially saving the video in text form. I don't really have a solution for that.

gomiigg:
> > > > This is a really complex but versatile extension! One question though. How would I be able to use this for local files? Let's say I want an in-game cutscene or preloader intro. Where could I place the file in the packaged project and what would I need to enter in the URL input?
> > > 
> > > 
> > > If it's an mp4, ironically the best solution is to change the file extension .mp4 to .wav, then upload it as a sound (it doesn't cut any of the video from it). Then, use an extension like Asset Manager or Sound Expanded (somewhere on the PRs) to get the sound as a data URI; they both have a block that can do that. Then use that block inside of the URL input.
> > > Alternatively, use the files extension's "upload file as dataURL" block (something like that) and add the result to a list or something, then use that item from the list.
> > 
> > 
> > Hey that second one works better than I thought. The data shows as an INSANELY long text string. Does that affect how much memory or storage the project uses? Is it essentially saving the video but in text form or is it compressed?
> 
> It's essentially saving the video in text form. I don't really have a solution for that.

Well as long as it's not going over any memory or storage limits I'm fine with that

extensions/Lily/Video.js
```diff
@@ -0,0 +1,441 @@
...
+          const videoObject = this.videos[name];
+
+          const video = videoObject.video;
+          const videoVolume = videoObject.volume;
+          const skin = videoObject.skin;
+
+          const projectVolume = runtime.audioEngine.inputNode.gain.value;
+          video.volume = videoVolume * projectVolume;
+
+          vm.renderer.updateBitmapSkin(skin, video, 1);
```
GarboMuffin, line 61: Will this be constantly reuploading videos that are paused and not even visible?

extensions/Lily/Video.js
```diff
@@ -0,0 +1,441 @@
...
+        }
+
+        for (const id of Object.keys(this.targets)) {
+          const target = this.targets[id].target;
+          const drawableID = target.drawableID;
+
+          const drawable = runtime.renderer._allDrawables[drawableID];
+          // This was only a problem when targets weren't reset, I don't
+          // expect it to happen much now but just in case..
+          if (!drawable) return;
```
GarboMuffin, line 71: I think you mean to `continue`

extensions/Lily/Video.js
LilyMakesThings: ...probably

LilyMakesThings: I added a check for if any drawable has the video applied or if the video is paused. I encourage you to check it yourself in-editor.

LilyMakesThings:
Should I go the extra mile and check if any drawables are visible (ie, shown/hidden) before updating the bitmap? Or is it fine as-is

GarboMuffin:
Closes https://github.com/TurboWarp/extensions/pull/840

GarboMuffin:
last thing here - does it make sense to wait for the video to load at least a tiny bit before "load video from URL" completes?

LilyMakesThings:
> last thing here - does it make sense to wait for the video to load at least a tiny bit before "load video from URL" completes?

I think so, yes.

daisytheumbreon-B3:
how do i add a youtube video?

CubesterYT:
@daisytheumbreon-B3, use the iframe extension for YouTube videos.

Forysium:
can this extension somehow access local videos? Or is there another extension out there for this?

CST1229:
> can this extension somehow access local videos? Or is there another extension out there for this?

you could probably combine this with the Files extension
image

Forysium:
Thank you @CST1229 :D

PR #845, opened 2023-08-07
Experiment: Hacked Block Collection Extension

LilyMakesThings:
(this is one of many extensions I'm making with the brand new blocktype)
image
If there's a hacked block you think would be useful here, let me know!

## You can test this extension on staging.turbowarp.org


SurvExe1Pc:
xml block type :D

NexusKitten:
Personally I'd put "hacked" in quotations because it's somewhat inaccurate and sounds edgy lol

LilyMakesThings:
That's a really good point, honestly might be worth just renaming the extension entirely

horizontal-shading:
I need this

GarboMuffin:
sure we can make this happen

LilyMakesThings:
> sure we can make this happen

Is the list here sufficient or do you have any other ideas?

GarboMuffin:
if anyone asks they will be added

LilyMakesThings:
This and "All Menus" could benefit from a way to auto-load extensions, given that they don't have any extension-made blocks to keep them in the palette.

pumpkinhasapatch:
@LilyMakesThings A few suggestions on how you could improve this:

- The hacked sensing blocks like `key ( ) pressed` and `touching ( )` can be easily recreated by dragging a join block into their normal variants. I don't think they need to be included in the extension.
image

- The normal `when this sprite touches ( )` menu only lists mouse pointer and edge, but the block also works with sprites if you put a blue `sensing_touchingobjectmenu` input inside it, as you can see in [this project](https://scratch.mit.edu/projects/401109316/editor).
image

- The `while` loop is already present in TurboWarp's Control category, being the only block changed in the entire vanilla block palette since the `for` loop's removal. It's odd that the exact same block is listed twice when using this extension.

I would suggest to only include blocks that are impossible to get in TurboWarp without manually editing project.json. You've done a good job at bringing some of these hidden Scratch gems to a wider audience.

NexusKitten:
> @LilyMakesThings A few suggestions on how you could improve this:
> 
> * The hacked sensing blocks like `key ( ) pressed` and `touching ( )` can be easily recreated by dragging a join block into their normal variants. I don't think they need to be included in the extension.
>   image
> * The normal `when this sprite touches ( )` menu only lists mouse pointer and edge, but the block also works with sprites if you put a blue `sensing_touchingobjectmenu` input inside it, as you can see in [this project](https://scratch.mit.edu/projects/401109316/editor).
>   image
> * The `while` loop is already present in TurboWarp's Control category, being the only block changed in the entire vanilla block palette since the `for` loop's removal. It's odd that the exact same block is listed twice when using this extension.
> 
> I would suggest to only include blocks that are impossible to get in TurboWarp without manually editing project.json. You've done a good job at bringing some of these hidden Scratch gems to a wider audience.

Moved this to issue #960

PR #871, opened 2023-08-10
Create colortools.js

DaTripleBlock:
Add colors tools makes colors easier :D


GarboMuffin:
None of the conversion blocks in this extension are useful. I'm not sure what this is supposed to be used for.

PR #757, opened 2023-07-17
true-fantom/couplers: add more couplers

BlueDome77:
Added COLOR, ANGLE, MATRIX, and NOTE reporters from LMS Utils.

PR #879, opened 2023-08-12
Update comment blocks.

SurvExe1Pc:
Adds matrix comments, and also color comments.
sadly since C-Blocks don't work i could not add them 😭.

### DIFF
--- a/extensions/Lily/CommentBlocks.js
+++ b/extensions/Lily/CommentBlocks.js
@@ -61,7 +61,37 @@
                 type: Scratch.ArgumentType.BOOLEAN
               }
             }
-          }
+          },
+          {
+            opcode: 'commentMatrix',
+            blockType: Scratch.BlockType.REPORTER,
+            text: '[INPUT] // [COMMENT]',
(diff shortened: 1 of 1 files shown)


GarboMuffin:
@LilyMakesThings this is your extension

do you agree with this?

LilyMakesThings:
I don't like this. I'd prefer people to use a coupler for these values.

NexusKitten:
> I don't like this. I'd prefer people to use a coupler for these values.
> 
> Adding a comment to a color or matrix value is so niche that, for the one person who is **going** to do that, they should just use the extra block to do that if it keeps this palette to a minimum.

It's worth noting there's not currently a matrix/color coupler in any extensions (other than LMS utils, which is currently unlisted). If these blocks were in couplers i'd be inclined to agree, but as it stands, they are not.

GarboMuffin:
related: https://github.com/TurboWarp/extensions/pull/757

GarboMuffin:
Similar blocks are in couplers now

PR #881, opened 2023-08-13
Block to get details of gamepads

### LINKED ISSUE #147: Gamepad: Add a way to tell what kind of gamepad you're using
This is useful because different controllers have different button layouts. PlayStation controllers have face buttons with shapes where Xbox controllers have the A, B, X, and Y buttons. You could use this ability to make projects show buttons from the player's particular controller.

DNin01:
Resolves #147

Adds a block to the Gamepad extension that allows you to get information about connected gamepads.

At some point, I'd like to add a way to get the model (like is this an Xbox One controller or a PlayStation DualShock or...?) so you don't ever have to write your own code to use the vendor or product to tell what kind of controller it is, you just leave it up to the block and it'll tell you. I've already done some work on such a function, I just need to find more information so it works on more controllers.

It'd also be nice to add cross-browser compatibility. Currently, the vendor and product matching functions only work properly on Chromium-based browsers. For Firefox, see [Gamepad: id property](https://developer.mozilla.org/en-US/docs/Web/API/Gamepad/id).

Tested in Edge 115.

### DIFF
--- a/extensions/gamepad.js
+++ b/extensions/gamepad.js
@@ -71,6 +71,14 @@
     return axisValue;
   };
 
+  const matchVendor = (id) => {
+    return id.match(/vendor:\s*(\w+)/i)[1];
+  };
+
+  const matchProduct = (id) => {
+    return id.match(/product:\s*(\w+)/i)[1];
+  };
+
   class GamepadExtension {
     getInfo() {
       return {
@@ -89,6 +97,26 @@
               }
             }
           },
+          {
+            opcode: 'gamepadDetail',
+            blockType: Scratch.BlockType.REPORTER,
+            text: 'get [d] of pad [i]',
+            arguments: {
+              d: {
+                type: Scratch.ArgumentType.STRING,
+                defaultValue: 'id',
+                menu: 'detailMenu'
+              },
+              i: {
+                type: Scratch.ArgumentType.NUMBER,
+                defaultValue: '1',
+                menu: 'padMenu'
+              }
+            }
+          },
+
+          '---',
+
           {
             opcode: 'buttonDown',
             blockType: Scratch.BlockType.BOOLEAN,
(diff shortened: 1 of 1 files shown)


GarboMuffin:
I really do not like the idea of exposing vendor IDs directly to projects.

There are some example IDs on https://hardwaretester.com/gamepad, notably you have things like:

 - 054c-09cc-Wireless Controller
 - xinput
 - Wireless Controller Extended Gamepad

which aren't even *that* uncommon.

would much rather have the extension figure out all the details that matter instead

Martinelplayz:
would you be ok with me keeping my PR up?

DNin01:
> would you be ok with me keeping my PR up?

@Martinelplayz To sail smoothly, I think it makes more sense **not** to make PRs that combine work from other PRs unless they are **specifically meant to supersede and improve upon them** since, you know, if either yours or mine gets merged as they are, one of us will need to fix conflicts. Merging the two PRs separately will only cause a conflict if both edit the same line or something like that. Merging separately will work just as well and allow us to make edits as we see fit (there's a reason this PR is a draft).

Technically, you can keep it open if you want - I agree to the GPL-v3 license by submitting a PR, you've already provided credit on the PR description, and the code is not too hard for anyone to come up with. I would ask that you remove [L5](https://github.com/TurboWarp/extensions/pull/1020/files#diff-5f483f22db59792b3634c8f4063759cb213d6347385e94111e4157b81104772fR5) as you may not have written the added code, and we typically don't add comments indicating editors.

However, I will note that GarboMuffin [commented](https://github.com/TurboWarp/extensions/pull/881#issuecomment-1676468006) that it's probably a better idea to have the extension figure out what kind of controller is being used instead of exposing the small details to the project and having it figure out, so keep that in mind - it might not be accepted. You can try to make something like my idea of putting together details to return a type of controller if you want, but it can get pretty tricky as there are a lot of different controllers, from XInput to DualShock to Joy-Con controller to Stadia controller... yeah. Even I need to learn more about them.

DNin01:
If you just want to help, you can always make suggestions for the PR authors through the "Files changed" tab.

DNin01:
I don't really like making figuring out the controller model the responsibility of project creators either, so I'm going to close this pull request and propose a new implementation later.

PR #891, opened 2023-08-15
Asset Manager: Add "Asset Manager" extension.

LilyMakesThings:
Extension dedicated solely to editing assets within the project, with some similar blocks on the side.

I received explicit permission from TheShovel to make and pull request this, and we both agreed it would be better to have this as a separate extension from ShovelUtils. (Also much easier than to merge this with ShovelUtils).

Feedback is appreciated.


CubesterYT:
Let's gooooo!

echoless3484:
please add this in 10 minutes

extensions/Lily/Assets.js
```diff
@@ -0,0 +1,630 @@
+// Name: Asset Manager
+// ID: lmsAssets
+// Description: Add, remove, and get data from sprites, costumes and sounds.
```
NexusKitten, line 3: Minor nitpick, but the oxford commas in the description are inconsistent.

NexusKitten (changes requested):

extensions/Lily/Assets.js
LilyMakesThings: damn you're right

LIZARD-OFFICIAL:
You could make a DRM with this!!!

extensions/Lily/Assets.js
```diff
@@ -0,0 +1,656 @@
...
+          return sound.asset.encodeDataURI();
+        case "index":
+          return soundIndex + 1;
+        case "format":
+          return sound.asset.assetType.runtimeFormat;
+        case "header":
+          return sound.asset.assetType.contentType;
+        case "asset ID":
+          return sound.asset.assetId;
+      }
```
GarboMuffin, line 559: you gotta always return *something*

extensions/Lily/Assets.js
```diff
@@ -0,0 +1,656 @@
...
+        .then((buffer) => vm.loadProject(buffer));
+    }
+
+    getProjectJSON() {
+      return Scratch.vm.toJSON();
+    }
+
+    loadExtension(args) {
+      const url = Cast.toString(args.URL);
+      vm.extensionManager.loadExtensionURL(url);
```
GarboMuffin, line 593: you could await that

extensions/Lily/Assets.js
```diff
@@ -0,0 +1,656 @@
...
+      const targets = Scratch.vm.runtime.targets;
+      for (let index = 1; index < targets.length; index++) {
+        const target = targets[index];
+        if (target.isOriginal) {
+          const targetName = target.getName();
+          const targetId = target.id;
+          spriteNames.push({
+            text: targetName,
+            // We, rather unfortunately, have to rely on Target IDs. This is a terrible idea.
+            value: targetId,
```
GarboMuffin, line 644: what reason?

pretty sure those are entirely randomly generated each page load

LilyMakesThings: They do, that's a very good point. The problem comes from one of the blocks allowing you to rename a sprite. If the sprite changes names, then the block would suddenly not work so long as it used just the sprite's name.

LilyMakesThings: It's the same problem as List Tools, except here it's much more dire because one of the blocks from the extension can literally break itself if it used names as opposed to IDs

GarboMuffin:
i don't know why this was stalled anymore so it probably wasn't important

PR #896, opened 2023-08-16
Remove uses of window.Scratch

GarboMuffin:
Can't promise this will still work with upcoming extension loading changes


Martinelplayz:
> Can't promise this will still work with upcoming extension loading changes

 what exactly does this do though? is there something extension creators should use instead?

PR #904, opened 2023-08-20
Fix grammar and block in 'Time' extension

ziqiangao:
Block, Extract () from () was broken and was returning NaN.
Grammar also fixed in some blocks

### DIFF
--- a/extensions/-SIPC-/time.js
+++ b/extensions/-SIPC-/time.js
@@ -16,19 +16,19 @@
                     {
                         opcode: 'Timestamp',
                         blockType: Scratch.BlockType.REPORTER,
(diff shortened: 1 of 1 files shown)


LilyMakesThings:
don't have them capitalised, it's inconsistent with other extensions

ziqiangao:
Decapitalised the first letter of block names

CubesterYT:
Wait, I don't think you can change the argument type according to compatibility docs. I think you might have to make a new block with the fix. How was this not caught during it's development???

ziqiangao:
An argument type change from 'NUMBER' to 'STRING' is typically safe both ways however argument id's and block opcodes has to be the same

PR #905, opened 2023-08-20
Declarative extension list

GarboMuffin:
Say goodbye to `index.ejs` and hello to `extensions.json`

Extension metadata like titles are now stored at the start of the .js:

```js
// Name: Example Extension
// Description: Does a very cool thing. This must have punctuation at the end!
// By: GarboMuffin <https://scratch.mit.edu/users/GarboMuffin/>
// Original: TestMuffin
```

Because extensions are all defined in a parseable format, we now also generate a JSON version of the homepage, which may be useful for something else


LilyMakesThings:
You are awesome

PR #752, opened 2023-07-16
Updated Time Extension

SharkPool-SP:
MOVED: https://github.com/TurboWarp/extensions/pull/928

Added several new blocks that help calculate differences in time and data about time and calendars. Also made minor changes to the code.
New Blocks

PR #928, opened 2023-08-23
-SIPC-/time: some bug fixes and new blocks

SharkPool-SP:
Moved From: https://github.com/TurboWarp/extensions/pull/752#issuecomment-1689271015
(OG .js base file was outdated)

Adds several new calculation blocks for time!
Note: Fixed Several Bugs from Before AND added new blocks!!...
Screenshot 2023-09-16 212935


extensions/-SIPC-/time.js
```diff
@@ -70,20 +75,124 @@
...
+                type: Scratch.ArgumentType.STRING,
+                menu: "Time",
+                defaultValue: "day",
+              },
+            },
+          },
+          {
+            opcode: "calculatetimedurationfromtime",
+            blockType: Scratch.BlockType.REPORTER,
+            text: "time duration from [START_HOUR]:[START_MINUTE] to current time in [TIME_MENU]",
```
GarboMuffin, line 101: you dont need two inputs for the time here

GarboMuffin: you should also rename this and the above to be more like the below

"difference between 12:30 and current time in days"

extensions/-SIPC-/time.js
```diff
@@ -138,6 +247,94 @@
...
+      const days = Math.floor(hours / 24);
+
+      const timeString = `${hours % 24}:${minutes % 60}:${seconds % 60}`;
+      return timeString;
+    }
+
+    daysinmonth(args) {
+      const month = args.MONTH;
+      const year = args.YEAR ? args.YEAR : null;
+      const date = new Date(year, this._getMonthIndex(month) + 1, 0);
```
GarboMuffin, line 317: add better error handling if the month isn't recognized. better to return something that is obviously wrong than something that is subtly incorrect

extensions/-SIPC-/time.js
```diff
@@ -138,6 +247,94 @@
...
+        case "month":
+          return Math.floor(timeDiff / (1000 * 60 * 60 * 24 * 30));
+        case "day":
+          return Math.floor(timeDiff / (1000 * 60 * 60 * 24));
+        case "hour":
+          return Math.floor(timeDiff / (1000 * 60 * 60));
+        case "minute":
+          return Math.floor(timeDiff / (1000 * 60));
+        case "second":
+          return Math.floor(timeDiff / 1000);
```
GarboMuffin, line 266: must always return *some* value

SharkPool-SP:
Screenshot 2023-09-06 222521
Added 2 new Blocks!!!

GarboMuffin:
you're probably gonna be pretty unhappy:

 - removed countdown blocks: they are redundant, one can just invert the difference block that already exists
 - removed blocks that have unintuitive dependence on the unix epoch: the expression "January 1, 1970 + January 27, 2024 = January 27, 2024" just doesn't really make sense. If someone wants to unix them as unix timestamps, the blocks are already there for them.
 - Similarly "convert January 27, 2024 to years = 54.11020848940893" just doesn't seem very useful, and again its already easy for someone to compute with the existing blocks

SharkPool-SP:
> you're probably gonna be pretty unhappy:
> 
> 
> 
>  - removed countdown blocks: they are redundant, one can just invert the difference block that already exists
> 
>  - removed blocks that have unintuitive dependence on the unix epoch: the statement that "January 1, 1970 + January 27, 2024 = January 27, 2024" just doesn't really make sense. If someone wants to unix them as unix timestamps, the tools are already there for them to do that themselves.
> 
>  - Similarly the statement "convert January 27, 2024 to years = 54.11020848940893" just doesn't seem very useful, and again its already easy for someone to compute with the existing blocks

Fine by me

PR #932, opened 2023-08-23
rewrite More comparisons

NOname-awa:
rewrite extension

### DIFF
--- a/extensions/NOname-awa/More comparisons
+++ b/extensions/NOname-awa/More comparisons
@@ -85,7 +85,7 @@
                 color1: '#00a889',
                 color2: '#00977b',
                 color3: '#00866e',
-                id: 'moreComparisons',
+                id: 'NOnameMoreComparisons',
                 name: formatMessage({
                     id: 'EXTENSION_EXT_NAME',
                     default: 'More comparisons'
@@ -659,6 +659,7 @@
                 case 'compositeNumber': return isComposite(Cast.toNumber(INPUT)) && isNumRegex.test(INPUT);
                 case 'text': return !isNumRegex.test(INPUT) && INPUT.length >= 1;
                 case 'boolean': return INPUT === 'true' || INPUT === 'false';
+                default: return false;
             }
         }
         _compare(arg1, arg2, symbol) {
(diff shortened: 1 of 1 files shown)


Skyhigh173:
Your extension id is just 'EXTENSION', you need to change it to something like 'NoNameMoreComparisons'

extensions/NOname-awa/More comparisons
```diff
@@ -0,0 +1,750 @@
...
+                            BOOLEAN: {
+                                type: ArgumentType.BOOLEAN,
+                                defaultValue: ''
+                            }
+                        }
+                    }
+                ],
+                menus: {
+                    BOOLEAN_MENU: {
+                        acceptReporters: false,
```
LilyMakesThings, line 422: Why can't these (and the following menus) be `acceptReporters: true`?

extensions/NOname-awa/More comparisons
```diff
@@ -0,0 +1,750 @@
...
+                                }),
+                                value: 'boolean'
+                            }
+                        ]
+                    }
+                }
+            };
+        }
+        booleanMenu({ BOOLEAN }) {
+            if (BOOLEAN === 'random') return Math.random() < 0.5 ? true : false
```
LilyMakesThings, line 623: You don't have to do this, but internally Scratch's "random" value is `_random_`, consider changing to that for consistency

extensions/NOname-awa/More comparisons
```diff
@@ -0,0 +1,750 @@
...
+                return true;
+            }
+            const isComposite = (number) => {
+                if (number <= 1) return false;
+                for (let i = 2; i <= Math.sqrt(number); i++) {
+                    if (number % i === 0) return true;
+                }
+                return false;
+            }
+            switch (MODE) {
```
LilyMakesThings, line 650: And you're positive this switch will always return _something_, even without a default?

extensions/NOname-awa/More comparisons
yuri-kiss, line 1: Is this supposed to be in a new file, also please add a `.js` extension if so.

yuri-kiss: also please add the default header that all the other extensions have

PR #943, opened 2023-08-24
iframe

GarboMuffin:
only works in staging right now


LilyMakesThings (commented):
Try running this script; the X and Y values don't appear to line up with how someone would expect them to work.
image

GarboMuffin:
apparently i did just forget to test if y position worked

GarboMuffin:
i dont remember if we get any mouse events at all while an iframe is focused, will have to see

Ianyourgod:
the set x position and y position work kinda weird. it will only reset it to the new value if that value is not within the iframe at all. A better way to do this (imo) would be to just always set the middle to that position. I think this is what lily was saying. It seems that the mouse x and y do not  change while it is over the iframe. You could add an event listener to the iframe that while set those things.

GarboMuffin:
Unfortunately browsers just dont give us mouse events while you're hovering over the iframe. you may notice another Scratch mod's version of this has the same problem.

Ianyourgod:
could yuo use messages to send back the mouse pos to the main page?

GarboMuffin:
How do you propose I run custom scripts on `https://example.com`?

Ianyourgod:
oh.

GarboMuffin:
There is very good reason for browsers to do that. Consider if I found a way to embed Wordle. By looking at your mouse movements I could figure out which words you're typing with the on-screen keyboard. That's relatively harmless, but now consider that it's not Wordle, but rather your bank. That's the problem

GarboMuffin:
might be wrong about this actually, wait

GarboMuffin:
Aside from the mouse events not firing when the frame is being hovered issue -- what other blocks should be added?

NexusKitten:
"open iframe contents in a new tab" could be useful

PR #953, opened 2023-08-26
Add TOTP + AES extension (Re-opened with cryptography branch)

LIZARD-OFFICIAL:
Not ``master`` branch anymore


CubesterYT:
@LIZARD-OFFICIAL the two extensions you want to merge still need to be in two different PRs.

LilyMakesThings:
> @LIZARD-OFFICIAL the two extensions you want to merge still need to be in two different PRs.

They don't need to be if they've already been made in 1 PR, it just means that if one has issues and the other doesn't, they will both not get merged immediately.

GarboMuffin:
one extension per pull request, please

DNin01 (changes requested):
I wouldn't really expect an authenticator program to run on TurboWarp because there are already desktop/mobile apps that get the job done (an option that the user doesn't have to program themselves). A TOTP extension sounds kind of like an extension you'd use only once and never again, which isn't accepted.
https://github.com/TurboWarp/extensions/blob/f8182b10b98d5fbb6f57d97c4dbb3c76db12f46a/CONTRIBUTING.md?plain=1#L21
AES encryption sounds fine to me, though.

PR #962, opened 2023-08-27
New image for browser fullscreen

veggiecan0419:
This is the new image:
browserfullscreen
It is more like the style of the other ones on the gallery.


veggiecan0419:
@GarboMuffin, what do you think?

GarboMuffin:
please provide an SVG instead of PNG if possible

veggiecan0419:
> please provide an SVG instead of PNG if possible

@GarboMuffin We don't have an SVG. sorry

GarboMuffin:
ok, that's fine

PR #969, opened 2023-08-27
Hacked Blocks Collection tweaks

### LINKED ISSUE #960: Suggestions for Hacked Blocks
> @LilyMakesThings A few suggestions on how you could improve this:
> 
> - The hacked sensing blocks like `key ( ) pressed` and `touching ( )` can be easily recreated by dragging a join block into their normal variants. I don't think they need to be included in the extension.
> image
> 
> - The normal `when this sprite touches ( )` menu only lists mouse pointer and edge, but the block also works with sprites if you put a blue `sensing_touchingobjectmenu` input inside it, as you can see in [this project](https://scratch.mit.edu/projects/401109316/editor).
> image
> 
> - The `while` loop is already present in TurboWarp's Control category, being the only block changed in the entire vanilla block palette since the `for` loop's removal. It's odd that the exact same block is listed twice when using this extension.
> 
> I would suggest to only include blocks that are impossible to get in TurboWarp without manually editing project.json. You've done a good job at bringing some of these hidden Scratch gems to a wider audience.

_Originally posted by @pumpkinhasapatch in https://github.com/TurboWarp/extensions/issues/845#issuecomment-1694514985_

pumpkinhasapatch:
Resolves #960 
This makes some changes to LilyMakesThings' [Hacked Blocks Collection](https://github.com/TurboWarp/extensions/pull/845) extension.

- Rename the extension from Hacked Block Collection to Hidden Blocks (suggested by [NexusKitten](https://github.com/TurboWarp/extensions/pull/845#issuecomment-1668582721))
- Add Documentation button linking to https://en.scratch-wiki.info/wiki/Hidden_Blocks that explains each block in more detail
- Change the dropdown in `when this sprite touches ( )` block to a sensing menu to also list other sprites, since the block supports it
- Remove sensing reporters with string inputs (these can be worked around by dragging a join block into their normal variants)
- Move the Pen hex reporter below control blocks, and add Music note and micro:bit dot matrix "pixel art" reporters.

It also adds a screenshot image into the extension gallery showing some of the blocks.
image

### [Try it out here!](https://turbowarp.org/editor?extension=https://pumpkinhasapatch.github.io/extensions/extensions/Lily/HackedBlocks.js)


extensions/Lily/HackedBlocks.js
```diff
@@ -1,7 +1,8 @@
-// Name: Hacked Block Collection
+// Name: Hidden Block Collection
 // ID: lmsHackedBlocks
-// Description: Various modified vanilla blocks.
+// Description: Various vanilla blocks not normally visible in the block palette.
```
LilyMakesThings, line 3: Change this to "not visible in the block palette".

extensions/Lily/HackedBlocks.js
```diff
@@ -26,38 +29,37 @@
...
           {
             blockType: Scratch.BlockType.XML,
-            xml: '<block type="sensing_touchingcolor"><value name="COLOR"><shadow type="text"><field name="TEXT">#ffffff</field></shadow></value></block>',
+            xml: '<block type="control_clear_counter"/>',
           },
+          "---",
           {
-            blockType: Scratch.BlockType.XML,
-            xml: '<block type="sensing_coloristouchingcolor"><value name="COLOR"><shadow type="text"><field name="TEXT">#ffffff</field></shadow></value><value name="COLOR2"><shadow type="text"><field name="TEXT">#ffffff</field></shadow></value></block>',
+            blockType: Scratch.BlockType.LABEL,
```
LilyMakesThings, line 47: Remove this label, it's redundant by the way the extension was reorganised

extensions/Lily/HackedBlocks.js
```diff
@@ -26,38 +29,37 @@
             xml: '<block id="while" type="control_while"/>',
```
LilyMakesThings, line 29: I think you're right in that we should remove "while", I left it in by mistake while testing the extension. It doesn't need to be here.

LilyMakesThings (commented):
This is brilliant, I left some comments

extensions/Lily/HackedBlocks.js
```diff
@@ -1,7 +1,8 @@
-// Name: Hacked Block Collection
+// Name: Hidden Block Collection
 // ID: lmsHackedBlocks
-// Description: Various modified vanilla blocks.
+// Description: Various vanilla blocks not visible in the block palette.
 // By: LilyMakesThings <https://scratch.mit.edu/users/LilyMakesThings/>
+// Contributors: pumpkinhasapatch
```
GarboMuffin, line 5: Do `// By: yourusername` so you get put in the list

pumpkinhasapatch: Is it possible to list more than one name? I can't see any extensions that show multiple contributors over at https://extensions.turbowarp.org/

LilyMakesThings: It is! No one's really done it before

extensions/Lily/HackedBlocks.js
pumpkinhasapatch: Honestly I think we should remove `while` from the regular Control category and keep it in this extension. I've never used it and it is the only block that has been changed in the entire default block palette now that `for` is gone.

LilyMakesThings: People were putting in issues when the "for" loop was removed, claiming it was a bug (and that was necessary to fix some strange behaviours).

Removing "while" from the main palette, while certainly the desirable option, is not a good idea

extensions/Lily/HackedBlocks.js
pumpkinhasapatch: Well we should probably do the same for [CST1229 in More Events](https://github.com/TurboWarp/extensions/blob/f8182b10b98d5fbb6f57d97c4dbb3c76db12f46a/extensions/Lily/MoreEvents.js#L84) and you in [Scratch Lab Animated Text](https://github.com/TurboWarp/extensions/blob/f8182b10b98d5fbb6f57d97c4dbb3c76db12f46a/extensions/lab/text.js#L679).

LilyMakesThings: I don't really about credit for Animated Text, and if I did that it'd look like I made the extension. 

@CST1229?

GarboMuffin:
can you find a way to fit the word "hacked blocks" into the description so it shows up when people search for that?

extensions/Lily/HackedBlocks.js
pumpkinhasapatch: Being one of those people, I was annoyed that there was no way to get the for loop back if I still needed it other than searching for an existing project that has the block or hacking it in myself.

With this, we only need to mention that the while loop has been moved to an extension category and you can easily import it from the extension gallery.

pumpkinhasapatch (commented):
I think the for loop would be less buggy if we changed the default variable from "i" to the "my variable" that already exists in default projects. Can't figure out how to set the value though since there's almost no documentation for [Scratch.BlockType.XML](https://github.com/TurboWarp/scratch-vm/pull/144).

LilyMakesThings:
> I think the for loop would be less buggy if we hanged the default variable from "i" to the "my variable" that already exists in default projects. Can't figure out how to set the value though.

That wouldn't resolve it fully, because you can just deleted "my variable" (like many people do), and the problem starts up again.

extensions/Lily/HackedBlocks.js
CST1229: idk, i don't really care

pumpkinhasapatch:
Thanks 😄

PR #873, opened 2023-08-10
Add resolve block to math.js

TheShovel:
Screenshot_20230810_185303

PR #981, opened 2023-08-29
Add clamp and scale blocks to Math

### LINKED ISSUE #888: Map (scale) numbers
For example, map 0 - 1 to 0 - 360.

DNin01:
Resolves #888

- Clamp restricts the outputted number between a certain range. This block was copied from the unlisted Math and String extension. Its new home in the _listed_ Math extension is much more accessible.
- Map (scale) maps a number to a range and then scales that range to a second range.

### DIFF
--- a/extensions/true-fantom/math.js
+++ b/extensions/true-fantom/math.js
@@ -326,6 +326,53 @@
           },
           "---",
           {
+            opcode: "clamp_block",
+            blockType: Scratch.BlockType.REPORTER,
+            text: "clamp [A] between [B] and [C]",
+            arguments: {
+              A: {
+                type: Scratch.ArgumentType.NUMBER,
+                defaultValue: "\n",
+              },
+              B: {
+                type: Scratch.ArgumentType.NUMBER,
+                defaultValue: "0",
+              },
+              C: {
+                type: Scratch.ArgumentType.NUMBER,
+                defaultValue: "100",
+              },
+            },
+          },
+          {
+            opcode: "scale_block",
+            blockType: Scratch.BlockType.REPORTER,
+            text: "map [A] from range [m1] - [M1] to range [m2] - [M2]",
+            arguments: {
+              A: {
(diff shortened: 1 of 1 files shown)


DNin01:
@True-Fantom Your extension is being modified.

extensions/true-fantom/math.js
```diff
@@ -496,6 +543,17 @@
     exactly_cont_block({ A, B }) {
       return cast.toString(A).includes(cast.toString(B));
     }
+    clamp_block({ A, B, C }) {
+      return Math.min(Math.max(A, B), C);
```
NexusKitten, line 547: Consider making this block work both ways around; so if you put a smaller number in B than C it adjusts the equation.

NexusKitten (approved):
You may need to update this PR if #873 goes through before this.

extensions/true-fantom/math.js
NexusKitten: Thank you!

extensions/true-fantom/math.js
```diff
@@ -325,6 +325,53 @@
...
           },
           "---",
+          {
+            opcode: "clamp_block",
+            blockType: Scratch.BlockType.REPORTER,
+            text: "clamp [A] between [B] and [C]",
+            arguments: {
+              A: {
+                type: Scratch.ArgumentType.NUMBER,
+                defaultValue: "\n",
```
GarboMuffin, line 335: you can just put empty strings

extensions/true-fantom/math.js
```diff
@@ -496,6 +543,17 @@
     exactly_cont_block({ A, B }) {
       return cast.toString(A).includes(cast.toString(B));
     }
+    clamp_block({ A, B, C }) {
+      return B < C ? Math.min(Math.max(A, B), C) : Math.min(Math.max(A, C), B);
```
GarboMuffin, line 547: can you make sure to Cast.toNumber() everything here too?

PR #984, opened 2023-08-30
[JSON] Reword "replace item" block to be more consistent

fernyrepos:
Just changed "replace item to" to "replace item with"


fernyrepos:
Screen Shot 2023-08-30 at 1 59 07 AM
this is the proposed current inconsistency, I think it's a good change

PR #988, opened 2023-08-30
[Box2D] Clarity + Consistency improvements

fernyrepos:
Screen Shot 2023-08-30 at 3 37 42 AM

This is mainly for clarity and consistency within itself and other extensions. This extentions specifically gets a lot of questions i'll try to hopefully alleviate with better worded blocks.
- change "setup stage <>" to "set stage boundaries to <>" makes it more clear that this is just setting bounds
- add "to" back to the older physics properties properties (set bounce <>) is now (set bounce to <>)
- the confusing "touching <>" block is now a much more clear "list sprites touching <>"
- added "to" to the set scroll block for internal consistency and global consistency

### DIFF
--- a/extensions/box2d.js
+++ b/extensions/box2d.js
@@ -12522,7 +12522,7 @@
             blockType: BlockType.COMMAND,
             text: formatMessage({
               id: "griffpatch.setStage",
-              default: "setup stage [stageType]",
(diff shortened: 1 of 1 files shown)


DNin01:
The documentation should be updated along eventually.

PR #999, opened 2023-09-02
Add Encode and Decode URIComponent block to HTTP

hdboye:
Title says it all. I'm not 100% sure if this block already exists.
image
image

### DIFF
--- a/extensions/godslayerakp/http.js
+++ b/extensions/godslayerakp/http.js
@@ -432,6 +432,17 @@
             },
             text: "encode URI component [urii]",
           },
+          {
+            opcode: "decodeURIcomp",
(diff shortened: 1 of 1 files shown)


GarboMuffin:
If we have encode that surely we should also have decode

hdboye:
> If we have encode that surely we should also have decode

Oh yeah. Give me a second

hdboye:
Done

SurvExe1Pc:
these blocks do exist fyi, there in encoding

hdboye:
> these blocks do exist fyi, there in encoding

Oh wow, I didn't know. PR closed

PR #1004, opened 2023-09-03
Geolocation Extension

familycomicsstudios:
well yeah

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ node_modules
 
 # Final built website
 build
+
+.gitpod.yml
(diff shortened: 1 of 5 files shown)


LilyMakesThings:
Please read the contributing guidelines before you submit an extension.

hdboye:
https://github.com/TurboWarp/extensions/blob/master/CONTRIBUTING.md
Please read and modify

samuellouf:
I had the same idea a while ago, I think you need to return only the city/country but not the exact location.

CubesterYT:
No, returning the exact coordinates is fine as long as it prompts users that location will be used. It was said in your PR as well.

PR #638, opened 2023-06-25
cursor: add more blocks

LilyMakesThings:
image

PR #1017, opened 2023-09-07
Add Mouse Utilities Extension

NexusKitten:
**This extension is filled with reporters for gauging mouse position and state.** The block pallet:

image

These blocks are (for the most part) based around the MouseEvent Web API, as well as several highly suggested blocks.

## Footnotes:

Testing on mouse/trackpad would be appreciated!
I know this'd fit better in Sensing+, however I'm waiting until it's clean-up to add any blocks to there. If/when that happens, we can move certain blocks from here to there.
Resolves #184
Partially resolves #638

# **_Any feedback/upvotes (👍) are apprecieted!_**

### DIFF
--- a/extensions/NexusKitten/mouse.js
+++ b/extensions/NexusKitten/mouse.js
@@ -3,11 +3,11 @@
 // Description: Several blocks for sensing mouse position and state.
 // By: NamelessCat <https://scratch.mit.edu/users/NamelessCat/>
 
-(function(Scratch) {
-  'use strict';
+(function (Scratch) {
+  "use strict";
 
   if (!Scratch.extensions.unsandboxed) {
-    throw new Error('Mouse Utilities must run unsandboxed');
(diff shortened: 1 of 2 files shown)


Martinelplayz:
do you think if the project were to be packaged there would be a way to force the mouse x and y?

NexusKitten:
> do you think if the project were to be packaged there would be a way to force the mouse x and y? for example(https://www.youtube.com/watch?v=VwZEuRJtI7g)

Likely not, even if the project is packaged it still relies on HTML APIs, which don't allow manipulation of the mouse.
Also, I think it would pose a security risk lol

SurvExe1Pc:
> do you think if the project were to be packaged there would be a way to force the mouse x and y? for example(https://www.youtube.com/watch?v=VwZEuRJtI7g)

you can change the x and y the mousex and mousey reporters return and even the mouse position display addon with some simple javascript.

PR #1025, opened 2023-09-09
Add Frames Extension

True-Fantom:
frames

Скриншот 09-09-2023 210449

Updated version of DeltaTime Extension (I decided to change the name due to the increased expansion possibilities).

Read the background here: https://github.com/TurboWarp/extensions/pull/622

PS: I placed this extension after Runtime Options since they are very related.

PPS: I also cut out "framesim" or "calc" blocks. This is due to the fact that now we have custom reporters, and these functions would simply clog up the palette.


Martinelplayz:
instead of every frame it should be every () frames for more "accessibility" like more events did.

True-Fantom:
@Martinelplayz I tried to implement this block, but there is one problem. "Every () frames" in the More Events extension works very crookedly. A common frame counter "runTimer" is used to count frames. This means that after a certain time it will reach the rounding mark of the number in JS, and all blocks will stop working properly. Also, all such blocks will be synchronized, regardless of when they were delivered, which is probably not bad, but clearly does not meet the user's expectations. So I decided that I would postpone this block for now. In any case, it is easily repeated manually, and my solution should not bring any inconvenience.

XeroName:
I like this extension. Those blocks would be very useful and easy to use with various type game development.

NOname-awa:
this is a good extension

LilyMakesThings:
These changes should really be added to DeltaTime as opposed to making a brand new file.

True-Fantom:
@LilyMakesThings Yes, I understand your point of view, but still there are 2 factors because of which I personally tend to a separate file.

The main factor is too much change in the extension in terms of its perception by users. The name, colors, a bunch of blocks, the main theme, and so on were changed. I think that many users may be disoriented due to the appearance of this extension instead of DeltaTime. It is also possible that some people, in principle, prefer the old version, with a narrower, but without unnecessary blocks for them functionality.

The second factor is backward compatibility. The fps block in my version is backwards incompatible, which also inclines to the choice of a separate file. Replacing 50% of the original blocks in an extension update doesn't sound very good. Plus, these blocks will look the same, which will probably create a lot of unnecessary problems for users.

LilyMakesThings:
> The main factor is too much change in the extension in terms of its perception by users. The name, colors, a bunch of blocks, the main theme, and so on were changed. I think that many users may be disoriented due to the appearance of this extension instead of DeltaTime. It is also possible that some people, in principle, prefer the old version, with a narrower, but without unnecessary blocks for them functionality.

So don't change the name, or the colors. It's an unnecessary aesthetic change.

Alternatively, don't underestimate the intelligence of TW users: Animated Text has seen some drastic updates since it was released and no one has complained or been confused. People are used to it.

> The second factor is backward compatibility. The fps block in my version is backwards incompatible, which also inclines to the choice of a separate file. Replacing 50% of the original blocks in an extension update doesn't sound very good. Plus, these blocks will look the same, which will probably create a lot of unnecessary problems for users.

What about it is incompatible?

NexusKitten:
> I don't see why this can't just be added to DeltaTime.

I think removing the two blocks with replicated functionality in deltatime would make since, however the other other blocks are really more related to _frames_ (the category of this extension) than deltatime.

NexusKitten (approved):

Eliaspierce:
yeah I should not be merging things this late

GarboMuffin:
closing at request of author

PR #1026, opened 2023-09-10
add scoped variable extension

SimonShiki:
# Scope Variables
This extension adds the ability to create, read and modify variables within a limited scope.

category screenshot

Considering that this extension may involve some concepts that are not easy to understand, we need to introduce a few concepts in advance.

### Thread
Each top-level script creates a thread when it is triggered, and the thread is destroyed after the script completes execution. It should be noted that calling custom block does not create a new thread, but runs in the original thread.

### Scope
Scope (called StackFrame in Scratch code) is a small area that is created before executing and destoryed them after block is executed. Each thread stores all the current scopes of the script. Scope is **not only created in C-shape block, but in all blocks.**   
Executing a custom block will also create a new scope.

## How this extension works
When getting variables, the block will first look up for the target variable from the scope one level above the current block's scope up. If the target variable still does not exist in the top-level scope, the extension will look for it in the thread. If it still doesnt't existed , the block returns a empty value.
The look up logic of  setting variables is same as getting. but If it doesn't existed, The block creates a new variable in the outer scope and set it.
The outer scope is only searched once when
```scratch
create scoped [variable] with [value] :: #9999FF
``` 
If it does not exist, The block creates a new variable in the outer scope and set it.


CubesterYT:
I just want to ask for curiosity sake, but what is this extension for and how does it work?

CST1229:
> I just want to ask for curiosity sake, but what is this extension for and how does it work?

I assume it's basically like Temporary Variables' temporary variables, but they are also scoped to C blocks and custom blocks (instead of just threads).

extensions/SimonShiki/scopeVar.js
```diff
@@ -0,0 +1,152 @@
...
+        `);
+  }
+  const { BlockType, ArgumentType, Cast } = Scratch;
+
+  class ScopeVar {
+    getInfo() {
+      return {
+        id: "scopeVar",
+        name: "Scope Variables",
+        collaborator: "Skyhigh173, SimonShiki",
```
NexusKitten, line 21: Not exactly sure where you got "collaborator" and "description" for, but those are already handled in the metadata comments above.

extensions/SimonShiki/scopeVar.js
```diff
@@ -0,0 +1,152 @@
...
+    throw new Error(`
+            "Scope Variables" must be loaded in a unsandboxed environment.
+        `);
+  }
+  const { BlockType, ArgumentType, Cast } = Scratch;
+
+  class ScopeVar {
+    getInfo() {
+      return {
+        id: "scopeVar",
```
NexusKitten, line 19: Typically, extension IDs should contain some way to identify the creator, like your name or initials. "simonScopeVar" would be an example.

NexusKitten (changes requested):
Also maybe consider adding some documentation; it's not very obvious what the extension does from looking at it. Other than that and the other two things I pointed out, this extension should be good to go👍

extensions/SimonShiki/scopeVar.js
SimonShiki: This field will be handled by other scratch mods, so I kept it. If this affects future compatibility on TurboWarp I will remove it.

SimonShiki:
> Also maybe consider adding some documentation; it's not very obvious what the extension does from looking at it. Other than that and the other two things I pointed out, this extension should be good to go👍

documentation added

Martinelplayz:
Change by block plz

SimonShiki:
> Change by block plz

added

Skyhigh173:
suggestion: add a “scope (variable) exists?”

LilyMakesThings:
This appears to work only with the compiler disabled.

Because TurboWarp's main feature is the compiler, we don't usually accept extensions that require it to be off.

SimonShiki:
> This appears to work only with the compiler disabled.
> 
> Because TurboWarp's main feature is the compiler, we don't usually accept extensions that require it to be off.

Turbowarp does not provide an official solution for obtaining a compiler instance or registering an extension block generator, It is ugly and unsafe to modify to the virtual machine instance directly.

LilyMakesThings:
> > This appears to work only with the compiler disabled.
> > Because TurboWarp's main feature is the compiler, we don't usually accept extensions that require it to be off.
> 
> Turbowarp does not provide an official solution for obtaining a compiler instance or registering an extension block generator, It is ugly and unsafe to modify to the virtual machine instance directly.

Which is why we just don't make extensions that require thread data or stuff the compiler can't handle like this.

SimonShiki:
> > Turbowarp does not provide an official solution for obtaining a compiler instance or registering an extension block generator, It is ugly and unsafe to modify to the virtual machine instance directly.
> 
> Which is why we just don't make extensions that require thread data or stuff the compiler can't handle like this.

Are there currently examples of extensions that patch the compiler instance to work properly and accepted by this repository?

LilyMakesThings:
> > > Turbowarp does not provide an official solution for obtaining a compiler instance or registering an extension block generator, It is ugly and unsafe to modify to the virtual machine instance directly.
> > 
> > 
> > Which is why we just don't make extensions that require thread data or stuff the compiler can't handle like this.
> 
> Are there currently examples of extensions that patch the compiler instance to work properly and accepted by this repository?

No, they're expected to just work with the compiler.

extensions/SimonShiki/scopeVar.js
```diff
@@ -0,0 +1,415 @@
...
+    }
+
+    isAlwaysNumberOrNaN() {
+      return this.type === TYPE_NUMBER || this.type === TYPE_NUMBER_NAN;
+    }
+
+    isNeverNumber() {
+      return false;
+    }
+  }
```
yuri-kiss: Compiler patches are unstable and discouraged

extensions/SimonShiki/scopeVar.js
```diff
@@ -0,0 +1,415 @@
...
+            ).asString()}, ${this.descendInput(
+              node.increment
+            ).asNumberOrNaN()}, thread);\n`;
+            break;
+          default:
+            return js_descendStackedBlock.call(this, node, ...otherParams);
+        }
+      };
+    }
+
```
yuri-kiss, line 493: Compiler patches are unstable and discouraged

yuri-kiss (changes requested):
You are gonna have to find a way to make this work without compiler patching/injecting.

Compiler patches are unstable and discouraged.

extensions/SimonShiki/scopeVar.js
SimonShiki: It is almost impossible to implement the original functionality without modifying the compiler, except by forcing the use of scope blocks to create stackFrames.

yuri-kiss: i'd try your luck with that, patching the compiler is dangerous as we don't have a native way to do this as of now

SimonShiki:
When all variable names in a thread are constants, JavaScript `let` is used to greatly improve performance. (though still rely on compiler patches XD)
also fix a bug that occurs while creating a new scope variable at top-level in interpreter/sequencer mode.

SharkPool-SP:
With the new compiler this sorta breaks and emits errors in the console:
image

Is this PR still gonna be worked on and polished or nah close it?

PR #1035, opened 2023-09-14
Policy on AI (ChatGPT, Bard, etc.)

GarboMuffin:
Example of why you shouldn't use ChatGPT: https://chat.openai.com/share/245724e7-0a3c-4001-9fef-fb72c4a4acc6

Everything except the last code block is hallucinated garbage, and even the code block is still worse than useless as it uses ScratchX so it can never be accepted here.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,6 +31,10 @@ We're all volunteers who all have lives outside of Scratch extensions. Many have
 
 Every extension is also covered under [our bug bounty](https://github.com/TurboWarp/extensions/security/policy), so mindlessly merging things will have a direct impact on my wallet.
 
+## On AI language models
+
+**Generative AI language models like ChatGPT, Bing Chat, and Bard DO NOT know how to write proper extensions for TurboWarp.** Remember that the ChatGPT knowledge cutoff is in 2021 while our extension system did not exist until late 2022, thus it *literally can't know*. Pull requests submitting extensions that are made by AI (it's really obvious) will be closed as invalid.
+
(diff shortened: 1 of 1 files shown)


horizontal-shading:
I've never used Chatgpt, but I agree

BunkerStriker:
That is why in the prompt you got to include a real TurboWarp extension for the ChatGPT to try to replicate it
https://chat.openai.com/share/2d46a624-5dca-4e66-b159-4bd1d0a92a07

GarboMuffin:
There are several things wrong with the example you have provided.

 - No one submitting an AI extension actually prompts it like that
 - Math operations is a trivial extension. Try something that we would actually accept here, and watch it painlessly fumble over itself
 - It doesn't include the mandatory // headers at the top
 - The divide block returns a string form of "Infinity" for no reason
 - The extension ID is improperly formatted

It's small things like this that make extension review painful at times because it's obvious someone has never even peeked at the documentation or another extension. What's worse is that ChatGPT, although it's awful at making real extensions, is really good at making things that look real, which makes people submit them in good faith and end up wasting a lot of our time fixing a broken heap of garbage.

IonMakesGarbage:
i dont think people should use it for actually making the extensions but having a quick tool that can check basic errors like a missing comma or colon is extremely useful

GarboMuffin:
if you need an AI to help you fix simple syntax errors (which your editor is already going to notify you about), then you probably shouldnt be submitting extensions here

IonMakesGarbage:
ok...

PR #1040, opened 2023-09-17
Adding my "Audio Recording" extension

samuellouf:
Hello!
I made an "Audio Recording" extension.
You can create a recording, get its base64, import it...
Then you can use the "sound" extension to play the recording.
Here are the blocks:
blocks
Enjoy 😄 !


samuellouf:
I got past eslint!

GalacticCoderReturns:
I've been waiting for an extension like this for a while!

Brackets-Coder:
This is _not_ a bad extension, but closing in favor of #1060, which has more features
This resolves one of the tasks in #2271

If you'd like to collaborate with @SharkPool-SP on a single extension that'd be even better.

PR #1042, opened 2023-09-18
Add DogeisCut/BeepBoxPlayer extension

DogeisCut:
Hello, I've created an extension to play Beepbox, Jummbox, Goldbox, and Ultraabox songs directly from the URL, or JSON. These are the blocks within the extension:
msedge_SzYsQOrk0e
msedge_bMMukP2VKb

This is the icon
beepBoxPlayerIcon

You may be concerned about the use of `new Function()` within the third-party code, it uses it to not have to branch on the number of voices while playing the song. The FM code does the same. The use of `new Function()` does not take user input nor has the potential to, and no variables leak into the global scope. I hope in this very specific case that use of `new Function()` is acceptable, especially since it is part of the third-party code and not my extension.


DogeisCut:
How the heck do you rename commits? Just "oops" as a commit name is not acceptable.

PR #1044, opened 2023-09-18
Add DogeisCut/BeepBoxPlayer extension

DogeisCut:
Hello, I've created an extension to play Beepbox, Jummbox, Goldbox, and Ultraabox songs directly from the URL, or JSON. These are the blocks within the extension:
msedge_SzYsQOrk0e
msedge_bMMukP2VKb

This is the icon
beepBoxPlayerIcon

You may be concerned about the use of `new Function()` within the third-party code, it uses it to not have to branch on the number of voices while playing the song. The FM code does the same. The use of `new Function()` does not take user input nor has the potential to, and no variables leak into the global scope. I hope in this very specific case that use of `new Function()` is acceptable, especially since it is part of the third-party code and not my extension.

Also, I had to create a new pull request due to me messing with branches. Old pull request at https://github.com/TurboWarp/extensions/pull/1042


extensions/DogeisCut/BeepBoxPlayer.js
NexusKitten, line 1: [Line 22085](https://github.com/TurboWarp/extensions/blob/cab0c1cb7694a7d1e3b58e8a434ae2d2e81e112d/extensions/DogeisCut/BeepBoxPlayer.js#L22085C96-L22085C96)
Perhaps add a header or something to indicate that emojis don't work, "I don't care" is the simple solution right now but it'll eventually lead to issues and bug reports.

NexusKitten, line 1: [Line 22121](https://github.com/TurboWarp/extensions/blob/cab0c1cb7694a7d1e3b58e8a434ae2d2e81e112d/extensions/DogeisCut/BeepBoxPlayer.js#L22121C1-L22121C1)
Personally, I'd expect this block to both set the song position to the beginning _and_ play the song if it's not already playing.

NexusKitten, line 1: [Line 22135](https://github.com/TurboWarp/extensions/blob/cab0c1cb7694a7d1e3b58e8a434ae2d2e81e112d/extensions/DogeisCut/BeepBoxPlayer.js#L22135)
The change volume block doesn't appear to work with negative values.

NexusKitten, line 1: This is more of a nitpick, however it'd be nice if the example song had a name instead of just "Untitled."

NexusKitten, line 1: [Line 21731](https://github.com/TurboWarp/extensions/blob/cab0c1cb7694a7d1e3b58e8a434ae2d2e81e112d/extensions/DogeisCut/BeepBoxPlayer.js#L21731)
Preferably we'd find a way to only have to list the example song once in the code to save as much storage space as we can.

NexusKitten (changes requested):
I'll leave the uses of fetch() and new Function() to Garbo as I'm not very experienced in that area, I don't imagine it being much of a vulnerability considering neither of them take user input but I'll wait for feedback from him.

Of course, the big problem with this extension is the fact that the entire BeepBox engine is stored inside of it, there's not much that can be done to get around that but see if you can find a way to minify it or something. Other than all that, nice extension, looking forward to playing around with it!

extensions/DogeisCut/BeepBoxPlayer.js
DogeisCut: Wow...how did I completely miss that lol.

DogeisCut: The example song is in the original BeepBox format which doesn't support song names.

DogeisCut: Oh that comment was a leftover from an earlier version of the block code, It should support emojis now.

DogeisCut: Changed!

DogeisCut: Fixed!

DogeisCut:
Alright! Everything should be fixed and ready to go!

NexusKitten (approved):

DogeisCut:
So uh...

DogeisCut:
man does this just happen to every extension

CubesterYT:
@DogeisCut, what happens?

DogeisCut:
Maybe I'm just impatient or stupid but it seems a lot of extension just get ignored after a long time

LilyMakesThings:
They're not ignored, it's just that this is an open source project riding on the motivation of its sole, human-being manager

NexusKitten:
> man does this just happen to every extension

yeah pretty much

NexusKitten:
If you'd like your extension to be merged somewhere, consider https://github.com/PenguinMod/PenguinMod-ExtensionsGallery, we have more staff onboard and we have cookies

DogeisCut:
Unfortunately, last I checked, my extension is compatible witht MIT but can't use the MIT license due to the Third party code. That's okay for the turbowarp gallery but not the peguinmod one

CubesterYT:
@DogeisCut, all we can say is, when it comes to open source development, there never is anything set in stone. When contributing, keep in mind things may happen instantly all the ways upward to taking months, maybe even a year from what I've seen, even in the most maintained and active scenario.

DogeisCut:
So... i noticed that other mods seem to be adding a similar thing... and it kind sucks that my thing is kinda left in the dust depsite being first........

DogeisCut:
Going to update this soon

DogeisCut:
Ah great to see the review got removed

CubesterYT:
Later today, I'll do a review, add translation functions, and some other things, to prepare it for merging.

PR #1050, opened 2023-09-22
Delete extensions/TheShovel directory

TheShovel:
I do not wish my extensions to be available on TurboWarp anymore. Of course, you can just decline this PR and keep my code, but that doesn't mean I won't complain about it publicly.


GarboMuffin:
As you were aware when you submitted the extension, we can't remove them later as doing so would break real projects by real people. See https://github.com/TurboWarp/extensions/pull/1052

GarboMuffin:
You are, of course, still welcome to participate on the GitHub if you choose to, just as you did before the other place existed

PR #1060, opened 2023-09-26
Add Microphone Recording Extension

SharkPool-SP:
**Perfect for anything... microphone.. related I guess....**

image

Enjoy!!!


snowboyz0825:
oh boy I'm definitely using and archiving this

snowboyz0825:
aww where's the "save recording as x" block that was the one I was most excited for

SharkPool-SP:
> ~aww where's the "save recording as x" block that was the one I was most excited for~ nevermind I loaded the wrong file

lol

snowboyz0825:
@SharkPool-SP when you load the file in turbowarp it turns these menus into random character 
Screenshot 2023-09-26 191121

SharkPool-SP:
> @SharkPool-SP when you load the project file in turbowarp it turns these menus into random character strings
> Screenshot 2023-09-26 191121
> (full project for reference just in case)
> block_9_26_2023-7_15_14 PM

Thanks for bringing that up! Should be fixed now in the new file :)

snowboyz0825:
@SharkPool-SP now it says the correct thing but I can't save recordings to sprites at all

SharkPool-SP:
> @SharkPool-SP now it says the correct thing but I can't save recordings to sprites at all

Crap, I forgot to set the id 🤦‍♂️
I’ll fix it now.

CubesterYT:
What happened to now????

SharkPool-SP:
> What happened to now???? 😭

school happened, Ill get on it in a few hours or so

SharkPool-SP:
> @SharkPool-SP now it says the correct thing but I can't save recordings to sprites at all

Now its 100% Fixed :)

SharkPool-SP:
image

extensions/SharkPool/Recording.js
```diff
@@ -0,0 +1,285 @@
...
+              TYPE: { type: Scratch.ArgumentType.STRING, menu: "FILETYPES" }
+            }
+          }
+        ],
+        menus: {
+          TARGETS: { acceptReporters: true, items: "_getTargets" },
+          ACTIVE: ["on", "off"],
+          MODE: { acceptReporters: true, items: ["enabled", "disabled"] },
+          FILETYPES: {
+            acceptReporters: true,
```
NexusKitten, line 126:
```suggestion
            acceptReporters: false,
```
It's probably not a good idea to accept reporters here, since the way you've programmed the function doesn't check whether a file extension actually exists or not. I didn't personally run into any bugs with this, however I imagine this could cause some unintended functionality.

extensions/SharkPool/Recording.js
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    allowAccess() {
+      return window.confirm("Allow access to record Microphone Audio? Be aware of privacy concerns if you Accept.");
+    }
+
+    // Block Funcs
+    recordingSet(args, util) {
+      if (args.MODE === "disabled") this.stopRecording();
+      else {
```
NexusKitten, line 163:
```suggestion
      else if (args.MODE === "enabled") {
```
Since this block accepts reporters, you should be checking if the argument is "enabled", rather than checking for the argument not being "disabled".

NexusKitten (commented):
I'm not terribly experienced with recording/audio APIs so this review isn't incredibly thorough. It'd be appreciated if someone a little more knowledgeable on the subject could look through this extension as well.

SharkPool-SP:
!format

SharkPool-SP:
> I'm not terribly experienced with recording/audio APIs so this review isn't incredibly thorough. It'd be appreciated if someone a little more knowledgeable on the subject could look through this extension as well.

all done

PPPDUD:
@SharkPool-SP I'm a bit occupied right now and might not get back to you soon. I'll try to handle it when I find a moment though, so thanks for bringing this to my attention!

Brackets-Coder:
I'll review when I can, but a concern I have really fast:
I'm assuming it's not possible to send encoded microphone recordings over cloud variables, correct? I know cloud variables don't work when video sensing is active and I just wanna know if it's the same here.

SharkPool-SP:
It is possible. But given the fact that there's cloudlink I don't think it's necessary

SharkPool-SP:
> remind me to review this please, I'm a bit busy with personal things but I'd be happy to assist you when I can

When

extensions/SharkPool/Recording.js
```diff
@@ -0,0 +1,300 @@
...
+      return new Promise((resolve) => {
+        const reader = new FileReader();
+        reader.onloadend = () => {
+          const base64 = reader.result.split(",")[1];
+          resolve(`data:audio/wav;base64,${base64}`);
+        };
+        reader.onerror = () => resolve("");
+        reader.readAsDataURL(blob);
+      });
+    }
```
Brackets-Coder, line 147: concerns about the potential performance of this when used with blocks, do users know it's asynchronous (e.g., block and wait)

Brackets-Coder (commented):
Alright everything looks good syntax wise

I don't have time to test atm so when I can if it works I'll approve

Brackets-Coder (dismissed):
Seems to look good, everything is working on my side
Safari Desktop

extensions/SharkPool/Recording.js
SharkPool-SP: Huh?

It just converts the unviewable recording data into a readable format (data.URI)

Brackets-Coder: I know that, but it's async so lets say you put in a massive recording, it takes some processing time but the next block might expect to be able to use a result immediately

It's just a thought about the 0.01% chance extremes I guess

SharkPool-SP: That's not possible

SharkPool-SP: If it's async the vm will wait until it's done before moving to the next block

PPPDUD:
Okay, I'm going to review this now.

SharkPool-SP:
Yeah I can remove that block cuz the files extension uses a specific api

Brackets-Coder:
> Okay @PPPDUD @Brackets-Coder , I removed the download file block + the data.uri block and add to sprite block is optimized

I'll check whenever I can but today is busy

PPPDUD (changes requested):
Seems mostly good, but I have two major complaints:
* The block for saving the audio as a sound should have an option to overwrite another sound with the same name.
* There should be a way to start a timed recording without blocking the rest of the script.

PPPDUD:
> > Seems mostly good, but I have two major complaints:
> > 
> > * The block for saving the audio as a sound should have an option to overwrite another sound with the same name.
> > * There should be a way to start a timed recording without blocking the rest of the script.
> 
> 1st point, asset manager exists, if you want to get technical on how to handle downloads, use that. 2nd point, thats why theres a separate start/stop recording block. People can develop their own timers on their own, or more practically, waiting until a condition is met before ending the recording

I suppose that those are good points.

Brackets-Coder (changes requested):
- [x] I verified that the extension follows all acceptance criteria and guidelines in CONTRIBUTING.md.
 - [x] I believe the block list is well-designed and intuitive.
 - [ ] I tested all functionality and found no bugs, security flaws, or unintuitive behavior.
 - [x] I reviewed the code and believe it is well-written and easy to maintain or extend in the future, even without help from the original author.
 - [ ] I have no additional concerns.

Here's my takes:
- [x] I was initially concerned about cloud security with this extension but the TW microphone alert warns users about this so their consent makes this acceptable.
- [x] I find it welcoming how saving a recording with the same name as an existing recording doesn't override the original but keeps both. 
- [ ] I don't enjoy how starting a recording, stopping it, and starting a new one does not override the original recording and erases all data from the new recording. Of course, I should've clicked the "clear" recording block, but I'm wondering if this behavior could be perceived as unexpected for some users.
- [ ] The new "recorded audio data" block that was added since my previous test is very slow in my testing: Clicking the reporter to see its value took over half a minute to parse and report back to me, and during the entire process the editor was frozen and unusable. This is unintuitive for users and I think with other extensions I could save it to the stage and get the URI from that faster than I could this block.

SharkPool-SP:
> [ ] The new "recorded audio data" block that was added since my previous test is very slow in my testing: Clicking the reporter to see its value took over half a minute to parse and report back to me, and during the entire process the editor was frozen and unusable. This is unintuitive for users and I think with other extensions I could save it to the stage and get the URI from that faster than I could this block.

This is not true. Any URI reporter blocks (files extension for example lets say), can potentially freeze the page if the value is too large. This is not exactly fixable for this PR without patches on the extension side.

Saving the URI to the stage does exist, asset manager has a block for that, so does this extension. But viewing it will always cause the page to freeze if the value is too large, regardless of extension.

PR #1069, opened 2023-10-01
itchio: updates

softedco:
- replaced indents with labels
- added blocks related to sub products not mentioned in the original documentation for some reason
- some undefined checking refactored


extensions/itchio.js
```diff
@@ -278,20 +318,25 @@
...
-      try {
-        return data.rewards[index][rewards];
-      } catch (e) {
-        return err();
-      }
+      return data.rewards?.[index]?.[rewards] || err();
     }
     rewardsLenght() {
-      return data.rewards ? data.rewards.lenght : 0;
+      return data.rewards?.lenght || 0;
```
GarboMuffin, line 324: realizing now that length here is misspelled

extensions/itchio.js
```diff
@@ -278,20 +318,25 @@
...
     }
     rewardsBool() {
       return data.rewards ? true : false;
     }
+    returnGameSubProducts({ subProducts, index }) {
+      data.sub_products?.[index]?.[subProducts] || err();
+    }
+    subProductsLenght() {
+      return data.sub_products?.lenght || 0;
```
GarboMuffin, line 333: and here

softedco: all fixed except one opcode changing which will break compatibility

PR #1082, opened 2023-10-07
HTML Encoding

ClaytonTDM:
This is an extremely simple extension. I'm suprised it doesn't exist yet tbh
It can be useful with the new iframe extension


GarboMuffin:
Because the PR you're pulling from is owned by an organization, I can't edit it. I need to make a couple changes so I'm replacing this with https://github.com/TurboWarp/extensions/pull/1126

PR #1086, opened 2023-10-08
Lily/MoreEvents: Add before project saves & after project saves

GarboMuffin:


extensions/when-saved.js
```diff
@@ -0,0 +1,125 @@
...
+      if (threads.length === 0) {
+        resolve();
+        return;
+      }
+
+      const startTime = performance.now();
+      const checkThreadStatus = () => {
+        if (
+          performance.now() - startTime > MAX_BEFORE_SAVE_MS ||
+          threads.every((thread) => !vm.runtime.threads.includes(thread))
```
GarboMuffin, line 25: TODO isActiveThread

extensions/when-saved.js
```diff
@@ -0,0 +1,125 @@
...
+            resolve,
+            methodName,
+            args,
+          });
+        });
+      }
+    };
+
+    const streamWrapper = {
+      on: (...args) => whenStreamReady("on", args),
```
GarboMuffin, line 69: TODO these should not return promise

PR #1098, opened 2023-10-13
Update CloudLink Extension to v0.1.0

MikeDev101:
^
This PR synchronizes cloudlink.js with [cloudlink_turbowarp_0.1.0.js](https://github.com/MikeDev101/cloudlink/blob/master/scratch/cloudlink_turbowarp_0.1.0.js) and is intended to fix bugs, improve performance, simplify/unify the extension's codebase across variants and deprecate older blocks.

### DIFF
--- a/extensions/cloudlink.js
+++ b/extensions/cloudlink.js
@@ -1783,33 +1783,35 @@
     // Reporter - Returns a RESTful GET promise.
     // url - String
     fetchURL(args) {
-      return fetch(args.url, {method: "GET"})
-      .then(response => response.text())
-      .catch(error => {
-        console.warn(`[CloudLink] Fetch error: ${error}`);
-      });
+      return Scratch.fetch(args.url, {method: "GET"})
+        .then(response => response.text())
+        .catch(error => {
+          console.warn(`[CloudLink] Fetch error: ${error}`);
+        });
     }
 
     // Reporter - Returns a RESTful request promise.
     // url - String, method - String, data - String, headers - String
     requestURL(args) {
       if (args.method == "GET" || args.method == "HEAD") {
-        return fetch(args.url, {
+        return Scratch.fetch(args.url, {
           method: args.method,
           headers: JSON.parse(args.headers)
-        }).then(response => response.text())
-        .catch(error => {
-          console.warn(`[CloudLink] Request error: ${error}`);
-        });
+        })
+          .then(response => response.text())
+          .catch(error => {
+            console.warn(`[CloudLink] Request error: ${error}`);
+          });
       } else {
-        return fetch(args.url, {
+        return Scratch.fetch(args.url, {
           method: args.method,
           headers: JSON.parse(args.headers),
           body: JSON.parse(args.data)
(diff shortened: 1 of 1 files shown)


MikeDev101:
???

SurvExe1Pc:
??? lol, glad to see bug fixes!
________________________________

extensions/cloudlink.js
```diff
@@ -1,1838 +1,2406 @@
...
+          "---",
+
           {
             opcode: "onListener",
-            blockType: "hat",
-            text: "When I receive new packet with listener [ID]",
-            blockAllThreads: "true",
+            blockType: Scratch.BlockType.HAT,
+            text: "When I receive new message with listener [ID]",
+            isEdgeActivated: true, // Set to false when runtime.startHats is fixed
```
GarboMuffin, line 1149: what issue?

MikeDev101: I kept receiving errors relating to a runtime/VM issue. Something similar to a .value error, as well as a "not a function" error outside the scope of my extension code.

MikeDev101:
> This one commit is something you will want to bring upstream: [7d533b6](https://github.com/TurboWarp/extensions/commit/7d533b6a0b847930dfcaf6809b928e00aea95ef5)

Updated. Thank you for that.

> I assume this is supposed to be backwards-compatible?

Yes, but mainly for old server support. Should be compatible with older extension versions S4.1 to B3.0. _I also wanted to rewrite it because the code formatting wasn't sane at all..._

extensions/cloudlink.js
GarboMuffin: That error would imply you passed in an object like `{something: undefined}` as the second argument, which isn't valid, the values should just be strings. If you don't want to filter by a field then just don't list it there.

MikeDev101:
I've decided I'll just stick to using edge-triggered events. Should be ready to merge.

extensions/cloudlink.js
```diff
@@ -1,1838 +1,2349 @@
...
+      runtime.startHats('cloudlink_onClose');
+      // Return promise (during setup)
+      return;
     }
+  }
+
+  // GET the serverList
+  try {
+    Scratch.fetch(
+      "https://mikedev101.github.io/cloudlink/serverlist.json"
```
GarboMuffin, line 795: sorry for delay here, looks fine to me, just so you know, you can bypass the fetching permission prompt if you use `https://raw.githubusercontent.com/MikeDev101/cloudlink/master/serverlist.json` here instead

GarboMuffin: We did that in the old version before this PR

I'd like to see that resolved before I merge it (I'll do it this week if you don't), otherwise everyone using this will now be getting a permission prompt before the project even starts, that's just not ideal

MikeDev101: Fixed in https://github.com/MikeDev101/cloudlink/commit/a5bcfa3cc58b8cd613c3babf9ab0703794a1b89c

PR #1100, opened 2023-10-15
Reset various extensions when runtime disposed

GarboMuffin:
Matches what pen does


CubesterYT:
Oh this will be very useful!

PR #1106, opened 2023-10-22
gamejolt: major update

softedco:
### Changes

- Added docs.
- Replaced some blocks with more efficient ones, replaced blocks marked as deprecated.
- Added batch blocks.
  There isn't any way to independently test them without a game so we'll need people that are devs on GJ,
here are the [api docs](https://gamejolt.com/game-api/doc), they cover what fields and params namespaces have, and the extension got rest of the URL construction covered.
- Added debug mode and debug blocks, debug mode solves error messages being interpreted as a valid response, on by default to prevent significant block behaviour change.
- Unwrapped some resolves to put the error message into `err.last` for the `Last API error` block to work properly, solves people improperly using non-returning blocks and telling me that the extension is not working instead of opening the console to see the errors. Thought of doing this with alerts but that would significantly change most existing projects' behaviour.
- Changed text on blocks and some error messages to be more readable.
- Changed GameJolt to Game Jolt because apparently Game Jolt with a space is their actual name, didn't check if this breaks something because it shouldn't (IDs and ID-like stuff untouched)

### And most importantly
- The `bool` object is now just `trueStr`.

### Wrapping up

Lint will probably complain about `data.stuff = undefined;`.
It is there to make blocks not skip errors.
scratchblocks (2)

Should be it.

### DIFF
--- a/docs/gamejolt.md
+++ b/docs/gamejolt.md
@@ -1,8 +1,6 @@
 # Game Jolt API
----
 This extension allows you to easily implement the Game Jolt API using a public domain library.
 ## Blocks
----
 Blocks that the extension uses to send requests to the Game Jolt API.
 
 ```scratch
@@ -10,7 +8,6 @@ Blocks that the extension uses to send requests to the Game Jolt API.
 ```
 Checks to see if the URL is the Game Jolt website.
 ### Session Blocks
----
 Operating on the game's session.
 
 ```scratch
@@ -52,7 +49,6 @@ Sets the session status to active/idle.
 Checks to see if there is an open session for the user.
 - Can be used to see if a particular user account is active in the game.
 ### User Blocks
----
 Login, logout and fetch users.
 
 ```scratch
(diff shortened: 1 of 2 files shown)


softedco:
scratchblocks (3)

softedco:
- Batch blocks are now safer and easier to use.
- Batch blocks are now documented.
- Documentation now also features the `Handling Common Errors` section.

GarboMuffin:
same as last time I don't really have a way to test this, so will just have to trust you

softedco:
> same as last time I don't really have a way to test this, so will just have to trust you

I tested the blocks with @devid.bitt.five. on the discord server / extdev / search: batch blocks (no thread cuz I forgor)
They seem to be working fine, and most of the other blocks' core functionality went untouched apart from some of them being split into separate blocks for efficient data extraction

softedco:
> same as last time I don't really have a way to test this, so will just have to trust you

I tested the blocks with @devid.bitt.five. on the discord server / extdev / search: batch blocks (no thread cuz I forgor)
They seem to be working fine, and most of the other blocks' core functionality went untouched apart from some of them being split into separate blocks for efficient data extraction

PR #1108, opened 2023-10-23
Add get_lines and get_line to Text.js

b1bradders:
I've added `get_lines` and `get_line` which allows to get a specific line from a bit of text. I found this helpful when I'm using Animated Text's `displayed text` block, and the `add line` block.

# Added Blocks
> [!NOTE]
> The purple blocks used here are from an extension called 'Animated Text'
## `get_line`
get_line

## `get_lines`
get_lines
> [!NOTE]
> For `get_lines`, you can see I put the output all on one line. When running this block, it would be split onto seperate lines.
> Unfortunately, the comment would look ugly if I split the text across lines, so I didn't do that. However, the output would look like this in Scratch:
> ```
> Hello!
> Meow!
> Woof!
> ```

### DIFF
--- a/extensions/text.js
+++ b/extensions/text.js
@@ -130,13 +130,13 @@
             arguments: {
               LINE: {
                 type: Scratch.ArgumentType.NUMBER,
-                  defaultValue: 1,
-                },
-                STRING: {
-                  type: Scratch.ArgumentType.STRING,
-                  defaultValue: "Hello, world!\nGoodbye, world!",
-                },
+                defaultValue: 1,
               },
+              STRING: {
+                type: Scratch.ArgumentType.STRING,
+                defaultValue: "Hello, world!\nGoodbye, world!",
+              },
+            },
           },
           {
             opcode: "get_lines",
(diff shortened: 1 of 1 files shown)


CST1229:
get_line is pretty much just the split block with the newline character (which a block for exists in Utilities iirc).

SharkPool-SP:
> I made, what I think is, a better & simpler image for Text too. Should I put that in a seperate PR or this one?
> text_ext
> (the [.afphoto file](https://github.com/TurboWarp/extensions/files/13070686/text_ext.afphoto.zip) in case anyone wants to mess about with it in Affinity Photo)

the extension thumbnail is fine as is. A simple gradient as the background is pretty bland.

GarboMuffin:
> Why can't it just format the file itself and then make changes to the PR? I think GH Actions can do that.

Usually we have to clone PRs locally and make some other changes anyways, so we usually just stuff it as part of that. It is possible to do in Actions but that can mangle git history pretty bad and you still have to manually enable GitHub Actions yourself

PR #1119, opened 2023-10-30
utilities: fix some blocks breaking if specific costumes exist

CST1229:
Resolves [#Clamp bug](https://discord.com/channels/837024174865776680/1168243452899774494)

For some weird, inexplicable reason that I do not know why it happens, ArgumentType.NUMBER arguments are auto-casted to number, but _only_ if a costume with that name does not exist.
This breaks a few math-related blocks in Utilities under specific circumstances (for example, `clamp 30 between 25 and 100` returns `25` if costumes named 25 and 100 exist).
This PR fixes that, by casting the arguments to numbers in `clamp` and using Scratch.Cast.compare on `isLessOrEqual` and `isMoreOrEqual`. `exponent`'s arguments are also casted now just to be safe.

### DIFF
--- a/extensions/utilities.js
+++ b/extensions/utilities.js
@@ -317,9 +317,9 @@
       MIN = Scratch.Cast.toNumber(MIN);
       MAX = Scratch.Cast.toNumber(MAX);
       if (MIN > MAX) {
-        return Scratch.Cast.toNumber(Math.min(Math.max(INPUT, MAX), MIN));
+        return Math.min(Math.max(INPUT, MAX), MIN);
       } else {
-        return Scratch.Cast.toNumber(Math.min(Math.max(INPUT, MIN), MAX));
+        return Math.min(Math.max(INPUT, MIN), MAX);
       }
     }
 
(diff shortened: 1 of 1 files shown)


extensions/utilities.js
```diff
@@ -311,6 +313,9 @@
     }
 
     clamp({ INPUT, MIN, MAX }) {
+      INPUT = Scratch.Cast.toNumber(INPUT);
+      MIN = Scratch.Cast.toNumber(MIN);
+      MAX = Scratch.Cast.toNumber(MAX);
       if (MIN > MAX) {
         return Scratch.Cast.toNumber(Math.min(Math.max(INPUT, MAX), MIN));
```
GarboMuffin, line 320: you can remove the toNumber cast here and below then

GarboMuffin:
extensions are supposed to cast all of their inputs themselves, for example when someone drops in a variable you don't know what you're going to get. in this case not casting inputs to numbers if it matches the name of a costume fixes compatibility bugs with "set costume to (...)" blocks

PR #1138, opened 2023-11-12
easily struct extension in scratch

qxsck:
easily struct extension in scratch:
it can create and use struct and struct list,but there are still many functions that have not been implemented,such as struct list sorting.
image
image
image

The extension icon I designed:
easily struct


qxsck:
don't merge,this is not completed

qxsck:
new block:
image

qxsck:
stop pr it，need fix

PR #1140, opened 2023-11-12
PenPlus 6.1 Bug Fixes

David-Orangemoon:
Backport the V7 depth buffer
Backport tri drawing from V7
Fix colour bug
Fix dot and line drawing bug.

### DIFF
--- a/extensions/obviousAlexC/penPlus.js
+++ b/extensions/obviousAlexC/penPlus.js
@@ -159,12 +159,10 @@
     });
 
     vm.runtime.on("BEFORE_EXECUTE", () => {
-      let calcSize = (renderer.useHighQualityRender
+      let calcSize = renderer.useHighQualityRender
         ? [canvas.width, canvas.height]
-        : renderer._nativeSize);
-      if (
-        (calcSize[0] != nativeSize[0]) || (calcSize[1] != nativeSize[1])
-      ) {
+        : renderer._nativeSize;
+      if (calcSize[0] != nativeSize[0] || calcSize[1] != nativeSize[1]) {
         nativeSize = renderer.useHighQualityRender
           ? [canvas.width, canvas.height]
           : renderer._nativeSize;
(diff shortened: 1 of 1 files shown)


David-Orangemoon:
Tried to make it work with the draw regions system. Didn't work. only drew 1 triangle lol. The pr is done.

GarboMuffin:
done?

extensions/obviousAlexC/penPlus.js
```diff
@@ -154,18 +158,48 @@
...
         nativeSize = renderer.useHighQualityRender
           ? [canvas.width, canvas.height]
           : renderer._nativeSize;
         updateCanvasSize();
+        resizeCall = false;
+      }
+    });
+
+    //Penguinmod
+    vm.runtime.on("RUNTIME_STEP_START", () => {
```
GarboMuffin, line 186: its on them to implement BEFORE_EXECUTE which would also fix other extensions.

GarboMuffin: the rationale here is 

 - this is the TurboWarp extension gallery
 - we don't need to set a precedent of having every BEFORE_EXECUTE also have an identical RUNTIME_STEP_START
 - them fixing it will fix other extensions automatically
 - it takes them one line to implement

David-Orangemoon:
> done?

Yes

extensions/obviousAlexC/penPlus.js
David-Orangemoon: I'm just trying to have it so I don't have to main multiple branches like I did for the old turboloader builds.

GarboMuffin: hopefully other mods add support for APIs such as BEFORE_EXECUTE so you don't need to maintain separate versions like that

David-Orangemoon: They did

PR #1158, opened 2023-11-19
Reacquire wake lock as needed

### LINKED ISSUE #1120: Listen for wake lock release
Currently, my Wake Lock extension does not know when a wake lock has been released. A wake lock can be released for reasons as simple as navigating away from the browser tab it was set in. Fortunately, you can add [an event listener](https://developer.mozilla.org/en-US/docs/Web/API/Screen_Wake_Lock_API#listening_for_wake_lock_release) for when the wake lock is released.

I'll try to make some changes later.

DNin01:
Resolves #1120

Currently, if you request a wake lock with the Wake Lock extension and then navigate away from the page, the wake lock gets released in the background because wake locks are automatically released when the document gets hidden, plus, the extension isn't aware of when this happens, so you have no way of knowing. 

This change makes the extension automatically request another wake lock when the document regains visibility after having been hidden. It still reports to the project that the wake lock is enabled when this happens, therefore projects will act like nothing happened but the wake lock will still be there as long as you come back. I also considered the possibility of projects enabling or disabling wake lock when the document is hidden and now handle it properly.

I tested these changes in Edge to the best of my ability.

### DIFF
--- a/extensions/DNin/wake-lock.js
+++ b/extensions/DNin/wake-lock.js
@@ -100,6 +100,13 @@
           .then(() => navigator.wakeLock.request("screen"))
           .then((sentinel) => {
             wakeLock = sentinel;
+            wakeLock.addEventListener("release", () => {
+              if (document.visibilityState === "visible") {
+                // If the document is hidden, wake lock should be reacquired when it's visible again.
+                wakeLock = null;
+                latestEnabled = false;
+              }
+            });
           })
           .catch((error) => {
             console.error(error);
--- a/extensions/DNin/wake-lock.js
+++ b/extensions/DNin/wake-lock.js
@@ -19,6 +19,16 @@
     constructor(runtime) {
       this.runtime = runtime;
       this.runtime.on("PROJECT_STOP_ALL", this.stopAll.bind(this));
+
+      document.addEventListener("visibilitychange", () => {
+        // If enabled, reacquire wake lock when document becomes visible again
+        if (wakeLock !== null && document.visibilityState === "visible") {
+          latestEnabled = false;
+          this.setWakeLock({
+            enabled: true,
+          });
+        }
+      });
     }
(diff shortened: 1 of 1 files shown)


extensions/DNin/wake-lock.js
```diff
@@ -77,9 +87,14 @@
         // Not supported in this browser.
         return;
       }
+      const enable = Scratch.Cast.toBoolean(args.enabled);
+      if (enable && document.visibilityState === "hidden") {
+        // Can't request wake lock while document is hidden.
```
GarboMuffin, line 92: what happens if you try anyways?

DNin01: Error:
```
DOMException: Failed to execute 'request' on 'WakeLock': The requesting page is not visible
```

From there, it catches the error and sets `latestEnabled` to false so it can try again. But I kinda feel like we shouldn't have any errors or time wasted trying to acquire a wake lock when we know we can't. What do you think?

DNin01:
I also realized I didn't do anything about what happens when a wake lock is unexpectedly released. At that point, it still thinks that the wake lock is active. It's a little complicated to do while also making sure it gets reacquired if the user leaves and comes back because it leaves three possible states for the wake lock to be in: _off_, _on_, and _will be reacquired when the document is visible again_, but I have an idea.

DNin01:
Alright! It now detects when wake lock is released. The comment I added is a little confusing, basically what it's saying is that we know that if the document is hidden, that's the reason the wake lock was released and to reacquire it when the document is visible again, so if the wake lock is released while the document is _visible_, it must be for some other reason so we don't want to reacquire it automatically and we report the wake lock status as disabled.

GarboMuffin:
sorry for delay, this is a bit hard to test but I'll take your word on it

PR #1159, opened 2023-11-19
Add Advanced Numerical Encoding extension

Stiwen02:
I've made an extension that lets you encode strings to numbers with advanced blocks such as encoding multiple strings into a single number and getting the amount of items encoded in a number.
image
image
I can’t modify the already existing Numerical Encoding extension because it would break projects that use that extension because my extension has a different character list.


InfinityLoopGames:
I like this! There is already a Numerical Encoding extension, but this one is much better! I hope this gets added because this looks excellent!

GarboMuffin:
The numerical encoding concept of having separate "Encode [xyz]" and "get encoded" blocks was silly and shouldn't be emulated without good reason -- do you have a good one?

Stiwen02:
> The numerical encoding concept of having separate "Encode [xyz]" and "get encoded" blocks was silly and shouldn't be emulated without good reason -- do you have a good one?

The "encode [xyz]" command is used to encode a string and add it to the encoded result, which you can get with the "get encoded" reporter, if you're talking about the "encode [xyz]" reporter then that's used if you only want to encode a single string and keep the code small.

Stiwen02:
Hm. The large .JS file should be causing the error, since the new icon has more colors the data URI is way larger.

Stiwen02:
I've successfully changed the icon, should be readable in light mode

PR #1169, opened 2023-11-27
Add my IP Addresses extension

PPPDUD:
The IP Addresses extension contains blocks for handling public IP addresses.


SharkPool-SP:
This block already exists in Cloudlink

wolfieboy09:
I've made an extension that contains that very block

SharkPool-SP:
> I've made an extension that contains that very block

I did too but expands on it :trollface:

CubesterYT:
Nope, it's not possible to access private ips. Plus that ip is useless.

PPPDUD:
Looks like this is already implemented; sorry about that, devs.

PR #1172, opened 2023-11-29
gamejolt.js: Scratch.translate("Batch Blocks") and a bit more

softedco:
- Added missed Scratch.translate() for the "Batch Blocks" label.
- Added translations for known errors.
  - Errors are in english by default for backwards compatibility.
  - The language is changeable through buttons in debug options as this only affects the editor.
  - Also I need to know what language does the project use to hide the second button if the project's language is english
    I tried this:

    ```js
    hideFromPalette: Scratch.Language == "en", // Doesn't work
    ```
    So this needs to be fixed before merging.

### DIFF
--- a/extensions/gamejolt.js
+++ b/extensions/gamejolt.js
@@ -1323,11 +1323,116 @@
    * Used for storing API error messages
    */
   let err = {
+    // Backwards compatibility
     debug: true,
+    projectsLanguage: false,
 
     noLogin: "No user logged in.",
     noData: "Data not found.",
     noIndex: "Data at such index not found.",
+
+    known: [
+      "No user logged in.",
+      "Data not found.",
+      "Data at such index not found.",
+      "Unknown fatal error occured.",
+      "You must enter a signature with your request.",
+      "The signature you entered for the request is invalid.",
+      "The game ID you passed in does not point to a valid game.",
+      "No such user with the credentials passed in could be found.",
+      "This key has restrictions placed on it. Please pass in a restriction user with valid permissions.",
+      "You must enter in a user ID or username.",
+      "No such user could be found.",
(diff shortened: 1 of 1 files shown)


softedco:
I can't seem to find what stores the language of the project, can someone help me? If it's not stored anywhere the extension has access to I'll go with changing what the buttons say so they don't need to be hidden.

extensions/gamejolt.js
```diff
@@ -1323,28 +1323,150 @@
...
+      Scratch.translate("You must enter an value with the request."),
+      Scratch.translate(
+        "Mathematical operations require the pre-existing data stored to also be numeric."
+      ),
+      Scratch.translate("Value must be numeric if operation is mathematical."),
+      Scratch.translate({
+        id: "err.divisionByZero",
+        default:
+          "GAME JOLT STOP: 0x00000019 (0x00000000, 0xC00E0FF0, 0xFFFFEFD4, 0xC0000000) UNIVERSAL_COLLAPSE",
+        description: "Try to keep it as cool.",
```
GarboMuffin, line 1433: you have to be a bit more specific than that, they won't understand that

GarboMuffin:
The major piece of feedback from translators was that a *lot* of the strings in here *need* descriptions. Something like `Add score [value] in table of ID [ID] with text [text] and comment [extraData]` is really hard to decipher and translate properly

GarboMuffin:
> Also I need to know what language does the project use to hide the second button if the project's language is english

New api is being added right now: `if (Scratch.translate.language !== 'en') { }`

GarboMuffin:
lets just say this is superseded by https://github.com/TurboWarp/extensions/pull/1540

PR #1176, opened 2023-12-02
More Control: Add More Control extension

LilyMakesThings:
This was a collaborative effort between myself and SharkPool
image


LilyMakesThings:
A downside to a collaborative effort is errors you don't know how to fix.

LilyMakesThings:
> still no `break` and `continue` 🥲

break and continue is hard 😔

SurvExe1Pc:
> still no `break` and `continue` 🥲

these require "cheeky" methods that Garbo would not approve of.
see my switch case for more info.

SurvExe1Pc:
i will see if "hasOwn" is viable

Skyhigh173:
Is it possible to add a "switch value"? I want to test something

SurvExe1Pc:
> Is it possible to add a "switch value" reporter? I want to test something

yes.

SurvExe1Pc:
done @Skyhigh173

Skyhigh173:
A "minor" bug (a old bug I also found in my attempt of switch case):

image

Switch value in same block will overwrite previous one.

SurvExe1Pc:
> A "minor" bug (a old bug I also found in my attempt of switch case):
> 
> image
> 
> Switch value in same block will overwrite previous one.

tis fine, its not the intended use of the block, just store it in a variable

CubesterYT:
@GaypataponALT, It's been cancelled, sadly.

SurvExe1Pc:
can we mark thoes comments as off topic please.

horizontal-shading:
Cool

LilyMakesThings:
This extension is finally finished as far as I'm concerned.

👍 reactions mean the world.

qwertyy-the-artist:
if you haven't already, you should definitely add a documentation... i don't understand the `default`, `switch`, `case`, and `launch` blocks

LilyMakesThings:
> if you haven't already, you should definitely add a documentation... i don't understand the `default`, `switch`, `case`, and `launch` blocks

documentation will come eventually

extensions/Lily/MoreControl.js
```diff
@@ -344,6 +365,34 @@
...
+        return thread.returnValue ?? "";
+      } else {
+        util.thread.peekStackFrame().waitingReporter = true;
+        util.yield();
+      }
+    }
+
+    inlineReturn(args, util) {
+      util.thread.returnValue = args.VALUE;
+      util.thread.status === 4;
```
CST1229, line 393: i think this should be an assignment

LilyMakesThings:
As great as it was having the inline block use the same return block as custom reporters, it wasn't going to work in the interpreter or the compiler.

SharkPool-SP:
@LilyMakesThings can you fix Mr.LongChin for the "if repeat until" block. Set the branch icon to nothing maybe

LilyMakesThings:
You can now add a variable reporter in place of the I reporter on any of the for loop blocks to set that variable as well.
This save me from having to add a variable menu (and subsequently 4 more blocks).

https://github.com/TurboWarp/extensions/assets/127533508/387b12d7-930d-4489-a204-68875b404180

The docs will be updated accordingly.

extensions/Lily/MoreControl.js
```diff
@@ -0,0 +1,1214 @@
...
+      runtime._restartThread(util.thread);
+    }
+
+    forArg(args, util) {
+      const param = "i";
+      const stackFrames = util.thread.stackFrames;
+      if (typeof stackFrames === "undefined") return 0;
+
+      const params = stackFrames[0].moreControlParams;
+      if (typeof params === "undefined") return 0;
```
AlexSchoolOH, line 745: A problem I encountered while working on a project there should be a condition for json Objects in the JSON array. I don't know exactly how it is fetched from the stack but
there should be a check for the param being an object!
```js
if (typeof params[param] == "object") return JSON.stringify(params[param]);
```

before returning
```js
return params[param] ?? 0;
```

AlexSchoolOH (changes requested):
That is all I can find

extensions/Lily/MoreControl.js
LilyMakesThings: Thank you! this will be fixed shortly

raynepaws (approved):
[ removed - old, i was a rather inexperienced github user at the time of writing ]

Skyhigh173:
I think "for each (I) in array" should be in json extension

LilyMakesThings:
> I think "for each (I) in array" should be in json extension

Ehh, a lot of the setup to get it to work is already done in this extension for the other 3 blocks, and frankly you could argue about that for all of them.

Ashimee:
you made a pr "patch" so this could be native without fancy hacks right?

________________________________

LilyMakesThings:
> you made a pr "patch" so this could be native without fancy hacks right?

I'd rather just keep the 4 blocks within this extension, together

Ashimee:
that was not the question but ok lol
________________________________

Skyhigh173:
perhaps i can make mine, and you can keep yours

AshimeeAlt:
woops sorry bout the formatting 😅

extensions/Lily/MoreControl.js
```diff
@@ -0,0 +1,1294 @@
...
+    runNextCaseWhen(args, util) {
+      const blockId = this._getPeekStack(util.thread);
+      const outerC = this._getOuterCFromOpcode(
+        util.thread,
+        blockId,
+        "lmsSpAsMoreControl_switch"
+      );
+      if (!outerC) return;
+
+      if (args.CASE === outerC.switchValue) {
```
Johang727, line 659: This line (`659`) not allow integers and strings to match due to the **triple**(`===`) equals sign. I feel like a **double**(`==`) would be better.

Example: This does not work with `===`, because picking a random number returns an **integer** ~~(or float, I'm not entirely sure)~~ and the case inputs are **strings**.

This leads to a type mismatch which will always return false.

block_11_29_2024-2_33_29 PM

However, changing this to `==` resolved the problem I was having.

This will always result in the **default** case being run, if added.

Side note: If this is an intended feature, ignore this; I know nothing about JavaScript. :p

Brackets-Coder:
!format

Brackets-Coder:
@CubesterYT can we consolidate this and #1183 into one extension

CubesterYT:
I'm not merging controls+ with Lily's

I plan to rework my extension eventually

PR #1189, opened 2023-12-08
Add mbw/xml

### LINKED ISSUE #1099: XML extension idea
So there is already a JSON extension, but there is an alternative to JSON called XML, and I was wondering if there would be an extension for handling XML. If there was, then that would be cool.

mybearworld:
Resolves #1099

The logo of the XML extension. The symbols "</>" written across a purple backgrounds with a few circles over them.

This extension adds capability of interacting with XML. This includes SVG and XHTML.

The blocks of the XML extension in the block pallete.


Ashimee:
👏

CubesterYT:
@mybearworld, Welcome! I was wondering when you'll finally join us here at TurboWarp. This is a great first extension!

qxsck:
XML extension!good!

mybearworld:
I've just updated the extension to use `Scratch.translate` – am I using it correctly?

GarboMuffin:
at a first glance it seems good

this is much more complete than a lot of the extensions here so I will try to get it in soon -- it got lost in a flood of new pull requests

extensions/mbw/xml.js
```diff
@@ -0,0 +1,598 @@
...
+      return JSON.stringify([...xml.attributes].map((attr) => attr.name));
+    }
+
+    /**
+     * @param {object} args
+     * @param {unknown} args.XML
+     * @param {unknown} args.ATTR
+     */
+    hasAttribute({ XML, ATTR }) {
+      return this.getAttribute({ XML, ATTR }) === "";
```
VoxelDoesCode, line 404: Shouldn't this be the opposite? It's checking to see if an attribute exists inside an existing XML script, yet it returns true if getAttribute is null (meaning it doesn't have the attribute).

mybearworld: Thanks, you're right

PR #1190, opened 2023-12-09
box2d: disable physics and simulation rate blocks

CST1229:
Resolves #1171
Might resolve #925?

Adds 3 new blocks to the Box2D Physics extension:
image
Disables physics for the sprite.

image
Sets (or gets) the simulation rate; for example, you can set it and the project's framerate to 60 to make the physics run at 60FPS.

PR #1197, opened 2023-12-12
set framerate of box2d physics to framerate of project

jwklong:
this would prevent speeding up physics on higher framerates

### DIFF
--- a/extensions/box2d.js
+++ b/extensions/box2d.js
@@ -13219,7 +13219,7 @@
       this._checkMoved();
 
       // world.Step(1 / 30, 10, 10);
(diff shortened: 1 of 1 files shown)


jwklong:
> this isn’t a great idea as it makes the physics choppy. you could instead just implement delta time

it would only make it choppy on lower framerates tho

LilyMakesThings:
https://github.com/TurboWarp/extensions/pull/1190 allows you to modify the simulation rate separately.

jwklong:
> #1190 allows you to modify the simulation rate separately.

oh ok i guess i'll just close this one then

PR #1200, opened 2023-12-14
Create CommentPlus.js

Mastered-YT-Aditya:
Well I've created a extension called Comment Plus. Actually it's a developed version of Comment Blocks which was made by LilyMakesThings.


CST1229:
Why not just add the blocks to the existing Comment Blocks extension?

wolfieboy09:
image

Thats not supposed to happen

Mastered-YT-Aditya:
> Why not just add the blocks to the existing Comment Blocks extension?

Yeah I could but I just created my own.

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,383 @@
...
+            arguments: {
+              COMMENT: {
+                type: Scratch.ArgumentType.STRING,
+                defaultValue: "comment",
+              },
+            },
+          },
+          {
+            opcode: "commentCap",
+            blockType: Scratch.BlockType.CAP,
```
LilyMakesThings, line 33: This block causes a compiler error, remove it

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,383 @@
...
+    commentReporterP7(args) {
+      return args.COLOR;
+    }
+
+    commentBooleanP7(args) {
+      return args.COLOR || false;
+    }
+    
+    commentCPlus(args, util) {
+      return true;
```
LilyMakesThings, line 379: The second branch doesn't start so what's even the point of it to begin with?

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,383 @@
...
+    commentReporterPPPP(args) {
+      return args.INPUT;
+    }
+
+    commentBooleanPPPP(args) {
+      return args.INPUT || false;
+    }
+
+    commentBooleanPlus(args) {
+      return args.INPUT;
```
LilyMakesThings, line 343: `args.INPUT` is not defined so this returns nothing

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,383 @@
...
+    }
+
+    commentReporterPlus(args) {
+      return args.INPUT;
```
LilyMakesThings, line 347: `args.INPUT` is not defined so this returns nothing

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,383 @@
...
+                defaultValue: "comment",
+              },
+              COLOR: {
+                type: Scratch.ArgumentType.COLOR,
+                defaultValue: "#037b03",
+              },
+            },
+          },
+          {
+            opcode: "commentCPlus",
```
LilyMakesThings, line 287: Why are there 2 of these?

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,383 @@
...
+    commentReporterPlus(args) {
+      return args.INPUT;
+    }
+
+    commentCapPlus() {
+      // no-op
+    }
+    
+    commentReporterPlusPlus(args) {
+      return args.INPUT;
```
LilyMakesThings, line 355: This needs to return something even if there's nothing in the conditional

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,383 @@
...
+    commentReporterPPP(args) {
+      return args.INPUT;
+    }
+
+    commentBooleanPPP(args) {
+      return args.INPUT || false;
+    }
+
+    commentReporterPPPP(args) {
+      return args.INPUT;
```
LilyMakesThings, line 335: This needs to return something even if there's nothing in the conditional

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,383 @@
...
+                defaultValue: "comment",
+              },
+              INPUT: {
+                type: Scratch.ArgumentType.STRING,
+                defaultValue: "",
+              },
+            },
+          },
+          {
+            opcode: "commentBoolenPlus",
```
LilyMakesThings, line 152: You misspelled "Boolean" so this doesn't do anything (see later comment for when it does do something)

LilyMakesThings (approved):
This extension is relatively pointless; I don't see the point in over half of these blocks when the couplers extension exists and does a similar job.

Furthermore, as per the [contributing guidelines](https://github.com/TurboWarp/extensions/blob/master/CONTRIBUTING.md#acceptance-criteria), you shouldn't submit extensions that are similar to existing ones, and this is an adaptation of an extension on the gallery.

Consider adding the blocks you think are essential to the existing extension as others have suggested.

extensions/MYTAditya/CommentPlus.js
Mastered-YT-Aditya: Oh it was actually I wanted to write `commentCPlusPlus` at that line

Mastered-YT-Aditya:
> This extension is relatively pointless; I don't see the point in over half of these blocks when the couplers extension exists and does a similar job.
> 
> Furthermore, as per the [contributing guidelines](https://github.com/TurboWarp/extensions/blob/master/CONTRIBUTING.md#acceptance-criteria), you shouldn't submit extensions that are similar to existing ones, and this is an adaptation of an extension on the gallery.
> 
> Consider adding the blocks you think are essential to the existing extension as others have suggested.

All problems has been fixed. 
And I know that there's Couplers Extension do that, also there's already Comment Blocks made by you and people can just build mine blocks using those Extensions. But I just created it for those peoples whom don't want to add Couplers only for Comment Blocks. Maybe Couplers is a great extension almost every people uses it (me too).
And lastly, if you want then I would remove everything I've made.

Mastered-YT-Aditya:
> image
> 
> Thats not supposed to happen

Fixed

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,376 @@
...
+
+    commentBooleanPlus(args) {
+      // no-op
+    }
+
+    commentReporterPlus(args) {
+      // no-op
+    }
+
+    commentCap() {
```
LilyMakesThings, line 339: Remove your unused functions (this is also throwing an error)

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,376 @@
...
+                defaultValue: "comment",
+              },
+              COLOR: {
+                type: Scratch.ArgumentType.COLOR,
+                defaultValue: "#037b03",
+              },
+            },
+          },
+          {
+            opcode: "commentCPlusPlus",
```
LilyMakesThings, line 276: Can you specify what the point of a double-branch comment block is? As stated previously, the second branch doesn't even run because of how you return the function.

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,376 @@
...
+
+    commentReporterPPPP(args) {
+      return args.INPUT || false;
+    }
+
+    commentBooleanPPPP(args) {
+      return args.INPUT || false;
+    }
+
+    commentBooleanPlus(args) {
```
LilyMakesThings, line 331: This still doesn't return anything; reporters cannot return nothing.

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,376 @@
...
+      // no-op
+    }
+
+    commentReporterPlus(args) {
```
LilyMakesThings, line 335: This still doesn't return anything; reporters cannot return nothing.

images/MYTAditya/CommentPlus.svg
LilyMakesThings, line 1: Text elements are forbidden in the eyes of the muffin! In other words, remove the text or convert it.

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,376 @@
...
+    commentCap() {
+      // no-op
+    }
+
+    commentCPlus() {
+      // no-op
+    }
+    
+    commentReporterPlusPlus(args) {
+      return args.INPUT || false;
```
LilyMakesThings, line 348: This block is a reporter with a boolean input. Boolean blocks can already go into reporter holes so this isn't exactly opening up any other opportunities that the commentBoolean couldn't do.

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    commentBooleanPlusPlus(args) {
+      return args.INPUT || false;
```
LilyMakesThings, line 352: this is a boolean block, yet it sometimes returns a string and sometimes returns "false" if there's an absence of a value in the input.

apply the same rules as commentReporterPlusPlus, or `Scratch.Cast.toBoolean` the input.

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,376 @@
...
+
+    commentBooleanP6(args) {
+      return args.COLOR || false;
+    }
+    
+    commentReporterP7(args) {
+      return args.COLOR;
+    }
+
+    commentBooleanP7(args) {
```
LilyMakesThings, line 367: I don't understand where or why you'd need a color input inside of a boolean.

LilyMakesThings (changes requested):
This extension is really hard to review comprehensively.

extensions/MYTAditya/CommentPlus.js
Mastered-YT-Aditya: Yeah I know but just created a separate block.

extensions/MYTAditya/CommentPlus.js
Mastered-YT-Aditya: Yeah, that block is just pointless, but I created that for the blocks put in first branch would work (like previous ones) and next branch wouldn't work. If you can, please write and comment me the code where both branches would work in the same block.

extensions/MYTAditya/CommentPlus.js
Mastered-YT-Aditya: Yeah you're right because reporters could do that. You can say, I just created it for Boolean shaped block. If you say, I would remove Booleans which had colors strings.

extensions/MYTAditya/CommentPlus.js
LilyMakesThings: I'm not going to write this extension for you as a block like that shouldn't exist to begin with

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,372 @@
...
+
+    commentBooleanPlus(args) {
+      // no-op
+    }
+
+    commentReporterPlus(args) {
+      // no-op
+    }
+
+    commentCap() {
```
LilyMakesThings, line 335: Remove your unused functions

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,372 @@
...
+
+    commentBooleanPPPP(args) {
+      return args.INPUT;
+    }
+
+    commentBooleanPlus(args) {
+      // no-op
+    }
+
+    commentReporterPlus(args) {
```
LilyMakesThings, line 331: Please don't mark issues as resolved unless they are **actually** resolved. This still does not return anything. That is a **bug.** Please fix it

extensions/MYTAditya/CommentPlus.js
```diff
@@ -0,0 +1,372 @@
...
+
+    commentReporterPPPP(args) {
+      return args.INPUT || false;
+    }
+
+    commentBooleanPPPP(args) {
+      return args.INPUT;
+    }
+
+    commentBooleanPlus(args) {
```
LilyMakesThings, line 327: Please don't mark issues as resolved unless they are **actually** resolved. This still does not return anything. That is a **bug.** Please fix it

extensions/MYTAditya/CommentPlus.js
LilyMakesThings: In an ideal world you'd just remove **all** of the color blocks

LilyMakesThings (changes requested):

GarboMuffin:
Please do not mark conversations as resolved if they are not resolved.

GarboMuffin:
sorry to have to say no, but I don't see much in here that is really an upgrade beyond the existing comment extension, and we definitely don't need two competing comment extensions. if there's something innovative here I'm missing, please add it to the existing extension

extension review is currently months behind -- very sorry for the delay getting to that decision

PR #1217, opened 2023-12-30
Update extension-runtime.json

qxsck:
revised Chinese translation,the translation on transifex has been updated.


GarboMuffin:
It will be pulled as part of the next batch

PR #1218, opened 2023-12-30
Node 20

GarboMuffin:
We now support Node.js 16 and later; v14 has been dropped. v20 is the 'intended' version.


CubesterYT:
YES!

Issue #236, opened 2023-03-07
box2d extension - some suggestions


macimas:
1. `setup stage [boxed stage]` should be based on actual stage size
currently it sets up walls around 480×360. i think it would be nice if it makes the walls based on actual stage size

https://user-images.githubusercontent.com/42163944/223418023-ffa01552-cf75-4c14-9163-b7393ce15038.mp4

<br>

~~2. replace `set density […v] roughness […v] bounce […v]` with something that's similar to this:~~<br>
something similar was added in pull request #488
![image](https://user-images.githubusercontent.com/42163944/227524157-2c29e056-f9c5-465e-a01c-047b2cc470d3.png)

macimas:
i decided to add another suggestion here. lmk if i should move it to its own issue

GarboMuffin:
believe everything here has been added

PR #1224, opened 2024-01-03
box2d: support custom stage size

MacaylaMarvelous81:
This PR makes the Box2D extension create boundaries around the stage size when the stage size is (re)set to boxed, instead of using set sizes and positions for a 480 pixel by 360 pixel stage. For the floor stage type, it scales the floor width, since the floor was originally always 5000 units in width which could make stages 5000+ pixels in width not have a full-width floor.

It also adds comments where the stage type is set to explain the code and numbers. I think it could be useful since it took me many moments to figure out what the magic numbers meant.

**New Constants**
```patch
- bodyDef.position.Set(0, 1000 / zoom);
+ bodyDef.position.Set(0, (stageBounds.top + 820) / zoom);
...
- fixDef.shape.SetAsBox(10 / zoom, 800 / zoom);
+ fixDef.shape.SetAsBox(10 / zoom, (stageHeight + 820) / zoom);
```
Originally the ceiling was positioned at 1000 units, so to get 820 I subtracted 180 (default top bound) from 1000.

```patch
+ const floorY = (stageBounds.bottom - 100) / zoom;
```
Originally the floor boxes were positioned at `-280 / zoom`. I think this is from a y of -180 (bottom stage bound) with 100 units subtracted because the boxes have a height of 100.

```patch
- fixDef.shape.SetAsBox(5000 / zoom, 100 / zoom);
+ fixDef.shape.SetAsBox((stageWidth + 4520) / zoom, 100 / zoom);
```
I just subtracted 480 from 5000 since the stage is 480 pixels wide by default and the boxes were originally 5000 units wide.
___
This PR resolves suggestion 1 from #236.

### DIFF
--- a/extensions/box2d.js
+++ b/extensions/box2d.js
@@ -12435,12 +12435,12 @@
       // create one such box at the bottom of the stage, accounting for thickness...
       bodyDef.position.Set(0, (stageBounds.bottom - 10) / zoom);
       createStageBody();
-      // and one 640 units above the top of the stage.
-      bodyDef.position.Set(0, (stageBounds.top + 640) / zoom);
+      // and one 820 units above the top of the stage.
+      bodyDef.position.Set(0, (stageBounds.top + 820) / zoom);
       createStageBody();
       // For the left & right wall boxes...
-      // use a height equivalent to the stage height + 640, with a thickness of 10
-      fixDef.shape.SetAsBox(10 / zoom, (stageHeight + 640) / zoom);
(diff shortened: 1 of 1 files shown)


MacaylaMarvelous81:
Oops, I forgot: I was configuring my gpg keys while working on this since I'm on a "new" computer, which is why c15046d doesn't have a signature. If it's a problem I could try signing it and force-pushing to my branch.

GarboMuffin:
We don't really use signed commits here so that's okay (prs generally get squashed anyways)

PR #1226, opened 2024-01-05
Update var-and-list.js

qxsck:
add "add list to scratch list" and "reverse scratch list" blocks


GarboMuffin:
Everywhere you do

```js
this.formatMessage('qxsckvarandlist.name')
```

you should be doing

```js
Scratch.translate({
  default: "Variable and list",
  description: ""
})
```

Or if no description is needed:

```js
Scratch.translate("length of list [LIST]")
```

And remove the inlined Chinese translations, otherwise our bot won't be able to understand what you're writing (and then submit the translations to https://docs.turbowarp.org/translate when it gets merged)

qxsck:
> Everywhere you do
> 
> ```js
> this.formatMessage('qxsckvarandlist.name')
> ```
> 
> you should be doing
> 
> ```js
> Scratch.translate({
>   default: "Variable and list",
>   description: ""
> })
> ```
> 
> Or if no description is needed:
> 
> ```js
> Scratch.translate("length of list [LIST]")
> ```
> 
> And remove the inlined Chinese translations, otherwise our bot won't be able to understand what you're writing (and then submit the translations to https://docs.turbowarp.org/translate when it gets merged)

fix it

PR #1241, opened 2024-01-14
Add Google-Spreadsheets

SharkPool-SP:
image


CubesterYT:
I can't wait to push TurboHook v2 so that this works better and would only require one extension dependency.

ZXMushroom63:
Just tested, and it works very well!
One thing: maybe rename the get data block to specify that it returns JSON.

SharkPool-SP:
> Just tested, and reading works well.
> 
> One thing to do is rename the get data block to specify that it returns JSON. This way less experiences scratchers who can't recognise it by it's syntax know which other extensions to import to use alongside this extension.

might be just me, but if you are able to understand webhooks and http, I think youll be able to recognize json. However, I might change it when Turbohook V2 releases (ask @CubesterYT for more info)

SharkPool-SP:
Minor Update to allow you to choose what spreadsheet to get data from:
image

PR #1245, opened 2024-01-15
New Extension: Pause

SharkPool-SP:
## Various Utilities Related to Pausing
You can (un)pause:
-Scripts
-Sprites
-Project
image


extensions/SharkPool/Pause-Utilities.js
```diff
@@ -0,0 +1,347 @@
...
+
+  runtime.on("PROJECT_STOP_ALL", () => { storedScripts = {} });
+
+  function beginScan() {
+    let isChecking = false;
+    const check = () => {
+      if (isChecking) return;
+      const isPM = Scratch.extensions.isPenguinMod;
```
LilyMakesThings, line 29: It's hard enough to make sure that the code in these extensions will work in TurboWarp forever, even after you leave.

It's even harder to check if your code is going to work for other mods forever. We can't do that. Please don't write code like this.

LilyMakesThings (changes requested):
.

extensions/SharkPool/Pause-Utilities.js
SharkPool-SP: So just remove the penguinmod support?

SharkPool-SP: Cuz even though it does work both ways, it's nice to load the project in and out of penguinmod without issues

LilyMakesThings: Remove the PenguinMod support. They have a gallery, if you'd like to support a version for them, do it there. It's not our responsibility to make sure your code works on a different mod.

SharkPool-SP: Oki

SharkPool-SP: Done

mcnole25:
Oh no, now we're going to need to add a "super pause button" so the project can actually pause :P

SharkPool-SP:
> Oh no, now we're going to need to add a "super pause button" so the project can actually pause :P

:trollface:

SharkPool-SP:
Improved the Code, and added a new Block:
image

CubesterYT:
!format

SharkPool-SP:
rewrote everything to be, cleaner and better optimized. Fixed licensing issue and more

SharkPool-SP:
!format

extensions/SharkPool/Pause-Utilities.js
```diff
@@ -0,0 +1,394 @@
...
+      const spriteNames = [
+        { text: Scratch.translate("myself"), value: "_myself_" },
+      ];
+      if (includeStage)
+        spriteNames.push({
+          text: Scratch.translate("Stage"),
+          value: "_stage_",
+        });
+      const targets = runtime.targets;
+      for (let i = 1; i < targets.length; i++) {
```
yuri-kiss: this is not the correct way to skip the stage, use "isStage" instead

`let i = 1`

yuri-kiss (dismissed):
this is NOT and will not be a full review.
just pointing out smth i noticed you did in a commit

SharkPool-SP:
!format

SharkPool-SP:
!format

SharkPool-SP:
!format

SharkPool-SP:
!format

PR #1250, opened 2024-01-20
Fan Requested Changes Custom Styles

SharkPool-SP:
Whats new:
-Multi gradient support (radial and conic) (example usage: Color Master on my gallery):
image

-Ability to set Question Text Color:
image


qxsck:
good

PR #1254, opened 2024-01-23
New extension: Extendable Blocks

CST1229:
Blocks that extend, like in Snap<i>!</i>.
image

This uses a ton of scratch-blocks patches and also may be considered a utilities extension? (Though splitting this would require duplicating a bunch of code so...) So I don't know if this will be accepted or not.


LilyMakesThings:
I wouldn't consider this a Utilities extension because there is a consistent theme between each block - it also makes more practical sense to include these blocks into one extension because of the patches you're doing to ScratchBlocks.

fo-od:
i think having + and - icons would make more sense since you adding more, but it also makes sense to have arrows because the extension is calling extendable blocks

VoxelDoesCode:
The two arrows feel more in line with Scratch's design language, in my opinion.

LilyMakesThings:
The arrows came from the dropdown menus so they match the style, too.

Arrows are also more consistent with other familiar mods, such as Snap_!_

DilemmaGX:
image
I don't know why, but this extension cannot run on some versions of Turbowarp Desktop (for example, in this case, `v1.9.3`). Anyway, this extension is great running in turbowarp.org.

CST1229:
> I don't know why, but this extension cannot run on some versions of Turbowarp Desktop (for example, in this case, `v1.9.3`).

Use the latest 1.10 beta; this extension requires some things that were only added in the merge-upstream update (the update which added color theme options and a bunch of addons).

Gen1x-ALT:
this looks based

DilemmaGX:
Hello, may I embed it into [GekoWarp](https://dilemmagx.github.io/GekoWarp)'s extension library?

PPPDUD:
I haven't taken an in-depth review yet, but it looks interesting.

LSPECTRONIZTAR:
one prob: the json block will return 2 items even if there are more than 2 inputs

yuri-kiss (changes requested):
I haven't looked at the code or anything nor have I tested this, I am just making this review to hopefully stop someone from merging this to early before we have confirmation on these... patchy type extensions.

CubesterYT:
> I haven't looked at the code or anything nor have I tested this, I am just making this review to hopefully stop someone from merging this to early before we have confirmation on these... patchy type extensions.

Doubt ANYONE feels ok merging patch type extensions 😬

PR #1257, opened 2024-01-25
fix: downloadBlob isn't working properly on some old browsers

FurryR:
Affected browsers:
- Android WebView in mark.via (92.0.4515.105)


GarboMuffin:
I will note that your original patch broke the extension in all released versions of Safari

PR #1258, opened 2024-01-25
lab/text: holiday sized update

LilyMakesThings:
image


GarboMuffin:
- you know how you just replace the animate block when someone enables the non-scratch lab features? do the same thing for fonts
 - shake effect is a bit goofy. not a big fan of the blur, and the text ends up getting cut off. it's 1AM right now so maybe there's an easy solution but I'm not seeing it

LilyMakesThings:
> * you know how you just replace the animate block when someone enables the non-scratch lab features? do the same thing for fonts

The code for this was already there, I just forgot to make the menu accept reporters when I moved everything to the upstream version. (I wrote this update on an older build originally).



> * shake effect is a bit goofy. not a big fan of the blur, and the text ends up getting cut off. it's 1AM right now so maybe there's an easy solution but I'm not seeing it

If you're not a fan it can be removed; it was moreso a test to see how adding new effects would hold up.

WAYLIVES:
If you click on the "Enable Non-Scratch Lab Features" button, then one block disappears, while the very important "set font name", which is why you will now not be able to change to another font.

LilyMakesThings:
> If you click on the "Enable Non-Scratch Lab Features" button, then one block disappears, while the very important "set font name", which is why you will now not be able to change to another font.

Can you show me a screenshot of this? It appears to work fine for me.

WAYLIVES:
> > If you click on the "Enable Non-Scratch Lab Features" button, then one block disappears, while the very important "set font name", which is why you will now not be able to change to another font.
> 
> Can you show me a screenshot of this? It appears to work fine for me.

Oh, I'm sorry, it's okay, I made a mistake. ;)

PR #1260, opened 2024-01-27
Various: Fix compatibility for old VM builds

LilyMakesThings:
Scratch.translate is now a required function in a lot of the original few custom extensions that are listed in older versions of the vm, when extensions were recently introduced.

However, those builds do not have access to Scratch.translate, and neither do many forks of those builds. This makes the extensions completely unusable. (See https://experiments.turbowarp.org/)
image

For the sake of compatibility I think a good compromise is to include a few lines to the main 3 extensions to fix this.

### DIFF
--- a/extensions/cursor.js
+++ b/extensions/cursor.js
@@ -5,6 +5,12 @@
 (function (Scratch) {
   "use strict";
 
+  // For older versions of TurboWarp.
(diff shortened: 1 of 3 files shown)


GarboMuffin:
i'll just update pointerlock

PR #1066, opened 2023-09-30
Object extension

Skyhigh173:
# blocks
image

> [!NOTE]
> You might think that JSON extension is enough, however Object extension is **way more faster** than json extension, as data will not be reparsed.
> All variables **will be converted to JSON string** when you save the project.

# performance/benchmark[^1]
- a ~8.4MB file:
image
image

- a json string with length of 200
image
15x faster.

# Miscellaneous

(Read `IMPORTANT` part)

> [!IMPORTANT]
> it will be slower to parse/stringify the entire object due to reviver & replacer.

> [!IMPORTANT]
> variable monitor will not work properly, eg it shows `keyvalue` instead of `{"key":"value"}`

> [!IMPORTANT]
> use Cast Extension or `() to JSON` block to convert it to string.

> [!WARNING]
> this extension overwrites `Map.prototype.toString`.
> `Map._e` (extra utils) were added.

<details>
<summary>for developers (Map._e extra utils)</summary>

- use `Map._e.deepCopy` when you need to modify the content of the `Map`.
- use `Map._e.shallowCopy` when you only needs to read the content of the `Map`.
- use `Map._e.fromMapOrString` in set content related blocks.[^2]
- use `Map._e.fromArray` when you need to convert array to `Map`.
</details>

[^1]: Parsing time not counted.
[^2]: This function tries convert a string to `Map` if possible.

PR #1261, opened 2024-01-28
Cast: BigInt/Object Compatibility + New Block

LilyMakesThings:
I noticed recently that because of the way I programmed it, Cast doesn't read types aside from string, number and boolean. This is fine for vanilla TurboWarp, but for extensions like BigInt and the potential new Object extension, that's a problem.

(I also accidentally closed this a few weeks ago and didn't realise, but it's back now)


Drago-Cuven:
... please. merge.

extensions/Lily/Cast.js
```diff
@@ -43,19 +43,39 @@
...
+          },
         ],
         menus: {
           type: {
             acceptReporters: true,
             items: ["number", "string", "boolean", "default"],
           },
+          typeObject: {
+            acceptReporters: true,
+            items: ["number", "string", "boolean", "object"],
```
yuri-kiss, line 70: under any normal circumstance objects would not be returned, maybe just remove it, seeing as you can just use a join block to check other types that "may be returned" I don't see why a type that can break serialization should be an option as it shouldnt be returned normally.

yuri-kiss: discarding the fact that stuff like Unsandboxed exists or stuff like @SharkPool-SP JSON extension as this is meant for TurboWarp and not the others, as you know.

yuri-kiss: The https://github.com/TurboWarp/extensions/pull/1066 is generally buggy outside of PenguinMod from experience.

yuri-kiss: please correct me if im wrong*

Skyhigh173: > The https://github.com/TurboWarp/extensions/pull/1066 is generally buggy outside of PenguinMod from experience.

What's the bug? I want to know that

yuri-kiss:
Closed for being stale.

PR #1265, opened 2024-01-29
local-storage: assign unique UIDs by default

LilyMakesThings:
Extension storage is a wonderful thing.

### DIFF
--- a/extensions/local-storage.js
+++ b/extensions/local-storage.js
@@ -37,7 +37,7 @@
     }
     extStorage = runtime.extensionStorage["localstorage"];
     namespace = extStorage.namespace;
-  }
+  };
 
   const getFullStorageKey = () => `${PREFIX}${namespace}`;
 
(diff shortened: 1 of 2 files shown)


extensions/local-storage.js
```diff
@@ -5,29 +5,38 @@
...
+    const soupLength = soup_.length;
+    const id = [];
+    for (let i = 0; i < length; i++) {
+      id[i] = soup_.charAt(Math.random() * soupLength);
     }
-    return valid;
+    return id.join("");
   };
 
+  if (!runtime.extensionStorage["localstorage"]) {
```
GarboMuffin, line 31: think you're gonna have the same problem as the discord extension here

GarboMuffin: evidently there should probably be a simpler api for common usecases like this

LilyMakesThings: Good call

GarboMuffin:
https://github.com/TurboWarp/extensions/pull/2270

PR #1267, opened 2024-01-31
Add "Variables Extended"

Ashimee:
Allows for better variable control and usage. (Some of it is VERY json focused)

Blocks:
image

note:
  delete and create variable blocks are missing because they break when packaged and are unsafe / made redundant


Ashimee:
i hate git sometimes

VoxelDoesCode:
I tested this, and I do want to point out that in `changeVar`, if a variable is not available, it seemingly runs forever (or a long time, haven't tested when it ended). So if you use it in a loop, it stops completely when on the `changeVar` block. The `setVar` block does nothing and lets the loop continue, as expected.

Ashimee:
> I tested this, and I do want to point out that in `changeVar`, if a variable is not available, it seemingly runs forever (or a long time, haven't tested when it ended). So if you use it in a loop, it stops completely when on the `changeVar` block. The `setVar` block does nothing and lets the loop continue, as expected.

weird, i did not notice thx

Martinelplayz:
why is this closed then?

Ashimee:
I know what I just did, It was on purpose.

PR #1270, opened 2024-02-02
Var and list colors

Faris90:
added colors cuz motion+ did it


LilyMakesThings:
that was one of the pull requests of all time.

Faris90:
> that was one of the pull requests of all time.

image

PR #1271, opened 2024-02-02
Var List color

Faris90:
Gives them color


qxsck:
i think it need a new color to distinguish between scratch

qxsck:
can you merge my latest version of PR

PR #1272, opened 2024-02-02
Lily/Hitboxes: Add Hitboxes extension.

LilyMakesThings:
A small extension for setting the hitboxes of targets.
image

I might put this on the gallery if/when I add a few more blocks.

### DIFF
--- a/extensions/Lily/Hitboxes.js
+++ b/extensions/Lily/Hitboxes.js
@@ -5,7 +5,7 @@
   const runtime = vm.runtime;
   const renderer = vm.renderer;
 
-  class TurboCanvas {
+  class Hitboxes {
     constructor() {
       runtime.on("BEFORE_EXECUTE", () => {
         runtime.targets.forEach((target) => {
@@ -89,5 +89,5 @@
     }
   }
 
-  Scratch.extensions.register(new TurboCanvas());
+  Scratch.extensions.register(new Hitboxes());
 })(Scratch);
--- a/extensions/Lily/Hitboxes.js
+++ b/extensions/Lily/Hitboxes.js
@@ -1,0 +1,89 @@
+(function (Scratch) {
+  "use strict";
+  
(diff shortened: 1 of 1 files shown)


gomiigg:
FINALLY

extensions/Lily/Hitboxes.js
```diff
@@ -0,0 +1,93 @@
...
+          target.originalSkin = drawable.skin;
+          drawable.skin = target.hitboxSkin;
+        });
+      });
+      runtime.on("AFTER_EXECUTE", () => {
+        runtime.targets.forEach((target) => {
+          if (typeof target.originalSkin === "undefined") return;
+          const drawableID = target.drawableID;
+          const drawable = renderer._allDrawables[drawableID];
+          drawable.skin = target.originalSkin;
```
NexusKitten, line 24:
```suggestion
          if (target.hitboxSkin === drawable.skin) drawable.skin = target.originalSkin;
```
There's a bug where trying to change costumes while a hitbox was active would cause weird behavior. This somewhat fixes the problem, however the hitbox will still behave weirdly if you change a costume and then try to sense if that costume is touching another within the same thread.

EmimalayaYT:
hey uhh... I'm not sure if I'm right or not, but I think the new turbowarp update broke the extension

NexusKitten:
> hey uhh... I'm not sure if I'm right or not, but I think the new turbowarp update broke the extension

It's been broken for a while now, I don't think it's an update issue (or if it was, it'd be an update from a very long time ago)

EmimalayaYT:
> > hey uhh... I'm not sure if I'm right or not, but I think the new turbowarp update broke the extension
> 
> It's been broken for a while now, I don't think it's an update issue (or if it was, it'd be an update from a very long time ago)

It was a bit broken before, but I actually was able to use it. Now it's even more buggy.

PR #1288, opened 2024-02-08
obviousAlexC/penPlus: optimize

David-Orangemoon:
got the demo on my testing chromebook running from 5fps to 15fps also got the mast on my main computer running 15 fps from 10 fps on firefox (Firefox has the weakest webgl version.)


David-Orangemoon:
I can run pen+ on any chromium browser just fine it just seems firefox struggles.

PR #1293, opened 2024-02-11
Update the AI models' names

DNin01:
Microsoft Bing Chat was rebranded to Copilot and Google Bard was renamed to Gemini.


GarboMuffin:
maybe this whole section was a knee-jerk reaction that should be removed but at least let's update it i suppose

PR #1298, opened 2024-02-15
Create BinaryConverter.js

hiruja210:


hiruja210:
The Binary Converter extension allows you to seamlessly convert Text To Binary and Binary To Text.

PR #1303, opened 2024-02-17
Update runtime-options.js

qxsck:
Allow the upper limit of clones to be freely modified and increase the translation of monitoring blocks.

### DIFF
--- a/extensions/runtime-options.js
+++ b/extensions/runtime-options.js
@@ -143,17 +143,6 @@
               },
             },
           },
-          {
-            opcode: "setCloneLimitValue",
-            text: Scratch.translate("set clone limit to [limit]"),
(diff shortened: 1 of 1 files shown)


LilyMakesThings:
It already was though? If you put a join block in it you could access that.

qxsck:
> 不过已经是这样了？如果你在其中放置一个连接块，你可以访问它。

how to use?but i think it would be more convenient for me to do this

GarboMuffin:
you can do this already

image

the translate() fix is definitely something to merge though

GarboMuffin:
the text-inputs-that-are-also-dropdowns inputs would be a better way to add the clone limit input

qxsck:
> you can do this already
> 
> image
> 
> the translate() fix is definitely something to merge though

thanks

PR #1313, opened 2024-02-22
Update var-and-list.js

qxsck:
add some blocks, fix format


extensions/qxsck/var-and-list.js
```diff
@@ -5,292 +5,872 @@
...
-                type: Scratch.ArgumentType.STRING,
+                type: "string",
                 defaultValue: "value",
               },
             },
-            extensions: ["colours_data"],
+          },
+          {
+            opcode: "seriVarsToJson",
+            blockType: "reporter",
```
LilyMakesThings, line 123: Please use `Scratch.BlockType`. That's not a style thing, it's a compatibility thing.

extensions/qxsck/var-and-list.js
```diff
@@ -5,292 +5,872 @@
...
+            text: Scratch.translate(
+              "convert all variables starting with [START] to json"
+            ),
+            arguments: {
+              START: {
+                type: "string",
```
LilyMakesThings, line 129: Please use `Scratch.ArgumentType`. That's not a style thing, it's a compatibility thing.

extensions/qxsck/var-and-list.js
```diff
@@ -5,292 +5,872 @@
...
+            {
+              text: Scratch.translate("delete the excess parts"),
+              value: "d",
+            },
+          ],
+        },
       };
     }
 
+    //from turbowarp-vm
```
LilyMakesThings, line 821: no kidding.

extensions/qxsck/var-and-list.js
```diff
@@ -5,292 +5,872 @@
...
+              LEFT: {
+                type: "string",
+                defaultValue: "1",
+              },
+              RIGHT: {
+                type: "string",
+                defaultValue: "5",
+              },
             },
-            extensions: ["colours_data_lists"],
```
LilyMakesThings, line 279: These colours make the blocks the same colour as their intended categories. It's good for the recognition of blocks, so don't remove them.

extensions/qxsck/var-and-list.js
```diff
@@ -5,292 +5,872 @@
...
+                type: "string",
+                menu: "associateList.List",
+              },
+            },
+          },
+
+          {
+            opcode: "unSupportCompile",
+            blockType: "label",
+            text: Scratch.translate("⚠️Compilation not supported"),
```
LilyMakesThings, line 716: Compilation **is** supported, you're on TurboWarp, not Gandi IDE.

extensions/qxsck/var-and-list.js
```diff
@@ -5,292 +5,872 @@
...
               },
+              DO: {
+                type: "string",
+                menu: "associateList.List",
+              },
+            },
+          },
+
+          {
+            opcode: "unSupportCompile",
```
LilyMakesThings, line 714: You don't need an opcode for labels.

extensions/qxsck/var-and-list.js
```diff
@@ -5,292 +5,872 @@
...
+            blockType: "label",
+            text: Scratch.translate("⚠️Compilation not supported"),
+            arguments: {},
```
LilyMakesThings, line 717: You don't need arguments for labels.

extensions/qxsck/var-and-list.js
```diff
@@ -5,292 +5,872 @@
...
+              RIGHT: {
+                type: "string",
+                defaultValue: "5",
+              },
             },
-            extensions: ["colours_data_lists"],
           },
         ],
+        menus: {
+          "openCaseSensitive.List": [
```
LilyMakesThings, line 768: Menus should NOT be formatted like this: this locks them into not accepting reporters. See the [documentation on menus](https://docs.turbowarp.org/development/extensions/inputs#static-menus) for more details.

extensions/qxsck/var-and-list.js
```diff
@@ -359,134 +1016,465 @@
...
+          }
         }
       }
       return 0;
     }
     getIndexesOfList(args, util) {
       /** @type {VM.ListVariable} */
       const variable = util.target.lookupVariableByNameAndType(
-        Scratch.Cast.toString(args.LIST),
+        String(args.LIST),
```
LilyMakesThings, line 1226: Use Scratch.Cast.toString. Don't use "String".

LilyMakesThings (changes requested):
I can tell this is just a pseudo-port of the Gandi IDE version of this extension. Please understand when you submit extensions to TurboWarp, we expect you to follow the standards and practises guided in the TurboWarp documentation and contributing guidelines. Things like `Scratch.Cast`, `Scratch.ArgumentType` and so on are generally not optional, especially if they were already used as removing those could break compatibility, especially in the future. If Gandi IDE decided not to support what was already in this extension prior, like `Scratch.Cast`, that is their problem, the same way we don't have hat parameters.

Furthmore, I don't really understand the point of this extension; I can see why it would be useful in Gandi IDE, a place that does not have access to the TurboWarp Extension Gallery, but over here a lot of these needs have already been filled.

GarboMuffin:
haven't even read the above comments but this is really hard to review because the diff is the entire extension, which is rather unusual when updating something that was already working...

LilyMakesThings:
> haven't even read the above comments but this is really hard to review because the diff is the entire extension, which is rather unusual when updating something that was already working...

tldr: it's a port of the gandi ide version, which has less APIs, so it was downgraded.

extensions/qxsck/var-and-list.js
qxsck: does now Turbowarp supports compiling C-type blocks?

extensions/qxsck/var-and-list.js
qxsck: does now Turbowarp supports compiling C-type blocks?

extensions/qxsck/var-and-list.js
qxsck: chang it now compromise compatibility

extensions/qxsck/var-and-list.js
qxsck: OK

extensions/qxsck/var-and-list.js
qxsck: OK

extensions/qxsck/var-and-list.js
qxsck: i read it in turbowarp-vm

extensions/qxsck/var-and-list.js
qxsck: thank

extensions/qxsck/var-and-list.js
qxsck: different colors make it easier to distinguish them from the Scratch version

extensions/qxsck/var-and-list.js
LilyMakesThings: yes

qxsck: > yes

ok,delete it just now

PR #1326, opened 2024-02-25
update bigint, data analysis, var and list

qxsck:
add translate, support bigint to input


qxsck:
I'm not good at using GitHub and deleted my own branch without understanding it, Sorry.

PR #1327, opened 2024-02-25
Update var-and-list.js

qxsck:


LilyMakesThings:
If you're worried about the number of commits on a pull request, they get squashed down anyway; you don't need to reopen it 🙂

qxsck:
> If you're worried about the number of commits on a pull request, they get squashed down anyway; you don't need to reopen it 🙂

it is so sorry, and is it has questions? can marge it?

PR #1333, opened 2024-02-27
Lily/Skins: Remove "lms-" when returning current skin

SharkPool-SP:
Forgot to do this, sorry! :)


extensions/Lily/Skins.js
```diff
@@ -326,7 +326,7 @@
 
       const skinId = renderer._allDrawables[drawableID].skin._id;
       const skinName = this._getSkinNameFromID(skinId);
-      return skinName ? skinName : "";
+      return skinName ? skinName.replace("lms-", "") : "";
```
LilyMakesThings, line 329: You should substring it, not replace it

PR #1336, opened 2024-02-28
obviousAlexC/penPlus: fix missing renderer.dirty = true

David-Orangemoon:
Didn't factor in renderer.dirty when switching draw regions.
A penguinmod user alerted me of this.


extensions/obviousAlexC/penPlus.js
```diff
@@ -2536,7 +2532,7 @@
   //? A small hack to stop the renderer from immediatly dying. And to allow for immediate use
   {
     if (!Scratch.vm.renderer._penSkinId) {
-      Scratch.vm.renderer.createPenSkin();
+      window.vm.renderer.createPenSkin();
```
GarboMuffin, line 2535: why

PR #1339, opened 2024-02-29
Indexed db extension

ffghdfh:
Added an IndexedDB extension.

### DIFF
--- a/extensions/ffghdfh/indexedDB.js
+++ b/extensions/ffghdfh/indexedDB.js
@@ -1,7 +1,7 @@
 // Name: Indexed Database
 // ID: ffghdfh-IndexedDB
 // Description: Lets you use an IndexedDB database, allowing for you to store more than 5MB of data locally.
-// License: MIT AND LGPL-3.0
+// License: LGPL-3.0
 (function(Scratch) {
     "use strict";
   if (!Scratch.extensions.unsandboxed) {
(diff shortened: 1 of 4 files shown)


extensions/ffghdfh/indexedDB.js
```diff
@@ -0,0 +1,221 @@
...
+    // Get value from the IndexedDB
+    _getValue(key) {
+      return this._initDB().then((db) => {
+        const transaction = db.transaction(['data'], 'readonly');
+        const store = transaction.objectStore('data');
+        const request = store.get(key);
+  
+        return new Promise((resolve, reject) => {
+          request.onsuccess = () => resolve(request.result ? request.result.value : null);
+          request.onerror = () => reject(request.error);
```
CST1229, line 67: Slight issue, in unsandboxed extensions [blocks throwing errors can break stuff](https://docs.turbowarp.org/development/extensions/unsandboxed#increased-power-brings-increased-responsibility). You should probably just resolve with an empty string (for reporters) or undefined (for commands) in error cases. (This also applies to all other times `reject()` is used in blocks.)

ffghdfh: Oh gee, didn't know that! Thanks for letting me know! I'm rally new to github though, so how do I modify my changes 'n stuff?

ffghdfh:
I removed all the reject()-s, replaced 'em all with resolve()-s, although i'm a bit tired, it's getting late, so it might be a bit lame of a solution 😅

PR #1351, opened 2024-03-10
CST1229/zip: multi-archive

CST1229:
A major update to the Zip extension, mainly adding support for multiple archives at a time. Marked as draft because it probably still needs bug-finding.
image


Martinelplayz:
I'll do some testing

Martinelplayz:
it works just fine for me. I haven't found any bugs since I started using it.

Drago-Cuven:
breh ... this should have been accepted so long ago.

LilyMakesThings:
> breh ... this should have been accepted so long ago. its useful, an upgrade from the original, and works flawlessly as far as I've seen.

merges are slower because we've all got other stuff to do.

PR #1360, opened 2024-03-15
Add Tab Utilities

chummos:
My first extension. It's a collection of blocks that interacts with the tab.

It has 4 blocks:
Close Tab Confirmation -- Enables the browser pop-up that confirms you want to exit the tab (usually says something like "You may have unfinished changed")

Change Title -- Changes the title of the tab.

Change Favicon -- Changes the favicon of the tab.

Title -- a reporter block that returns the tab's title.

### DIFF
--- a/extensions/frostworx/TabUtils.js
+++ b/extensions/frostworx/TabUtils.js
@@ -1,0 +1,90 @@
+// Name: Tab Utilities
+// ID: tabUTILS
+// Description: A collection of blocks that interact with the tab.
+// By: Frostworx
+// Original: Frostworx
+// License: CC-BY-SA-2.0
+
+(async function(Scratch) {
+    const variables = {};
+    const blocks = [];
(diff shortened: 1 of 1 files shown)


chummos:
Sorry if the grammar is a bit rough, I wrote this rather quickly

Martinelplayz:
unnecessary due to the window control extension which is already merged.

Martinelplayz:
although @CubesterYT might want to add that favicon block.

CubesterYT:
I wanted to add the favicon block to window controls during initial development, but I discarded it because it only works in browser, not packaged apps like electron. HTML and in browser only worked. I could look into it again.

CubesterYT:
Also I'd like to ask the extension creator what format they used to make this extension, as this is not TurboWarp format at all. Potential AI or alternate format usage?

PR #1362, opened 2024-03-16
Mozilla Public License 2.0

GarboMuffin:
The GPL family of licenses are not perfect. Notably they are incompatible with
the SIL OFL 1.1 used by the fonts in scratch-render-fonts. In the packager we
'statically link' all of this those fonts into one HTML file. Is that legal?
Seems like you can make an argument that it's not. Who knows. I'm not a lawyer.

MPL 2.0 is a good alternative: It's still copyleft, used by real projects,
backed by a real organization, and GPL compatible.

My own extensions/contributions are now MPL-2.0, as is recommended.
LGPL-3.0 should continue to be allowed here. Use it if you want.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,7 +44,7 @@ Extensions must be self-contained. All libraries and hardcoded resources should 
 
 **We are not lawyers. This is not legal advice.**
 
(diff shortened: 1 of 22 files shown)


CubesterYT:
Another license change, let's go

GarboMuffin:
i love software licenses

GarboMuffin:
copyright law is awesome

PR #1363, opened 2024-03-16
Gandi IDE (CCW) Layer Management

FurryR:
# Features

- [x] Manage sprites & clones layers. Pretty simple.
- [x] 100% compatible with Gandi IDE version. This means that by changing the project a little, you can make your project load on Gandi IDE, vice versa.

# Checks Bypassed

- [x] Permission from Gandi
- [x] Prettier
- [x] ESLint

# Differences between Turbowarp and Gandi IDE

1. On Turbowarp version, you are not able to use Gandi Terminal feature. Alternatively we use `DevTools Console`.
2. Since Turbowarp does not support Gandi Quake or CCWCanvas, some blocks will be hidden on Turbowarp.
3. Turbowarp version injects renderer (more elegant) while Gandi IDE version modifies renderer source code to make layer manager work.
That is, on Gandi IDE, layer manager (`vm.renderer.layerManager`) is available **even without the extension**, while on Turbowarp you have to load the extension to make layer manager (`vm.renderer.layerManager`) available.
`vm.renderer.layerManager` is unset before you load the extension, so some advanced extensions, addons or userscripts that requires layer manager may fail to work. Please do not report that as a bug.
Other extensions like Spine Manager or Gandi Quake do modify renderer source code like mentioned above as well, so same issue may happen if they are ported one day.

# Staff

- Original version by Arkos@CCW and Gandi IDE internal code database
- Compatibility (thats probably a miracle actually) by FurryR


fath11:
grrr

qxsck:
wow

translations/extension-runtime.json
```diff
@@ -3193,7 +3193,45 @@
...
+        "Gandi/LayerManager@behind": "后面",
+        "Gandi/LayerManager@ascending": "升序",
+        "Gandi/LayerManager@descending": "降序",
+        "Gandi/LayerManager@activate": "启动",
+        "Gandi/LayerManager@deactivate": "关闭",
+        "Gandi/LayerManager@affected": "受",
+        "Gandi/LayerManager@unaffected": "不受",
+        "Gandi/LayerManager@level": "图层层级",
+        "Gandi/LayerManager@myPriority": "层级内排序值",
+        "Gandi/LayerManager@myLayer": "图层序号"
```
CST1229, line 3234: I'm pretty sure extension translations are automatically fetched from Transifex, so this should be removed.

FurryR: Okay I will restore them.

chummos:
Poggers :fishsticks:

Arkos123:
https://github.com/FurryR/extensions/blob/9c273feede816b0df203d4f7f81f9d46092407dd/samples/Layer%20Manager.sb3
example.sb3 added
- Sprites can be added to different layer levels. (e.g., buttons in the 'UI' layer will be at the top layer)
- Sprites in the same layer level can be sorted by the z-index (e.g., sorted by y position so that the sprites with higher y values will be layered behind).
image

CubesterYT:
You don't need to close any PRs, it's just Garbo is quite busy, but he'll get around to everything.

PR #1364, opened 2024-03-16
TheShovel/ColorPicker: add sample project

DNin01:
Adds a sample project that uses the Color Picker extension called "Recolor the Dango".


GarboMuffin:
do you actually need the wait (0.01)? seems to work for me without it

DNin01:
> do you actually need the wait (0.01)? seems to work for me without it

When I was testing, the color picker's position wouldn't update the first time I click, which is why I added the delay.

DNin01:
https://github.com/TurboWarp/extensions/assets/106490990/2cee9732-72a1-4861-9965-e6adaedaf4a5

DNin01:
I have removed the delay because #1365 was opened.

PR #1365, opened 2024-03-17
TheShovel/ColorPicker: fix incorrect position if moved and shown in same frame

GarboMuffin:
image


GarboMuffin:
@DNin01 does this fix it for you?

DNin01 (approved):
Yes, it does!

PR #1366, opened 2024-03-17
Longboost/color_channels: Change boolean inputs to menus

DNin01:
Workspace picture showing true boolean and only draw colors boolean input green boolean input blue input removed and set colors red menu green menu blue menu added

(Of course, the old blocks weren't removed, but just hidden from the palette.)

### DIFF
--- a/extensions/Longboost/color_channels.js
+++ b/extensions/Longboost/color_channels.js
@@ -30,6 +30,7 @@
             opcode: "true",
             blockType: Scratch.BlockType.BOOLEAN,
             text: "true",
+            hideFromPalette: true,
             disableMonitor: true,
           },
           {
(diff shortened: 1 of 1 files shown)


SharkPool-SP:
can it just be rewritten to have custom colors

DNin01:
> can it just be rewritten to have custom colors

The [colorMask](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/colorMask) API of WebGL used by the RGB Channels extension only seems to allow color channels to be turned on or off. I'm not sure if there are other APIs that might allow more color control.

SharkPool-SP:
> > can it just be rewritten to have custom colors
> 
> The [colorMask](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/colorMask) API of WebGL used by the RGB Channels extension only seems to allow color channels to be turned on or off. I'm not sure if there are other APIs that might allow more color control.

bleh, that sucks. lol

PR #1375, opened 2024-03-22
Add Captcha Extension

SharkPool-SP:
### Simple Captcha Extension with no APIs
Simple Captcha I made, although its not perfect, it works.

There is some ways you can cheat the system, but I tried in every way I can to prevent it. Hope you enjoy!
Screenshot 2024-03-21 at 11 00 48 PM


SharkPool-SP:
I don't like this extension

PR #1377, opened 2024-03-22
obviousAlexC/penPlus: v7

David-Orangemoon:
Adding features check the trello https://trello.com/b/yjRaNDDP/pen


mcnole25:
> What I'm currently doing Reorganizing the extension and adding the shader editor

Quick question, why is the first version v5 and not v1?

David-Orangemoon:
> > What I'm currently doing Reorganizing the extension and adding the shader editor
> 
> Quick question, why is the first version v5 and not v1?

Because v6 split off on v5 if you want v1 look here https://raw.githubusercontent.com/David-Orangemoon/Modified-Extension-Loader-Turbowarp/fd981583ff5ed8826bab6449fc517587d7f9a987/custom%20extensions/CustomRenderPipeline.js it's turboloader

VoxelDoesCode:
This is an EXCELLENT update to the pen feature so far! The shader function might be a complete game changer for decorative projects! I've tested it a few times, and I found a few bugs:

On desktop mode, when your window isn't maximized and you open the shader editor, the GLSL tab stays in that exact ratio, even when you maximize. This leaves an ugly black border at the bottom right corner.

The terminal where bugs are shown could be improved so that each error gains its own line. In its current state, it's hard to read when there are more than 2 bugs on screen.

All of these tests are on my Windows 10 computer, and I'm unsure if this is like that for any other OS. Overall I'm more than excited to see what this update has to offer in the future!

David-Orangemoon:
> This is an EXCELLENT update to the pen feature so far! The shader function might be a complete game changer for decorative projects! I've tested it a few times, and I found a few bugs:
> 
> On desktop mode, when your window isn't maximized and you open the shader editor, the GLSL tab stays in that exact ratio, even when you maximize. This leaves an ugly black border at the bottom right corner.
> 
> The terminal where bugs are shown could be improved so that each error gains its own line. In its current state, it's hard to read when there are more than 2 bugs on screen.
> 
> All of these tests are on my Windows 10 computer, and I'm unsure if this is like that for any other OS. Overall I'm more than excited to see what this update has to offer in the future!

I was aware of the glsl thing and I am planning on fixing that today. And thanks for the feedback. I think the hard edge is a bug on the browsers end. I'll see if there is a way I can fix rhag.

David-Orangemoon:
Ok that is the final stuff done. fully. If anybody wants to try newer stuff then pen-group site will have an experimental version. Though that could potentially break projects.

David-Orangemoon:
I think it is ironed out to a sheen now

PR #1384, opened 2024-03-27
New extension: Close Control

XmerStudio:
Tab closure confirmation paths can be activated and deactivated.


GarboMuffin:
@XmerStudio You need to pick which license to use. If you're not sure then just say "Mozilla Public License 2.0". Once you do that, this will get merged within 24 hours. If that doesn't happen then email contact@turbowarp.org because it got lost in my github notifications

GarboMuffin:
If I don't here from you within a couple days I'll just assume MPL-2.0 since it's small enough and I made enough changes that I could get away with whatever I want, but we want to respect the license choices of authors

PR #1390, opened 2024-04-04
Lily/MoreEvents 3 new blocks

veggiecan0419:
Hi everyone!
Here a **3** new blocks.
These are the first two,
2 new blocks
And the third is something I'm sure some of you have been waiting for...

## 2,
# 1,

# 🎉  Showtime!  🎉 
special block

## Now for those who don't see the difference...

Scratch's block is like this:

scratch
And ours is: 

special block

...Which allows us to do some cool things such as:

example1

And many other things too!

### What do they do?

The first one returns the name of the broadcast that started **this** thread, and the other one pretty much explains itself.
Here is a little illustration for you: 

example2

And that's about it, everyone! Let's see how many 👍s we can get. 

> [!IMPORTANT]
> This requires some changes to the VM, so it won't work until those are merged.

## Notes
I'd like to add the following option, but don't know how to add it to the menu:

any
Otherwise, should I just add the following instead?

alternative

Or should we just leave people to do

workaround?

### DIFF
--- a/extensions/Lily/MoreEvents.js
+++ b/extensions/Lily/MoreEvents.js
@@ -2,6 +2,7 @@
 // ID: lmsMoreEvents
 // Description: Start your scripts in new ways.
 // By: LilyMakesThings <https://scratch.mit.edu/users/LilyMakesThings/>
+// By: Veggiecan0419
 // License: MIT AND LGPL-3.0
 
 (function (Scratch) {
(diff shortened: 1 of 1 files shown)


extensions/Lily/MoreEvents.js
```diff
@@ -546,6 +602,16 @@
       return util.ioQuery("keyboard", "getKeyIsDown", [key]);
     }
 
+    whenbroadcastreceived(args, util) {
+      const broadcastOption = Scratch.Cast.toString(args.BROADCAST_OPTION);
+      return broadcastOption === runtime.lmsLastBroadcastSent;
```
veggiecan0419, line 607: Here we can't use `util.thread.broadcastName`, as`startHats` runs the hat code straight away, meaning it runs before broadcastName can even be assigned, which is pretty annoying

veggiecan0419, line 607: Here we can't use `util.thread.broadcastName`, as`startHats` runs the hat code straight away, meaning it runs before broadcastName can even be assigned, which is pretty annoying

CST1229: You could override `vm.runtime.startHats` (and call the original function in there) maybe? I think some extensions already do stuff like that.

veggiecan0419: @CST1229 Thanks for the idea. I was thinking about that, but I wasn't sure if it was possible. I think I'll do that.

veggiecan0419:
> I don't like this
@LilyMakesThings I'll just make a new extension then ☺️

PR #1413, opened 2024-04-18
Xmer Utilities

XmerStudio:
A lot of Advanced Tools!

### DIFF
--- a/extensions/xmerutilities.js
+++ b/extensions/xmerutilities.js
@@ -1,0 +1,907 @@
+// Name: Xmer Utilities [Advanced]
+// ID: xmerutilities
+// Description: A lot of Advanced Tools!
+// By: XmerOriginals <https://scratch.mit.edu/users/XmerOriginals/>
+// License: MPL-2.0
+// Version: 1.0.0
+
+class XmerUtilities {
+	"use strict";
+	
+  constructor() {
+    this.pressedKeys = {};
+    this.setupEventListeners();
+    this.hasInteracted = false;
+    this.closeEnabled = false;
+    this.handleCloseRequest = this.handleCloseRequest.bind(this);
+    this.activeMessages = [];
+    this.customTitle = "Xmer Extensions";
+	this.playingSounds = new Map();
+    this.currentSoundURL = '';
+    this.isPlaying = false;
+    this.errorStatus = '';
+    this.waitingPromise = null;
+    this.volume = 100;
+    this.paused = false;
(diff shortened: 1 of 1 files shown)


CST1229:
Please read [CONTRIBUTING.md](https://github.com/TurboWarp/extensions/blob/master/CONTRIBUTING.md)!

> Strictly, nothing is banned, but the following are *highly* discouraged:
> 
> - **Broad "Utilities" extensions (break them up into multiple extensions, see https://github.com/TurboWarp/extensions/issues/674)**
> - **Extensions that are very similar to existing ones (consider modifying the existing one instead)**
> - One-use personal extensions (load the extension as a local file instead)
> - Extensions whose primary purpose is monetization (not in the spirit of an open source project)
> - Joke extensions (they aren't funny when they cause us to get bug reports)

LilyMakesThings:
This extension will inevitably be declined because we do not except Utilities extensions as per the Contributing Guidelines ([CONTRIBUTING.md](https://github.com/TurboWarp/extensions/blob/master/CONTRIBUTING.md))

XmerStudio:
> This extension will inevitably be declined because we do not except Utilities extensions as per the Contributing Guidelines ([CONTRIBUTING.md](https://github.com/TurboWarp/extensions/blob/master/CONTRIBUTING.md))

i named it like that because I couldn't find a name.

LilyMakesThings:
> > This extension will inevitably be declined because we do not except Utilities extensions as per the Contributing Guidelines ([CONTRIBUTING.md](https://github.com/TurboWarp/extensions/blob/master/CONTRIBUTING.md))
> 
> i named it like that because I couldn't find a name.

It's not the name, it's the extension itself. "Utility" extensions without a consistent theme are not allowed.

XmerStudio:
How can you respond so quickly? Also, have you tried the extension?

LilyMakesThings:
After having to modify the code manually just to get it to function, I can safely say this is not going to be accepted.
image

XmerStudio:
I think you turned off unsandboxed for your security, it doesn't load when it's off, no need to fix it manually when it's on. never mind, also do you wait 24/7 on github, you respond very fast.

qxsck:
image
it is bad

XmerStudio:
I made the code in notepad++. When I look at it from vs code, it looks really complicated.

XmerStudio:
you can use any part of the code you want.

PR #1423, opened 2024-04-23
Lily/SoundExpanded: fix "change project volume" block

CST1229:
Lily messed up the clamping, so it used to just always set the volume to 0. This fixes it.
This also makes the set volume block be clamped to 0-100 instead of being wrapped around 0-199, which I think makes more sense and is consistent with the vanilla set volume block. Though I'm not sure if it wrapping around 199 instead of 100 (or 99) was a mistake or not, or if it'll break projects.


LilyMakesThings (approved):
This was 100% a mistake that I falsely thought I fixed ages ago

PR #1424, opened 2024-04-24
Web GPU Extension

zacoons:
This extension provides WebGPU bindings for TurboWarp.

I have a sample project but it breaks validation because it uses the localhost URL for the extension. So I'll add that after the extension is published and I can use the proper URL.

There is one feature I'd still like to add but it's not necessary and it may be a while before I get around to it.


zacoons:
lmao, looking through the other extensions it seems like I've been topped by that Web3D one.

zacoons:
@PolyPenguinDev you can now create a module with a string. This should make your extension possible. If you wanna talk about it further you can message me on Matrix (link on my website).

zacoons:
I figured some images would be useful so here you go.

image
image
image

I also made a video about it
https://www.youtube.com/watch?v=iwJ3_BGbKdY

extensions/zacoons/webgpu.js
```diff
@@ -0,0 +1,638 @@
...
+      return names.length ? names : [""];
+    }
+    _getLists() {
+      const names = [];
+
+      const globalVars = Object.entries(
+        Scratch.vm.runtime.targets[0].variables
+      );
+      const localVars = Object.entries(Scratch.vm.editingTarget.variables);
+      for (const [varId, varObj] of globalVars.concat(localVars)) {
```
Xeltalliv, line 604: If the selected sprite is stage, names of all lists are duplicated.

extensions/zacoons/webgpu.js
```diff
@@ -0,0 +1,638 @@
...
+          names.push(c.text.slice(5, firstLineEndIdx).trim());
+        }
+      }
+      return names.length ? names : [""];
+    }
+    _getLists() {
+      const names = [];
+
+      const globalVars = Object.entries(
+        Scratch.vm.runtime.targets[0].variables
```
Xeltalliv, line 601: It's better to use vm.runtime.getTargetForStage() which finds stage regardless of where it is in targets array.

extensions/zacoons/webgpu.js
```diff
@@ -0,0 +1,638 @@
...
+              "VERTEX",
+            ],
+          },
+          modules: {
+            acceptReporters: true,
+            items: "_getModuleNames",
+          },
+          lists: {
+            acceptReporters: true,
+            items: "_getLists",
```
Xeltalliv, line 363: acceptReporters is true, but in order to use it, users have to know and provide the list ID instead of name.

extensions/zacoons/webgpu.js
```diff
@@ -0,0 +1,638 @@
...
+      this._resetAttrs();
+
+      if (!navigator.gpu) return false;
+
+      return navigator.gpu.requestAdapter().then(async (adapter) => {
+        this.dev = await adapter.requestDevice();
+        this._createModules();
+        return true;
+      });
+    }
```
Xeltalliv, line 383: When using any modern browser which technically supports WebGPU, but doesn't support it in the specific conditions which it is being used, navigator.gpu will exist, but the promise returned from navigator.gpu.requestAdapter() will get rejected. This makes the block return undefined and show some warning in the console.
For example, Firefox on Linux shows this:
image

extensions/zacoons/webgpu.js
```diff
@@ -0,0 +1,638 @@
...
+
+      /** @type {any} */
+      this.tmp = undefined;
+    }
+    _getModuleNames() {
+      const names = [];
+      for (const commentName in Scratch.vm.editingTarget.comments) {
+        const c = Scratch.vm.editingTarget.comments[commentName];
+        if (c.text.startsWith("name:")) {
+          const firstLineEndIdx = c.text.indexOf("\n");
```
Xeltalliv, line 591: This can be -1.

extensions/zacoons/webgpu.js
```diff
@@ -0,0 +1,638 @@
...
+        Float64: Float64Array,
+        Int64: BigInt64Array,
+        Uint64: BigUint64Array,
+      };
+    }
+
+    getInfo() {
+      return {
+        id: "webgpu",
+        name: "WebGPU",
```
Xeltalliv, line 48: Since WebGPU is used for both rendering and compute, but this extension only provides compute functionality, maybe it's better to call this extension "WebGPU Compute"? Not sure, just an idea.

extensions/zacoons/webgpu.js
```diff
@@ -0,0 +1,638 @@
...
+        Float32: Float32Array,
+        Float64: Float64Array,
+        Int64: BigInt64Array,
+        Uint64: BigUint64Array,
+      };
+    }
+
+    getInfo() {
+      return {
+        id: "webgpu",
```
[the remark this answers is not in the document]
zacoons: makes sense

extensions/zacoons/webgpu.js
zacoons: I was gonna call it WebGPU bc I thought rendering functionality might be added eventually. But with your extension it seems unlikely that will happen, so I'll rename it. Nice job on Simple3D.

extensions/zacoons/webgpu.js
zacoons: I think I've resolved this, but I can't really test it so let me know.
Thanks for your thorough review, I really appreciate it.

extensions/zacoons/webgpu.js
```diff
@@ -0,0 +1,638 @@
...
+    init() {
+      this._resetAttrs();
+
+      if (!navigator.gpu) return false;
+
+      return navigator.gpu.requestAdapter().then(async (adapter) => {
+        this.dev = await adapter.requestDevice();
+        this._createModules();
+        return true;
+      });
```
Xeltalliv: Promise getting rejected does not throw an exception here, so wrapping it in `try {} catch {}` didn't fix it. The old code was fine, it just needed an extra `.catch()`.
```suggestion
      }).catch(() => false);
```

extensions/zacoons/webgpu.js
Xeltalliv: > Nice job on Simple3D.

Thanks!

> But with your extension it seems unlikely that will happen

They aren't really competing though, since both are targeting different people with different goals. The only  competing extension targeting the same people is [my WebGL 2 bindings](https://github.com/TurboWarp/extensions/discussions/378) extension, but I'm not sure if I'll ever finish that.

Xeltalliv:
All other changes look good

extensions/zacoons/webgpu.js
zacoons: oh yeah, beginner mistake, thanks

PR #1426, opened 2024-04-25
Minor improvements to AR and Clipping & Blending

Xeltalliv:


Drago-Cuven:
@Xeltalliv  it could also be used on Vr headsets like Oculus quest.

Xeltalliv:
https://github.com/Xeltalliv/extensions/blob/acea6eb99067ac488320225c7fcb8e77a3c82b0a/docs/ar.md?plain=1#L5

I had to specify that it's mobile devices only, because while it technically launches on headsets, it's not actually usable on them, because it lacks the blocks required to figure out where to render 2 images for 2 eyes, even though internally it gets those values. It outputs all of the camera vlues for only the first eye (which in mobile AR there is only 1, so it's fine there).

PR #1430, opened 2024-04-27
true-fantom/math: add x < y < z blocks

MaelkMark:
I've added a new boolean block that checks whether a number is in a given range.
`<A is in range B - C>`

block opcode: `is_in_range_block`

### DIFF
--- a/extensions/true-fantom/math.js
+++ b/extensions/true-fantom/math.js
@@ -362,6 +362,26 @@
             extensions: ["colours_operators"],
           },
           {
+            opcode: "is_in_range_block",
+            blockType: Scratch.BlockType.BOOLEAN,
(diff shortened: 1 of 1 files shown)


LilyMakesThings:
I think this exists in more comparisons as `() < () < ()`

MaelkMark:
> I think this exists in more comparisons as `() < () < ()`

Oh, I didn't know. However I think it would be practical to have such a block in the math extension too - you don't have to add the more comparisons ext. (I've added that block because I needed a block like that in my project)

GarboMuffin:
you have to use Cast.compare or this will break horribly when it gets a string containing a number instead of an actual number

MaelkMark:
Thanks for the correction and adding the blocks to the extension!

PR #1448, opened 2024-05-09
data-analysis: Add sum block

DNin01:
It's not uncommon to need the sum of a bunch of numbers.


qxsck:
GM hasn't passed my pr yet

Issue #1449, opened 2024-05-09
Make Numerical Encoding length prefixed encoding, or make it an option


snowboyz0825:
As it stands, numerical encoding is really, really inefficient. It (from what I can tell) converts each character into a 5 character long string using it's Unicode. What we can instead do is get whatever the characters Unicode is (e.g. 65 for A) and then "preface it" with how many characters that Unicode is. For example, A would be 256, 2 being how many characters it needs to read after is. 
If we encode the string "Hello World" using both methods, the current extension yields
000072000101000108000108000111000032000087000111000114000108000100 - 66 characters, horrible for limited cloud variables. If we instead use "Length Prefixed Encoding" we get
27231013108310831112322873111311431083100 - 41 characters. 
I also think that the extension should be a reporter, as having to run a block and then run the reporter just seems weird (I suppose with longer strings it could be a performance thing, but I believe we can leave both functionalities in just in case.)
We could probably just leave the old blocks in and make them invisible for compatibility reasons, or make it a new extension entirely. I might make it myself, but I well and truly know almost nothing about coding turbowarp extensions

snowboyz0825:
Nothing beats coding your own thing, but I feel like we should at least work to make the thing we have integrated the best it can be

GarboMuffin:
we could do a lot better than either if we did a proper base 10 encoding

GarboMuffin:
proper base 10 encoding mean something that uses actual binary options in the same way that base 64 and base 85 encodings do

GarboMuffin:
In V2 "Hello World" encodes to 331736654417855136778673441731 - 30 characters

PR #1450, opened 2024-05-09
New extension: Numerical Encoding V2

GarboMuffin:
1. Replaced separate commands and reporters with just reporters (no more ugly global state that can get broken by unlucky warp timer)
2. Significantly optimized encoding format.

The original extension always uses 6 numbers per UTF-8 character in the original text. On ASCII this has an efficiency of 1.33 bits/number.

https://github.com/TurboWarp/extensions/issues/1449 proposed length-prefixed encoding which is a good improvement for pure ASCII strings. Most of the commonly used parts of ASCII have 2 digit or 3 digit decimals codes so let's just say it's an average of about 3.5 numbers per character in the input after adding the length prefix. That's about 2.28 bits/number of efficiency.

This PR does all encoding and decoding operations on UTF-8 binary instead. Each group of 3 bits in the binary is encoded by 1 decimal in [1, 8] (not using 0 to avoid leading zeros), so this has 3 bits/number of efficiency.

In theory we should be able to get $\log_2{10} \approx 3.321$ bits/number of efficiency with some smarter encoding. But I think this is already pretty good and the encoding/decoding routines shouldn't be too hard to port to other languages for people making bots.


snowboyz0825:
As dumb as it is I'm proud I actually encouraged making an extension better

PR #1460, opened 2024-05-11
NOname-awa/graphics2d: Use atan2 to get coord to coord angle

DNin01:
Changed the "direction of _x1_,_y1_ to _x2_,_y2_" function to use Math.atan2. It does the same thing but with less lines of code.

Tested pretty much everything I can think of, atan2(0, 0) is 0, output ranges from -180 to 180, atan2 already auto-converts strings to numbers.

### DIFF
--- a/extensions/NOname-awa/graphics2d.js
+++ b/extensions/NOname-awa/graphics2d.js
@@ -281,19 +281,7 @@
       );
     }
     ray_direction(args) {
-      // Added by NexusKitten
-      // 由 NexusKitten 添加
-      const dx =
-        Scratch.Cast.toNumber(args.x2) - Scratch.Cast.toNumber(args.x1);
-      const dy =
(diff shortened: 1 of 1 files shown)


Xeltalliv:
Scratch.Cast.toNumber also convertes NaN to 0. As a result, all Scratch blocks treat NaN inputs as 0 instead of letting NaN propagate like they do in JS.

GarboMuffin:
you need the Cast.toNumber still

DNin01:
Oh, ok!

DNin01:
@GarboMuffin

GarboMuffin:
image

original gives 225, new ones gives -135. so they're both the same actual angle but I don't know if we want to be changing how this block works

DNin01:
This ok, @GarboMuffin?

PR #1466, opened 2024-05-13
CST1229/images: list on the website

GarboMuffin:
it's merged, it works, may as well actually list it


CST1229:
i was gonna list it after i originally added it but didn't because of looks plus and eventually skins

PR #1468, opened 2024-05-14
Xeltalliv/simple3d: fix other mod compatibility, fix transform to scratch units, improve docs

Xeltalliv:
Sorry for another PR less than 24 hours after the previous one. Both issues aren't critical, so it's okay if this takes a while to merge.


GarboMuffin:
it's better that they fix their bugs so everything just works instead of manually fixing every extension...

PR #1475, opened 2024-05-16
New extension: Anthropic API

NotFenixio:
image
You need an API key in order to use this extension.


NotFenixio:
You can get $5 dollars in free credits.

AshimeeAlt:
> You can get $5 dollars in free credits. I don't think they ever expire too.

still though, go through pulls that require api keys and i can bet they are either broken, rejected or some other form of a no.

NotFenixio:
> still though, go through pulls that require api keys and i can bet they are either broken, rejected or some other form of a no.

Well, there are exceptions. This one uses a prompt instead of a block which is more secure than asking through a block and risking leaking a key, has documentation on how to get one (and how to get the free credits) and is not broken. We'll see if it gets rejected.

(The only reason the workflow is failing is because of Prettier)

AshimeeAlt:
> > still though, go through pulls that require api keys and i can bet they are either broken, rejected or some other form of a no.
> 
> Well, there are exceptions. This one uses a prompt instead of a block which is more secure than asking through a block and risking leaking a key, has documentation on how to get one (and how to get the free credits) and is not broken. We'll see if it gets rejected.
> 
> (The only reason the workflow is failing is because of Prettier)



> > still though, go through pulls that require api keys and i can bet they are either broken, rejected or some other form of a no.
> 
> Well, there are exceptions. This one uses a prompt instead of a block which is more secure than asking through a block and risking leaking a key, has documentation on how to get one (and how to get the free credits) and is not broken. We'll see if it gets rejected.
> 
> (The only reason the workflow is failing is because of Prettier)



> > still though, go through pulls that require api keys and i can bet they are either broken, rejected or some other form of a no.
> 
> Well, there are exceptions. This one uses a prompt instead of a block which is more secure than asking through a block and risking leaking a key, has documentation on how to get one (and how to get the free credits) and is not broken. We'll see if it gets rejected.
> 
> (The only reason the workflow is failing is because of Prettier)

just override the prompt function your using :| or get the class but we will see if it gets rejected, i just wanted to warn you do you dont have to ask why later

NotFenixio:
This is probably better in some third-partu gallery.

PR #1485, opened 2024-05-24
Added safe Eval JavaScript block

BlueDome77:
Runs the JavaScript in the sandbox like this:

block_5_23_2024-7_18_15 PM

### DIFF
--- a/extensions/iframe.js
+++ b/extensions/iframe.js
@@ -153,6 +153,17 @@
               },
             },
           },
+          {
+            opcode: "executeJS",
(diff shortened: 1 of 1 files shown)


LilyMakesThings:
This doesn't really fit the scope of this extension imo.

GarboMuffin:
As iframe exists I don't see any reason to not have a sandboxed JavaScript eval extension but it should be a separate one as iframe is about displaying interactive pages

GarboMuffin:
(this also breaks if the script being executed contains `</script>` anywhere)

PR #760, opened 2023-07-18
Add CORS Proxy docs/extension

NexusKitten:
AN extension to help teach about CORS errors, and hopefully avoid future GitHub issues for them lol.

This is my first time using the documentation API, lmk if I messed anything up.

image
image

# **_Any feedback/upvotes (👍) are apprecieted!_**

PR #1492, opened 2024-05-28
Fetch+

PPPDUD:
Fetch+ is like Fetch, but it has support for a CORS proxy too.


LilyMakesThings:
The problem with having CORS proxies "built in" to extensions like this is that, the nature of the internet and APIs is that they tend to go offline after an amount of time. We need them to live for as long as TurboWarp theoretically will, and the more we depend on external websites like this, the more projects that might break in the future (projects by real people).

One example is "S-Grab", which is now fundamentally pointless because the API that it relied on appears to be down or dysfunctional.

Providing some kind of direct and more explicit way to tell people about CORS errors (or fetching errors in general) would encourage them to use google and find a CORS proxy of their own - that way if it goes down, TurboWarp isn't to blame.

PPPDUD:
As always, if a CORS proxy goes down, a fellow TurboWarp user can always
swap it out in the source code.

Plus, it's increasingly hard to find a good CORS proxy that's not paid
these days.

---
Even if CORS errors did show up, most users would not know that they need
to use a proxy, let alone which one.

In my extension, even if the default proxy doesn't work, you can always use
a fallback proxy with the (fetch [x]) block.

LilyMakesThings:
> As always, if a CORS proxy goes down, a fellow TurboWarp user can always
swap it out in the source code.

There are over 100 extensions in this repository that have only 1 consistent maintainer. Reducing the amount of maintenance on each extension is an absolute must, and submitting an entire extension that can only work if it is maintained consistently is simply an unreasonable expectation.

At least if a user of TurboWarp runs into an issue with the CORS proxy they're using in their own code, they can just change it themselves. But what happens when their code suddenly breaks for no apparent reason? Bug reports.

I don't know, this extension just seems like a fundamental nightmare.

PPPDUD:
Here's my *hopefully final* argument: the CORS proxy is for convenience,
not resilience.

LilyMakesThings:
> Here's my *hopefully final* argument: the CORS proxy is for convenience, not resilience.
Then it's not fit for here, I don't know what to tell you

Xeltalliv:
In addition to concerns listed above, there is another one: a lot of extensions have some kind of "load the thing from URL" blocks, but your current solution only addresses the issue for 1 specific extension.

mybearworld:
corsproxy.io is paid, the free version only works up to a certain number of requests and then doesn't let you make any after that

jabinstech:
> corsproxy.io is paid, the free version only works up to a certain number of requests and then doesn't let you make any after that

nuh uh

mybearworld:
Yes, sorry, that was incorrect. It requires you to create an account at a certain request point, though.

NotFenixio:
According to the contribution guide, shouldn't this just be an update to the original Fetch extension? The normal fetch mode on
this extension is just like the original one but with more boilerplaté and the CORS proxy mode is a copypaste but with a string
concatenation slapped in the url field.

AshimeeAlt:
> According to the contribution guide, shouldn't this just be an update to the original Fetch extension? The normal fetch mode on this extension is just like the original one but with more boilerplaté and the CORS proxy mode is a copypaste but with a string concatenation slapped in the url field.

yes it technically should be an update, but...
aside from that the extension is unnecessary so is the modification, to quote garbo:

> We do need to properly educate people about CORS and why their fetch blocks usually don't work, that much is true

which does not mean just slapping corsproxy as an option, plus what they said above.

SharkPool-SP:
> > According to the contribution guide, shouldn't this just be an update to the original Fetch extension? The normal fetch mode on this extension is just like the original one but with more boilerplaté and the CORS proxy mode is a copypaste but with a string concatenation slapped in the url field.
> 
> 
> 
> yes it technically should be an update, but...
> 
> aside from that the extension is unnecessary so is the modification, to quote garbo:
> 
> 
> 
> > We do need to properly educate people about CORS and why their fetch blocks usually don't work, that much is true
> 
> 
> 
> which does not mean just slapping corsproxy as an option, plus what they said above.

It's like making an extension that takes the join block, and making it join a string and "!" to it

namelessisbackbutnottodobadthingsiswear:
> [This extension shouldn't be merged.](https://github.com/TurboWarp/extensions/pull/1492#issuecomment-2206368378)

I feel like there's still a chance for this extension, despite it's many flaws. For reference, I made a similar PR a while ago (#760) and although it never got merged, some of the fundamental ideas people have brought up in this conversation do carry over from it.

- The extension would need to be more clear about it's use in the title. I agree "Fetch+" sounds a little misleading, and as this extension should act as a tutorial for CORS proxy, having it in the title would help.
- Adding documentation to explain the purpose of CORS proxies and how to use them without the extension would be a necessity. Considering we cannot count on external APIs to be up all the time, and the end goal is to teach people why they need a CORS proxy, having some form of textual explanation to link to would help provide the extension more purpose.
- Perhaps an option to swap CORS proxy providers would be in order. Of course, there's still the chance all are down at once, however it's exponentially smaller, and would be less work to maintain in the event one does permanently shut down.

LilyMakesThings:
A more appropriate solution would be to have a better error response for `Scratch.fetch` from a CORS error. An extension entirely dedicated to just being a CORS proxy is not a solution when it is a fundamentally flawed concept on a number of levels; allowing users to educate themselves on CORS proxies, however...

RockyTheProtogen:
Just gave it a shot. And I think there are tons of valid points.
But the main one *don't hard-code* the proxy. You might like a default, but don't lock it in place.
Adding an argument might be a better option.
Also, change the name to something like `CORS Fetch` which would be less misleading.

PR #1495, opened 2024-05-29
Proposal to Modify Existing Extensions in Violation of the Guidelines (PMEEVG)

PPPDUD:
Name says it all.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,15 +22,14 @@ Strictly, nothing is banned, but the following are *highly* discouraged:
  - Extensions whose primary purpose is monetization (not in the spirit of an open source project)
  - Joke extensions (they aren't funny when they cause us to get bug reports)
 
-Some extensions were added before these rules existed. That doesn't mean you will be exempted too.
-
+Some extensions were added before these rules existed. Such extensions should be modified to be compliant in a pull request.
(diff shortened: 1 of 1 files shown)


LilyMakesThings:
I don't exactly understand what you mean by "modify", but the reason existing extensions that were passed before the guidelines were changed have remained in-tact is due to backwards compatibility.

Xeltalliv:
It seems like your text editor, storage or ram have issues as after your editing, some random sections of text got duplicated and overwrote other random sections of text.

As for suggestion itself, it doesn't make much sense as the extensions that could be fixed were already fixed. The ones remaining are impossible to fix. The only way to make "McUtils" which was intentionally designed as a joke comply with "no joke extensions" rule is to remove it. Same goes for duplicate functionality and broad utilities extensions - they can only be fixed by either removing them or some of their features. But nothing can be removed for compatibility reasons.

LilyMakesThings:
Even just hiding the extensions or delisting them can cause problems, for months after LMS Utilities was delisted Garbo and I received comments on our profile asking where it was, which is the only reason why it's still up. It's really hard to "remove" extensions or features from extensions.

GarboMuffin:
If you think you can modify them into compliance then you are welcome to try

PR #1497, opened 2024-05-30
lab/text: add with/without word wrap option to # of lines block

AshimeeAlt:
Self explanatory (the translations need to be made + updated sorry)
image

This change allows us to circumvent having to reverse the math and do some stupid stuff.
image


GarboMuffin:
have to make a new block and hide the old one to avoid making existing  projects weird

AshimeeAlt:
> have to make a new block and hide the old one to avoid making existing projects weird

oki doki

PR #1498, opened 2024-05-31
Translate mobilekeyboard menu defaultValue

veggiecan0419:
Maybe this should be done for some other extensions too?


veggiecan0419:
@GarboMuffin

GarboMuffin:
You shouldn't translate menu values

GarboMuffin:
they are already translated so you dont need to do anything else

https://github.com/TurboWarp/extensions/blob/084034a6989e6baced5ba2811e2d68148d76a068/extensions/veggiecan/mobilekeyboard.js#L146-L169

if anything the defaultValue should be removed

PR #1501, opened 2024-06-01
obviousAlexC/penPlus: fix opening a project from project_url and the costume editor not having a editing target to attach to

David-Orangemoon:
yeah


David-Orangemoon:
It has no relation to the costume editor though.

GarboMuffin:
you put that in the title, i'm mostly just rubber-stamping these

PR #1509, opened 2024-06-04
New extension: Vibration

PwLDev:
Controls Navigator Vibration API for supported devices.
It will only have effect on mobile devices and other that have the feature, else it will have no effect.
image


AshimeeAlt:
you need to check for navigator.vibrate and add a "can vibrate" block :|

qxsck:
cool

DNin01:
I have wanted to suggest this for a while. Nice to see this feature make it in! 🚀

Issue #1508, opened 2024-06-03
pen v7 breaks my project


penta-quark-neutro:
project designed with pen v6.
The v7 update causes the triangles to not be drawn.
with v6
![Captura de pantalla (472)](https://github.com/TurboWarp/extensions/assets/134885702/01b7001d-4bdc-4da9-9f16-e46a51b718d4)
with v7
![Captura de pantalla (473)](https://github.com/TurboWarp/extensions/assets/134885702/598424c0-dd02-48a5-bedf-5254c9b77b81)

GarboMuffin:
@David-Orangemoon

David-Orangemoon:
Ill look into it

David-Orangemoon:
@penta-quark-neutro could you send the project file?

penta-quark-neutro:
here the project. 
Several others rely on this to function so it is important.
[Motor 3D.zip](https://github.com/user-attachments/files/15542511/Motor.3D.zip)

PR #1510, opened 2024-06-04
obviousAlexC/penPlus: Fix texture handling between versions of pen+ by backporting from the experimental version.

David-Orangemoon:
this fixes one half of https://github.com/TurboWarp/extensions/issues/1508 the other half is from the blocks making the depth being negative.
Thanks to [penta-quark-neutro](https://github.com/penta-quark-neutro) for reporting this bug.
I have migrated the costume handling from the experimental version


David-Orangemoon:
block_6_4_2024-1 20 15 PM
Here @penta-quark-neutro here is the other half for your project.

PR #1516, opened 2024-06-06
obviousAlexC/penPlus: don't use isPackaged

GarboMuffin:
@David-Orangemoon isPackaged can be false in the packager, is there anything wrong with doing this instead?


David-Orangemoon:
That works

PR #1517, opened 2024-06-09
Preserve SVG rotation centers in exported costume URLs across several extensions

### LINKED ISSUE #1499: When adding a costume in Asset Manger, the costume appears as if it's centered despite the content saying otherwise
This title's confusing, so hopefully explaining helps. If you add a costume using asset manager, it will always be centered. This is annoying, but can probably be worked around somehow using invisible squares in the svg data. The main thing that bugs me is that the content of the costume still says it's translated by whatever the input data was. 
This makes very little sense unless you see it in program, so I created a mockup explanation in a project. Just open it in the actual editor and read the comments while following along and you'll see exactly what I mean. 
[https://github.com/snowboyz0825/snowscodestuff/blob/main/SB3s/Bug%20example%201.sb3](https://github.com/snowboyz0825/snowscodestuff/blob/main/SB3s/Bug%20example%201.sb3)
If anything I said here is objectively wrong I'm sorry, all I know about how scratch uses svgs is through reverse engineering them, and through all my experiments I have been able to figure out how they work and manipulate them to get the svg I would want. All I know is this is an issue that bugs me personally.

### LINKED ISSUE #1963: LilyMakeThings • looks+ Extension
When changing an svg color, it resets its x and y offset and changes its rotation point. I could manually fix the x and y offset but i can't solve the rotation point problem.

I need it because i suppose there is way to make a sprite change color AND brightness. 
For my character, to make its skin customizable, i created (for each body segment) 1 layer for the outline (back layer) that is black, 1 for the skin (red by default then the color can change depending on the skin tone, and brightness effect can change) and 1 other for the skin brightness (black).

skin and its outline, with the 3 layers (1og and 2 clones. It may slow the game a lot when i'll add the 7 others athletes (21 sprites for each, and 42 clones (even more clones if not naked...)))
Image

for example, the arm new position when i use the extension to change its color is this : 
Image

GarboMuffin:
Fixes https://github.com/TurboWarp/extensions/issues/1499

Fixes https://github.com/TurboWarp/extensions/issues/1963

### DIFF
--- a/extensions/Lily/Assets.js
+++ b/extensions/Lily/Assets.js
@@ -641,7 +641,7 @@
 
       const costume = target.sprite.costumes[costumeIndex];
       switch (attribute) {
-        case "dataURI": { 
+        case "dataURI": {
           if (!requireNonPackagedRuntime("dataURI of costume")) return "";
           const base64 = vm.getExportedCostumeBase64(costume);
           return `data:${costume.asset.assetType.contentType};base64,${base64}`;
--- a/extensions/Lily/Assets.js
+++ b/extensions/Lily/Assets.js
@@ -641,9 +641,11 @@
 
       const costume = target.sprite.costumes[costumeIndex];
       switch (attribute) {
-        case "dataURI":
+        case "dataURI": { 
           if (!requireNonPackagedRuntime("dataURI of costume")) return "";
-          return costume.asset.encodeDataURI();
+          const base64 = vm.getExportedCostumeBase64(costume);
+          return `data:${costume.asset.assetType.contentType};base64,${base64}`;
+        }
         case "index":
           return costumeIndex + 1;
         case "format":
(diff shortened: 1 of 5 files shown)


SharkPool-SP:
quick note, Skins also suffers the same centering issue. This seems to be a renderer issue if you don't already know

SharkPool-SP:
> This would fix #1499 but it won't be merged yet, until it works on more than just vectors

editors note I LIED and im stoopid. Not a renderer issue, its just not adding a rotation center when making the skin. Defaulting to 0,0. Most reasonable approach is adding a block to manually set the rot center of a skin/a block to toggle auto-centering

SharkPool-SP:
Costume 1:
image

Costume 2 is the same but centered.

Rotation center doesnt seem to be read:
image
image

Issue #1079, opened 2023-10-04
Various Extension Images Changed


xTvii:
I've noticed that a _lot_ of the images in extensions don't really follow the "scratch extension style" which is what TurboWarp follows as well.

I've remade all of them to fit the style better. I hope you enjoy them!

**Fetch**
![Fetch(2)](https://github.com/TurboWarp/extensions/assets/137563418/8d700b36-4c5f-4152-8fba-297606066771)
**IFrame**
![IFrame](https://github.com/TurboWarp/extensions/assets/137563418/27e929bb-924b-4f05-9f1f-87702c34c236)
**Clipboard**
![Clipboard](https://github.com/TurboWarp/extensions/assets/137563418/0c633237-2e24-4b1c-b399-265f535e0ca1)
**Pen Plus**
![Pen Plus](https://github.com/TurboWarp/extensions/assets/137563418/f21275a2-f07e-449f-8eaf-8717dd244803)
**Sensing Plus**
![Sensing Plus](https://github.com/TurboWarp/extensions/assets/137563418/2f998b0d-3860-4943-a69d-9769ccb5ab72)
**Browser Fullscreen**
![Browser Fullscreen](https://github.com/TurboWarp/extensions/assets/137563418/91b1d061-c2be-48c3-82ad-d50fc4a31449)
**Screen Resolution**
![Screen Resolution](https://github.com/TurboWarp/extensions/assets/137563418/9cf6dc6f-31d6-470b-a64c-d945b90b5624)
**Deltatime**
![Deltatime](https://github.com/TurboWarp/extensions/assets/137563418/aa082120-2211-45fa-97ae-0ed08977772e)
**All Menus**
![All Menus](https://github.com/TurboWarp/extensions/assets/137563418/f31e8486-319b-4805-91af-ea538ce5e3b5)
**JSON**
Font used is _Gotham_
![JSON](https://github.com/TurboWarp/extensions/assets/137563418/b140d041-7ade-4b78-9ae1-73489fd311ea)
**Numerical Encoding**
Font used is _Gotham_
![Numerical Encoding](https://github.com/TurboWarp/extensions/assets/137563418/9f2444cc-1b99-4799-bd1d-014144660e63)
**S-Grab**
![S-Grab](https://github.com/TurboWarp/extensions/assets/137563418/0694fdcd-bc19-42e3-95b9-2fdf6a9138f0)
**Graphics 2D**
[...]

PR #1518, opened 2024-06-09
Image update batch

GarboMuffin:
About 8 months late

Various things from https://github.com/TurboWarp/extensions/issues/90 and https://github.com/TurboWarp/extensions/issues/1079


SharkPool-SP:
Im gonna miss the old fetch

SharkPool-SP:
also you forgot about key simulation

GarboMuffin:
@veggiecan0419 instead of just thumbs down reacting, can you elaborate which images you think are worse?

veggiecan0419:
@GarboMuffin Actually, I preffered how most of them were before.

With sensing+, I feel the old image better represented the extension (as it isn't only touch blocks).
As for fetch, I liked the old concept (the dog **fetch**ing the bone).

As for browser fullscreen, I do have to admit it was a 30min job, but I don't think it looked that bad, just the colours were a bit dull (and it wasn't SVG).

The only ones I really feel got an improvement were deltatime and all menus.

As for all the others, I think that they were good, because they gave the gallery a diverse feel, which I quite liked (I feel now it looks pretty monotone and repetitive).

GarboMuffin:
Thank you for elaborating

Let me explain the motivation for why those images were due to get changed:

 - Many contained the name of the extension. This is redundant and can't be translated.
 - The old sensing+ has trademarks that we definitely should not be using in that context (griffpatch has gotten letters about his use of the Scratch Cat).
 - SVG is easier for other people to edit and looks better. Usually smaller too (old browser fullscreen 58KB, new 3KB)
 - There is a general style that the images should try to follow, but there is still plenty of room for the images to look unique within that style.
 - 2 of the extensions didn't even have images before

If you have better versions that don't regress the problems above, you can submit them and they will get looked at in less than 8 months this time

GarboMuffin:
Any change that wasn't one of those bullet points, like the dog fetching metaphor being replaced, was largely coincidental and can be reverted if someone submits a better one

veggiecan0419:
@GarboMuffin 👍

xTvii:
Sorry if soem of the ones I submitted were a bit of a downgrade !!! ( I made the new fetch and sensing + ones . I also made quite some more if you check #1079 . I'm going to try and make some more too !!! I hope they're better !!

xTvii:
HHOYAH ! i present to you , garbo , my latest invention !!! THE #1607 INATOR THREE THOUSAND !!! YYAAAAAYA more icons is always a good thing especially when like 4 million ( at least 3 of them ) of them have text

PR #1522, opened 2024-06-09
PwLDev/vibration: Vibration pattern block

DNin01:
The `navigator.vibrate()` function isn't limited to just buzzing your phone once for a specific amount of time. You can actually use it to play a pattern where the device alternates between vibrating and not vibrating. You can learn more [on MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/API/Vibration_API#vibration_patterns).

I added a block to bring this functionality to the vibration extension.

Tested on Edge for Android.

### DIFF
--- a/extensions/PwLDev/vibration.js
+++ b/extensions/PwLDev/vibration.js
@@ -67,7 +67,7 @@
           // Change to array split by ','
           .split(",")
           // Change to numbers in milliseconds
-          .map((val) => Scratch.Cast.toNumber(val.trim()) * 1000);
+          .map((val) => Scratch.Cast.toNumber(val) * 1000);
         navigator.vibrate(pattern);
       }
     }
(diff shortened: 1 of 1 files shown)


LilyMakesThings:
Given that JSON is already sort of an accepted format for TurboWarp extensions, I don't think using strings with a delimiter is the right thing to do.

CubesterYT:
Yeah, I agree. We can easily just use a normal Array here.

GarboMuffin:
if you do something like `.replace(/[^0-9\-.]/g, "")` then people can give the comma list or a JSON list or whatever they want and it should just work

DNin01:
I changed `.split(",")` to `.match(/[\w\-.]+/g)`, which handles inputs like this:
![RegEx match results of '[1, "2", -3.5 , Infinity, ,  3e4, 0x06, 6-.6.2, NaN]': 1, 2, -3.5, Infinity, 3e4, 0x06, 6-.6.2, NaN](https://github.com/TurboWarp/extensions/assets/106490990/f6cdd579-6167-4e94-b459-3f818bf0c570)

PR #1527, opened 2024-06-10
[new extension] Big Decimal

qxsck:
High precision operation, support decimals.


mcnole25:
Why does 1/3*3 equal 0.9999999999... rather than 1?

CST1229:
> Why does 1/3*3 equal 0.9999999999... rather than 1?

Because division only has 100 digits of decimal precision.

PPPDUD:
Seems a little similar to BigInts, perhaps consider adding to that extension instead?

qxsck:
> 为什么 1/3*3 等于 0.9999999999...而不是 1？

maybe i can add round?

Skyhigh173:
I suggest the extension name to be "precise numbers" since "Big Number" is commonly referred as number greater than 1.79e308

qxsck:
> I suggest the extension name to be "precise numbers" since "Big Number" is commonly referred as number greater than 1.79e308

Big Number is relative to BigInt, one is an integer and the other is a number that includes decimals

qxsck:
> Why does 1/3*3 equal 0.9999999999... rather than 1?

image

qxsck:
some new blocks:
image
set precision

image
math including abs,ceil,floor,trunc,max,min,round,random

FurryR:
imo it should name `Big Decimal`

qxsck:
> imo it should name `Big Decimal`

https://github.com/TurboWarp/extensions/pull/1527#issuecomment-2180956131
i think Big Number is suitable

GarboMuffin:
Big Decimal would clarify difference from Big Int

qxsck:
> Big Decimal would clarify difference from Big Int

maybe
If you think bignumber is easily confused with bignumber.js, and bigdecimal is more suitable, then it should indeed be changed, but I don't want to change the extension ID but only the name

qxsck:
> Big Decimal would clarify difference from Big Int

also, can I ask when merge it?

qxsck:
> > Big Decimal would clarify difference from Big Int
> 
> maybe if you think bignumber is easily confused with bignumber.js, and bigdecimal is more suitable, then it should indeed be changed, but I don't want to change the extension ID but only the name

LilyMakesThings:
> > Big Decimal would clarify difference from Big Int
> 
> also, can I ask when merge it?

On a day that doesn't include the letter "A"

CubesterYT:
So, on a Tuesnight.

qxsck:
> > > Big Decimal would clarify difference from Big Int
> > 
> > 
> > also, can I ask when merge it?
> 
> On a day that doesn't include the letter "A"

what

fath11:
> On a day that doesn't include the letter "A"

does changing your language counts?

qxsck:
big-decimal
image

FurryR:
You reimplemented [big.js](https://www.npmjs.com/package/big.js) or [bigdecimal](https://www.npmjs.com/package/bigdecimal). Consider using existed libraries instead.

Ender-Studio:
exponent with decimals, pick random and a few more doesn't seem to work as intended it gives stuff far from the exact one. One exampls is 25^0.5 it returns 1 instead of 5

qxsck:
forever.

Brackets-Coder:
> forever.

Doesn't seem like it's been closed forever

PR #1546, opened 2024-06-16
ProjectCover

0832k12:
Make a link.

### DIFF
--- a/extensions/0832/projectcover.js
+++ b/extensions/0832/projectcover.js
@@ -38,7 +38,7 @@
       this.storage.URL = prompt('URL: ');
     }
     load() {
-      fetch(this.storage.URL)
+      Scratch.fetch(this.storage.URL)
         .then((r) => r.arrayBuffer())
         .then((buffer) => Scratch.vm.loadProject(buffer))
         .then(() => { })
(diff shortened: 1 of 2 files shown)


SharkPool-SP:
What

Drago-Cuven:
what does it do? what blocks are there? why did you make it? WE NEED INFO PLEASE!

CubesterYT:
I'm guessing this extension would be used in a case scenario where the project is just a thumbnail, and also has a bonus point of having very little file size (initially), and then imports the actual project from the link provided.

0832k12:
> I'm guessing this extension would be used in a case scenario where the project is just a thumbnail, and also has a bonus point of having very little file size (initially), and then imports the actual project from the link provided.

Yes.

0832k12:
image

SharkPool-SP:
> I'm guessing this extension would be used in a case scenario where the project is just a thumbnail, and also has a bonus point of having very little file size (initially), and then imports the actual project from the link provided.

couldn't you just use asset manager

RockyTheProtogen:
I can only say one thing... Why?

PR #1555, opened 2024-06-21
Security Key

XmerStudio:
Authenticate Users on the Systems/Applications you make available to them or make them unlock the data they store in their Storages/Servers with a key that only they can access. Requires Security Key (Hardware).


Thebloxers998 (commented):
Nice

GarboMuffin:
Just so you're aware, reviews from non-organization members could sometimes be valuable but do not count towards the 2 review requirement, and this pull request is closed anyways so here it especially is just spam.

The notable problem with this is extension is that despite being called "Security Key" it does not actually provide any security as it is client-side-only

Thebloxers998:
Oh

PR #706, opened 2023-07-08
TurboHook V.1.1.0 Update

CubesterYT:
Hello again! I'm putting up the PR again because I did some fatal merges to the master branch of my fork, causing me to have to delete my entire fork. So using my backups, I remade the PR. So let me reiterate the last PR:

The moment everyone's been waiting for... TurboHook V.1.1.0!!!

This update includes full embed support, JSON support, bugfixes, typo fixes, some semi colons, extra space removals, and so much more.

Here is the current roster:
image

PR #1568, opened 2024-06-25
TurboHookV2 which supports discord's embeds

fcures:
It tested it beforehand to make sure it all works correctly.
It sends embeds data like title, timestamp, color, images, links, thumbnail, footer, and description.


CubesterYT:
There's no need for this PR. I made a PR that already did this, but I closed it because I decided the current version of TurboHook is too messy have a rewritten version that loads better that I plan to PR soon.

CubesterYT:
This was the update: https://github.com/TurboWarp/extensions/pull/706
I closed it because I rewrote the entire extension from the ground up, that makes it a fully capable normal multi purpose WebHook extension that'll have a lot of useful functionality. So I request this be closed, I don't need the bad code to be expanded on.

fcures:
> This was the update: #706 I closed it because I rewrote the entire extension from the ground up, that makes it a fully capable normal multi purpose WebHook extension that'll have a lot of useful functionality. So I request this be closed, I don't want the bad code to be expanded on.

ok cool, thanks, also how the hell do I use prettier 💀💀💀😭 also where is the better version

CubesterYT:
@fcures you have to either have to install the deck server and command line use prettier using `npm run format`, or use the prettier VSCode extension

CubesterYT:
The better version still needs one more thing till release

fcures:
> The better version still needs one more thing till release

Alright cool

PR #1576, opened 2024-06-29
Resubmit "Trampoline" Extension

namelessisbackbutnottodobadthingsiswear:
Do forgive any git/syntax errors, it's been a while since I've last submitted a pull request here so I'm not exactly up to date on all the new APIs.

I've decided to resubmit this extension to **take the place of S-Grab** due to the closure of ScratchDB (for the foreseeable future). While @GarboMuffin already did some patchwork on the extension in #1511, I have still received several complaints about the extension despite it's warnings.
Additionally, I don't see a point to withhold the extension now that ScratchDB support is offline for S-Grab. They do the same thing, except this one is more well-rounded and will ultimately have less complaints on my profile and in the issues tab.

**Please let me know of any errors** you may find and I'll try my best to fix them.

_PS. I do hope that this pull request stays up despite my previous misconduct in this repository, I have made this account to post this extension and likely abandon it after the pull request is merged._

### DIFF
--- a/extensions/NexusKitten/trampoline.js
+++ b/extensions/NexusKitten/trampoline.js
@@ -5,288 +5,308 @@
 // License: MIT
 
 (function (Scratch) {
-  'use strict';
+  "use strict";
 
   if (!Scratch.extensions.unsandboxed) {
-    throw new Error('Trampoline must run unsandboxed');
+    throw new Error("Trampoline must run unsandboxed");
   }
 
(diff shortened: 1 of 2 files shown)


CubesterYT:
Though we usually don't want extensions that require a CORS proxy to operate on the gallery, as we can't depend on it if it goes down temporarily or for a long time till we replace it, I think we can(?) make an exception for this one. It'll be up to GarboMuffin, though.

LilyMakesThings:
based username

> Though we usually don't want extensions that require a CORS proxy to operate on the gallery, as we can't depend on it if it goes down temporarily or for a long time till we replace it, I think we can(?) make an exception for this one. It'll be up to GarboMuffin, though.

This is the problem, S-Grab going down because of ScratchDB was an inevitability, but a necessary evil given the feature-set of the api. In this case I think it could be preventable, and it certainly isn't entirely dependent on it.

veggiecan0419:
@namelessisbackbutnottodobadthingsiswear So are you @NexusKitten?

CubesterYT:
> @namelessisbackbutnottodobadthingsiswear So are you @NexusKitten?

Yes, they are, but let's not dwell on it.

veggiecan0419:
namelessisbackbutnottodobadthingsiswear = `Nameless (Cat) is back but not to do bad things I swear`? 🤓

rkxspace (changes requested):
I suggest making the CORS proxy configurable. That way people can use their own, or some other CORS proxy on the internet.
If it is, my fault, I just didn't see a value to change it.

extensions/NexusKitten/trampoline.js
yuri-kiss, line 1: I think this would work better if TurboWarp added proxy's for these api's itself so that it does not have to rely on something like corsproxy just to work, as if turbowarp is the failing point then it can most likely be attributed to something like a school blocker etc.

NexusKitten: in that case, would it be best for me to draft/close this extension for now? there's not really a good way to get around using a CORS proxy here, and the option to make it configurable (https://github.com/TurboWarp/extensions/pull/1576#pullrequestreview-2166337388) feels like a messy solution to that problem.

yuri-kiss: yeah mark it as draft

yuri-kiss: or close it, whatever you feel is best

Issue #1587, opened 2024-07-04
Proposal: dependency system based on `await import`


FurryR:
Extensions are now able to require external libraries via `await import`. When the gallery starts, it downloads all dependencies (via `acorn` parser) and replace their URLs with downloaded data URLs.

Example code:

```js
(async function (Scratch) {
  // Dynamic URLs are not supported.
  const editor = await import('https://cdn.jsdelivr.com/...');
})(Scratch);
```

After processing:

```js
(async function (Scratch) {
  const editor = await import('data:text/javascript;base64,...');
})(Scratch);
```

There will be a `cache` directory to store downloaded dependencies so that you do not need to download them twice.

namelessisbackbutnottodobadthingsiswear:
The reason external libraries aren't allowed is because we cannot be sure whether the linked URL will always be up or not. For extensions that primarily utilize `Scratch.fetch`, this is a necessary evil, however it's just easier to paste in a DataURI for any needed libraries to ensure it's always online when the extension is.

FurryR:
Implemented by #1591

FurryR:
> however it's just easier to paste in a DataURI for any needed libraries to ensure it's always online when the extension is.

This proposal is for automatically converting await import URLs into data URLs when building so it makes reviewers easier to ensure the integrity of dependencies.

LilyMakesThings:
> The reason external libraries aren't allowed is because we cannot be sure whether the linked URL will always be up or not.

This proposes that as part of the build process it will automatically convert the linked library to JS, since you're 100% right that we can't always rely on 3rd-party dependencies on such critical parts of projects.

In the grand scheme of things, this won't affect the front-end and will just make development a bit more readable.

PR #1591, opened 2024-07-04
All you need to slightly improve experience

FurryR:
# Changes

1. refactor development kits with ESM (better than #1564), requires extra ESM adaptation for `desktop` (by... renaming `prepare-extensions.js` to `prepare-extensions.mjs` and use `await import`.)
2. make everything asynchronous
3. implement #1587 via esquery


Drago-Cuven:
upvoted

PR #1212, opened 2023-12-22
Skyhigh173/JSON: v2.7

Skyhigh173:
# misc change
- used single quotes
- renamed a block
- optimized get all keys, values, entries
- ` {"key":123}  ` are considered valid json now (leading & trailing spaces)
# blocks added
- minify json
- filter list based on text
- flip key-value pair
- import all variables (**works for clones & local & global vars**)
- export all variables (**works for clones & local & global vars**)
- show error message
image

PR #1596, opened 2024-07-08
Skyhigh173/JSON: fix json_is_valid()

Procybit:
The function returns false when valid checked JSON is not whitespace-trimmed. Made it check parsed type.

### DIFF
--- a/extensions/Skyhigh173/json.js
+++ b/extensions/Skyhigh173/json.js
@@ -597,20 +597,13 @@
     }
 
     json_is_valid({ json }) {
-      if (typeof json != "string") {
-        return false;
-      } else if (
-        (json.slice(0, 1) != "[" || json.slice(-1) != "]") &&
-        (json.slice(0, 1) != "{" || json.slice(-1) != "}")
-      ) {
-        return false;
-      } else {
-        try {
-          JSON.parse(json);
-          return true;
-        } catch {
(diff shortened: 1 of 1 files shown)


RockyTheProtogen (commented):
Code no longer returns false when whitespace is introduced.
Fulfils PR but could break existing projects.

CubesterYT:
I don't think it'll break projects considering projects would've already accounted for this bug and used non whitespace valid json

RockyTheProtogen (approved):
After a bit of thinking, this shouldn't really affect anything unless projects, for some reason, use a rudimentary way of reading the JSON. Though that is doubted, as the get key block is in this extension.

GarboMuffin:
this changes the output of

image

what do you think that should output?

`true` `false` `0` `"hello"` are all valid JSON too, should those not return true here as well?

regardless this will need to be a is_valid_v2 kind of situation

LSPECTRONIZTAR:
uh this problem is already addressed in #1212

Skyhigh173:
Just trim trailing and leading white spaces and test it with original script

Procybit:
So for now it just trims given JSON string before processing it

GarboMuffin:
okay, i think these blocks need to be replaced but this should be an unambiguous improvement for now

Issue #1166, opened 2023-11-26
why can't run extensions


qxsck:
![image](https://github.com/TurboWarp/extensions/assets/118588391/809bd1fe-7659-4487-b399-c619f15877f6)

such as encoding , notifications.

GarboMuffin:
Encoding's Chinese localization has a bug, will be fixed shortly

Notifications hasn't been touched for a month, what error does it give you?

qxsck:
> 编码的中文本地化有一个bug，将很快修复
> 
> 通知已经一个月没有触碰了，它会给你什么错误？

like this

PR #1599, opened 2024-07-09
encoding: Fix unicode base64 encode/decode

Procybit:
Fixes https://github.com/TurboWarp/extensions/issues/1166

Previously, encodings outside the Latin range were not supported.
[Base64 - "The Unicode Problem"](https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem)


yuri-kiss (approved):
Seems good to me as well.

snowboyz0825:
Ironically I'm the one that asked to have encoding changed, I knew this would be an issue eventually

CubesterYT:
Fixes https://github.com/TurboWarp/extensions/issues/1166

PR #1432, opened 2024-05-01
New extension: Multi Touch

Skyhigh173:
Multi-touch extension (to replace sensing+'s)

image

x: x position
y: y position
dx: change of x position compared to previous frame
dy: change of y position compared to previous frame
sx: avg speed x of finger in a second
sy: avg speed y of finger in a second
duration: time since finger press
force (some devices only): force of finger press

Tested on:
- iOS device
- android device

Not tested on:
- laptop with touchscreen

PR #1608, opened 2024-07-13
New extension: Touches

DNin01:
This may not be final yet, but I've been wanting to share it for a little while. It's a multi-touch extension.

In this extension, I tried to add blocks that require minimal coding to implement multi-touch. Here's a picture that shows the Scratch versus Touches extension equivalents to a few scripts:
Scratch blocks and Touches extension equivs side-by-side

Here is the full palette:
Touches extension block palette

There is a learning curve since the blocks are a bit complicated, but once you understand the extension, it does a lot of the work for you and behaves quite similarly to Scratch's own mouse sensing blocks which you're surely very familiar with.

I also made a sample project to go with it. You can [download and try it](https://github.com/user-attachments/files/16203631/Touch.Game.zip) or [watch the video](https://github.com/user-attachments/assets/61a93b2d-1f78-465c-88ea-d0a3d6bcd58b). My plan is to publish the sample project via a separate PR if and once this PR is merged.

This extension uses the [Pointer events web API](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events), so it accepts all types of pointer input events, including mouse, touch, and even pen input. I have tested it on Chromium on desktop and mobile, plus Safari on mobile.

In a way, this extension is a superset of Scratch's mouse and click sensing blocks. They both work very similarly, it's just that one can deal with multiple pointers at a time.

<details><summary>Technical notes</summary>

- **Pointer IDs**
  
  When a finger touches the screen, it is assigned a unique numerical ID. Touch IDs start at 1 and each new touch will get an ID that isn't assigned to an active touch.
  
  For example, if two fingers touch the screen one after another (or at the same time), one will have an ID of 1, and the other, an ID of 2. If the first finger is then lifted, its "slot" of ID 1 is freed up for the next touch.
- **Ghost slots** 
  
  When a finger is released, the ID of its touch is actually reserved for one extra frame. This gives code time to respond to that touch being released; if a new touch was assigned an ID of a touch that had ***just*** been released, code would just think the finger had moved.
  
  Instead, these touches will be assigned a different ID, which prevents problems like dragged objects jumping from one finger to another. However, this means you might have touches with an ID greater than the number of current or maximum active touches (e.g. more than 10). These reserved IDs, however, don't count to the total number of current active touches, and aren't considered to be "pressing" anything.
- **Non-touch devices**
  
  If the user's device or browser doesn't support touch input, the `(max touch points)` block will return 1, because the mouse is a pointer.

</details>

<details><summary>Why this extension?</summary>

The fact that I made a sample project demonstrates that it really is practical to make projects with this extension once you get used to it. Compared to simpler touch sensing extensions, this one does a lot of the work for you — tracking touched sprites, which fingers are touching which sprites, etc. Some of the blocks are familiar to Scratch's own mouse sensing blocks, to help bridge the gap for current coders to add multi-touch support to their projects.

</details>


CubesterYT:
https://github.com/TurboWarp/extensions/pull/1432 is a similar extension.

I like this PR though, matches category styling, and more, comprehensive blocks.

Drago-Cuven:
something I must alert any "touch" based extension about, not every device with a touch screen is a mobile device. there are even windows 11 computers with touch screens. please keep this in mind when creating/editing extensions like this.

DNin01:
> something I must alert any "touch" based extension about, not every device with a touch screen is a mobile device. there are even windows 11 computers with touch screens. please keep this in mind when creating/editing extensions like this.

Thanks.

The `is touch screen?` block was not intended to be a surefire method to determine whether a device would be considered a mobile device. This would be worth pointing out if this extension ever gets documentation.

PR #1618, opened 2024-07-17
Xeltalliv/simple3D: Feature update v1.1.0

Xeltalliv:
image

Adds support for SharkPool's "Layer Control" extension.
Fixes bug where it was always using centroid interpolation for `v_color`.
Fixes [set clear color R:() G:() B:() A:()] to take premultiplied alpha into a count. 
Adds 2 new blocks.
Adds new blending mode for drawing behind what is already drawn.
Adds ability to measure text's horizontal advance distance, which is different from text's bounding box width.
Adds Math.ceil to text borders to prevent text from jumping around when smoothly scaling border.
Adds PenguinMod and Snail IDE extension removal support.
Clarifies which parts are MPL-2.0 and which are BSD-3-Clause.
Changes default near plane to be further to reduce Z-fighting.

### DIFF
--- a/docs/Xeltalliv/simple3D.md
+++ b/docs/Xeltalliv/simple3D.md
@@ -177,7 +177,7 @@ draw to the screen at X,Y. Use Z for depth check.
 And now it's time to make some actual Simple3D scratch blocks code.
 In Simple3D those transformations are combined into 1 big transformation, and all the steps have to be specified in reverse.
 ```scratch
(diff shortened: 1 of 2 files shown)


SharkPool-SP:
wowie

GarboMuffin:
have some opinions about how other people decided to do extension removal but ill leave it in because i dont have a better thing yet

PR #1633, opened 2024-07-29
Add hidden "ArrayBuffer" option (for ZIP)

yuri-kiss:
This adds a hidden arrayBuffer option to the menu which can be enabled by setting "_showArrayBufferOption" to true. (its meant for other extensions to do it)

I am making a buffer extension and wanna PR it, but the current method to add support is 6KB of patch code so I wanted to do it internally.

### DIFF
--- a/extensions/CST1229/zip.js
+++ b/extensions/CST1229/zip.js
@@ -24,6 +24,7 @@
       // for developers who want to integrate their extensions with this one
       // @ts-ignore
       Scratch.vm.runtime.ext_cst1229zip = this;
+      this._showArrayBufferOption = false;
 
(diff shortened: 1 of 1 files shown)


CST1229:
i expected this to be in the same pr as the buffer extension itself

yuri-kiss:
> i expected this to be in the same pr as the buffer extension itself

:melting_face:

CST1229:
but ok

yuri-kiss:
> but ok

I mean I can close this and move it with the buffer ext if you want :I

SharkPool-SP (approved):

PR #1664, opened 2024-08-29
new extension about abbreviation

CoolDevXD:
**NOTE:** i was really bored and i had never really made an extension before or used java script so there may be some problems

### **This extension adds a lot of block to abbreviate numbers**

- it adds blocks to abbreviate and un-abbreviate
- blocks to do simple math functions with the abbreviated numbers like: (1k + 2.3m)
- and blocks to compare abbreviated numbers like: (4.3k > 8.1m)

Thats basically it.

### DIFF
--- a/extensions/Cool_skratcher/abbreviations.js
+++ b/extensions/Cool_skratcher/abbreviations.js
@@ -7,10 +7,6 @@
 (function(Scratch) {
     'use strict';
   
-    if (!Scratch.extensions.unsandboxed) {
-      throw new Error('This Hello World example must run unsandboxed');
-    }
-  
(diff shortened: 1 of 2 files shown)


EmimalayaYT:
What is this for again? Sorry, I don't get it

CoolDevXD:
> What is this for again? Sorry, I don't get it

for abbreviating numbers, so 1000 = 1k and 1000000 = 1m. so you can use larger numbers in your game

yuri-kiss:
I don't understand the use of this when the user can do the same thing, + this could be merged into another extension instead

PR #1673, opened 2024-09-05
text extension small expansion

BludIsAnLemon:


SharkPool-SP (approved):

extensions/text.js
```diff
@@ -613,6 +696,29 @@
           return string;
       }
     }
+    posWith(args) {
+      const STRING = args.STRING.toString();
```
GarboMuffin, line 700: ought to be Scratch.Cast.toString....

PR #1691, opened 2024-09-16
Jack Halloween Pet Extension

snowboyz0825:
Basically just an asset swap with TheShovel's Oneko, but Halloween themed. 

Adds a pet that follows your cursor around in the editor - no blocks.


snowboyz0825:
Ignore the dummy files ._. 
I don't know how to create a folder in a repo without creating a file directly to it

CubesterYT:
We don't add "fun" extensions to this repository, nor do we add joke ones either. Please look over our extension criteria again.

snowboyz0825:
> We don't add "fun" extensions to this repository, nor do we add joke ones either. Please look over our extension criteria again.

To cite Contributing, "Acceptance criteria
Strictly, nothing is banned, but the following are highly discouraged:

Broad "Utilities" extensions (break them up into multiple extensions, see https://github.com/TurboWarp/extensions/issues/674)
Extensions that are very similar to existing ones (consider modifying the existing one instead)
One-use personal extensions (load the extension as a local file instead)
Extensions whose primary purpose is monetization (not in the spirit of an open source project)
Joke extensions (they aren't funny when they cause us to get bug reports)
Some extensions were added before these rules existed. That doesn't mean you will be exempted too."

"Nothing is banned" - it was worth trying to add it, just to add a little holiday cheer
"Joke Extensions" - I don't believe this would classify as a joke extension - it is serious - maybe we could just leave it in for Halloween season, who knows, but I completely understand if it's a definite no. 

(also it failed Immediately because I forgot to add a comma when editing extensions.json)

CubesterYT:
This extension gives no functionality other than having no blocks and something following your cursor, it's not meant for this repository. Also, like the criteria mentions, it can cause a lot of bug reports, asking, "what is wrong", etc.

PR #1694, opened 2024-09-22
sweetalert.js

xmeroriginals:
It allows you to send modern alerts using the Sweet Alert library.


CubesterYT:
This repository doesn't allow having dependencies that import from the Internet. It's required practice that the extension includes an offline version of the dependency that you use.

PR #1699, opened 2024-09-24
RSA Extension

familycomicsstudios:
An extension to use RSA in TurboWarp.

### DIFF
--- a/extensions/Themadpunter/rsa.js
+++ b/extensions/Themadpunter/rsa.js
@@ -1,0 +1,83 @@
+(function(Scratch) {
+    'use strict';
+
+    // Load the JSEncrypt library from a CDN
+    const script = document.createElement('script');
+    script.src = 'https://cdn.jsdelivr.net/npm/jsencrypt/bin/jsencrypt.min.js';
+    document.head.appendChild(script);
+
+    class RSAExtension {
+        getInfo() {
+            return {
+                id: 'themadpunter_rsa',
+                name: 'TurboRSA',
+                color1: '#FF0000', // Primary color (red)
+                color2: '#CC0000', // Secondary color (darker red)
+                blocks: [
+                    {
+                        opcode: 'generateKeys',
+                        blockType: Scratch.BlockType.REPORTER,
+                        text: 'Generate RSA keys',
+                        arguments: {}
+                    },
+                    {
+                        opcode: 'encrypt',
+                        blockType: Scratch.BlockType.REPORTER,
+                        text: 'Encrypt [TEXT] with public key [KEY]',
+                        arguments: {
+                            TEXT: {
+                                type: Scratch.ArgumentType.STRING,
+                                defaultValue: 'Hello, world!'
+                            },
+                            KEY: {
+                                type: Scratch.ArgumentType.STRING,
+                                defaultValue: ''
+                            }
+                        }
+                    },
+                    {
+                        opcode: 'decrypt',
+                        blockType: Scratch.BlockType.REPORTER,
+                        text: 'Decrypt [TEXT] with private key [KEY]',
+                        arguments: {
+                            TEXT: {
+                                type: Scratch.ArgumentType.STRING,
+                                defaultValue: ''
+                            },
+                            KEY: {
+                                type: Scratch.ArgumentType.STRING,
+                                defaultValue: ''
+                            }
+                        }
+                    }
+                ]
+            };
+        }
+
+        generateKeys() {
+            const crypt = new JSEncrypt({ default_key_size: 1024 });
+            const publicKey = crypt.getPublicKeyB64();
+            const privateKey = crypt.getPrivateKeyB64();
+            return JSON.stringify({ publicKey, privateKey });
+        }
+
+        encrypt(args) {
+            const crypt = new JSEncrypt();
+            crypt.setPublicKey(args.KEY);
+            const encrypted = crypt.encrypt(args.TEXT);
+            return encrypted ? encrypted : 'Encryption failed';
+        }
+
+        decrypt(args) {
+            const crypt = new JSEncrypt();
+            crypt.setPrivateKey(args.KEY);
+            const decrypted = crypt.decrypt(args.TEXT);
+            return decrypted ? decrypted : 'Decryption failed';
+        }
+    }
+
+    // Wait for the script to load before registering the extension
+    script.onload = () => {
+        Scratch.extensions.register(new RSAExtension());
+    };
+})(Scratch);
--- a/extensions/Themadpunter/rsa.js
+++ b/extensions/Themadpunter/rsa.js
@@ -1,6 +1,7 @@
 // Name: RSA
 // ID: themadpunter-rsa
-// Description: Encrypt text and files using RSA, the world's best asymmetric encryption algorithm. 
+// Description: Encrypt text and files using RSA, the world's best asymmetric encryption algorithm.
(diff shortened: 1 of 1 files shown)


extensions/Themadpunter/rsa.js
```diff
@@ -0,0 +1,89 @@
...
+// By: Themadpunter <https://scratch.mit.edu/users/The_Mad_Punter/>
+// License: MIT
+
+(function(Scratch) {
+    'use strict';
+
+    // Load the JSEncrypt library from a CDN
+    const script = document.createElement('script');
+    script.src = 'https://cdn.jsdelivr.net/npm/jsencrypt/bin/jsencrypt.min.js';
+    document.head.appendChild(script);
```
SharkPool-SP, line 13: You shouldnt be fetching a file url like that as:
- this forces people to need internet for this extension to work
- url could go down at any point

You should use a data.uri instead.
Also add `/* global [library name here] */` to fix the lint errors

extensions/Themadpunter/rsa.js
```diff
@@ -0,0 +1,89 @@
...
+                id: 'themadpunter_rsa',
+                name: 'TurboRSA',
+                color1: '#FF0000', // Primary color (red)
+                color2: '#CC0000', // Secondary color (darker red)
+                blocks: [
+                    {
+                        opcode: 'generateKeys',
+                        blockType: Scratch.BlockType.REPORTER,
+                        text: 'Generate RSA keys',
+                        arguments: {}
```
SharkPool-SP, line 27: unnecessary key

SharkPool-SP (changes requested):

extensions/Themadpunter/rsa.js
yuri-kiss: Ok being real here, jsdelivr isnt going down any time soon, and no they should find a way to import the library without using a script tag.

CST1229: the desktop app still needs to work without internet (this is why the extension library there can only update when the app does, because it's stored offline).
for importing, i think using `await import()` and making the extension function async is a simple option, but you should still make the extension url a data url because of the offline part

yuri-kiss: maybe yeah but that does bring the question is it allowed because stuff like a script tag isnt allowed and they do similar things

extensions/Themadpunter/rsa.js
```diff
@@ -0,0 +1,89 @@
+// Name: RSA
+// ID: themadpunter-rsa
```
yuri-kiss, line 2: this id should be the same as the extension ID

extensions/Themadpunter/rsa.js
```diff
@@ -0,0 +1,89 @@
...
+
+    // Load the JSEncrypt library from a CDN
+    const script = document.createElement('script');
+    script.src = 'https://cdn.jsdelivr.net/npm/jsencrypt/bin/jsencrypt.min.js';
+    document.head.appendChild(script);
+
+    class RSAExtension {
+        getInfo() {
+            return {
+                id: 'themadpunter_rsa',
```
yuri-kiss, line 18: `_`'s should not be used in ids

extensions/Themadpunter/rsa.js
```diff
@@ -0,0 +1,89 @@
...
+                name: 'TurboRSA',
```
yuri-kiss, line 19: this name should be the same as the one listed on the gallery which you put as `RSA`

extensions/Themadpunter/rsa.js
```diff
@@ -0,0 +1,89 @@
...
+                color1: '#FF0000', // Primary color (red)
+                color2: '#CC0000', // Secondary color (darker red)
+                blocks: [
+                    {
+                        opcode: 'generateKeys',
+                        blockType: Scratch.BlockType.REPORTER,
+                        text: 'Generate RSA keys',
```
yuri-kiss, line 26: nitpick: please make the block titles lowercase so "generate RSA keys" as currently it does not follow scratch's conventions, also blocks are not sentence's so yeah

extensions/Themadpunter/rsa.js
```diff
@@ -0,0 +1,89 @@
...
+                            KEY: {
+                                type: Scratch.ArgumentType.STRING,
+                                defaultValue: ''
+                            }
+                        }
+                    },
+                    {
+                        opcode: 'decrypt',
+                        blockType: Scratch.BlockType.REPORTER,
+                        text: 'Decrypt [TEXT] with private key [KEY]',
```
yuri-kiss, line 47: same here as the other name comment

extensions/Themadpunter/rsa.js
```diff
@@ -0,0 +1,89 @@
...
+        decrypt(args) {
+            const crypt = new JSEncrypt();
+            crypt.setPrivateKey(args.KEY);
+            const decrypted = crypt.decrypt(args.TEXT);
+            return decrypted ? decrypted : 'Decryption failed';
+        }
+    }
+
+    // Wait for the script to load before registering the extension
+    script.onload = () => {
```
yuri-kiss, line 86: after doing what the other reviews suggests (removing the script tag) just remove this

extensions/Themadpunter/rsa.js
```diff
@@ -0,0 +1,89 @@
...
+                                defaultValue: ''
+                            }
+                        }
+                    }
+                ]
+            };
+        }
+
+        generateKeys() {
+            const crypt = new JSEncrypt({ default_key_size: 1024 });
```
yuri-kiss, line 64: nitpick: this could be customizable

extensions/Themadpunter/rsa.js
```diff
@@ -0,0 +1,89 @@
...
+            const publicKey = crypt.getPublicKeyB64();
+            const privateKey = crypt.getPrivateKeyB64();
+            return JSON.stringify({ publicKey, privateKey });
+        }
+
+        encrypt(args) {
+            const crypt = new JSEncrypt();
+            crypt.setPublicKey(args.KEY);
```
yuri-kiss, line 72: please cast your values

extensions/Themadpunter/rsa.js
```diff
@@ -0,0 +1,89 @@
...
+            const encrypted = crypt.encrypt(args.TEXT);
+            return encrypted ? encrypted : 'Encryption failed';
```
yuri-kiss, line 74: instead of returning text which could be the same as the actual output value you should just return an empty string

extensions/Themadpunter/rsa.js
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        decrypt(args) {
+            const crypt = new JSEncrypt();
+            crypt.setPrivateKey(args.KEY);
+            const decrypted = crypt.decrypt(args.TEXT);
+            return decrypted ? decrypted : 'Decryption failed';
```
yuri-kiss, line 81: same goes here for the returning an empty string

Thebloxers998:
@familycomicsstudios, This extension is **broken**, you need to fix this

yuri-kiss:
> @familycomicsstudios, This extension is **broken**, you need to fix this

do not ping people because their stuff doesnt work, they will get to it if they have time.
and dont approve random prs cause you find them "cool", you are slowing down the process.

yuri-kiss:
Closed for being stale.

PR #1702, opened 2024-09-27
Create HamzaMath

aya895:
some math blocks for calculators

### DIFF
--- a/extensions/HamzaMath
+++ b/extensions/HamzaMath
@@ -1,0 +1,178 @@
+(function() {
+    const extension = {
+        getInfo: function() {
+            return {
+                id: 'hamzaMath',
+                name: 'HamzaMath',
+                color: '#006400',
+                blocks: [
+                    // Separator between the two groups
+                    {
+                        opcode: 'separator1',
+                        blockType: Scratch.BlockType.LABEL,
+                        text: '--- Calculate ---',
+                        group: 'Calculates' // Group for arithmetic operations
+                    },
+                    // Arithmetic operations group
+                    {
+                        opcode: 'calculate',
+                        blockType: Scratch.BlockType.REPORTER,
+                        text: 'Calculate: [EXPRESSION]',
+                        arguments: {
+                            EXPRESSION: {
+                                type: Scratch.ArgumentType.STRING,
+                                defaultValue: '1 + 1'
+                            }
+                        },
+                        group: 'Calculates' // Group for the blocks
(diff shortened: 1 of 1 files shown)


SharkPool-SP:
You cant use eval, its dangerous and can run JS code.
`alert("see what I mean?")`

Also, the calculate blocks are useless as you can use the "abs" block from operators. You could just add the other blocks to the Math extension

Drago-Cuven:
Oh yeeeah a message popping on your screen to simply say something is
sooooo dangerous. Sooooo many bad things could happen. It deeefinetely be
such a catastrophe.

SharkPool-SP:
@Drago-Cuven now you're real silent :trollface:

extensions/HamzaMath
```diff
@@ -0,0 +1,178 @@
+(function() {
```
SharkPool-SP, line 1: should be:
```js
(function(Scratch) {
  "use strict";
  ```

File: extensions/HamzaMath [hunk 2]
```diff
@@ -0,0 +1,178 @@
...
+    const extension = {
```
SharkPool-SP, line 2: your extension needs to be wrapped in a class, and its block functions

extensions/HamzaMath
```diff
@@ -0,0 +1,178 @@
...
+        },
+        toPositive: function(args) {
+            return Math.abs(args.NUMBER); // Convert the number to positive
+        },
+        toNegative: function(args) {
+            return -Math.abs(args.NUMBER); // Convert the number to negative
+        }
+    };
+
+    Scratch.extensions.register(extension);
```
SharkPool-SP, line 177: should be
```js
  Scratch.extensions.register(new SPAddedMotion());
})(Scratch);
```

extensions/HamzaMath
```diff
@@ -0,0 +1,178 @@
...
+                ]
+            };
+        },
+        calculate: function(args) {
+            try {
+                const result = eval(args.EXPRESSION);
+                return result;
+            } catch (error) {
+                return 'Error'; // Return error message for invalid expression
+            }
```
SharkPool-SP, line 142: You should **never** be using eval like this. People will be able to run any javascript code, which is dangerous

extensions/HamzaMath
yuri-kiss: actually any object is fine lol

PR #1748, opened 2024-11-06
Add Extensions login page

Eaielectronic:
Creat login page


Drago-Cuven:
Layering issue
image
that being said, there's also no way to make it show specifically on the stage?

CST1229:
> there's also no way to make it show specifically on the stage?

it's actually really easy to make html elements show up on the stage (Scratch.renderer.addOverlay, the iframe extension uses it if i recall correctly)

Eaielectronic:
yes, I know but I don't want to put it in the scene because otherwise we can no longer select the text inside and use copy paste or tab. And once in pakage it appears in the middle of the screen.

Eaielectronic:
I fixed the bug.

Thebloxers998:
> Layering issue image that being said, there's also no way to make it show specifically on the stage?

I can see BlockifyVR extension

Thebloxers998 (approved):

Thebloxers998:
We need a <is authenticated/registered>

GarboMuffin:
this seems to be in the same boat as captcha & security keys -- it's only client side so there's no actual security but people will still treat it as though there is security.

Eaielectronic:
I just wanted to create a simple interface to connect and then use an extension like https websockets or cloudlink for the actual connection between client and server.

Thebloxers998:
> I just wanted to create a simple interface to connect and then use an extension like https websockets or cloudlink for the actual connection between client and server.

Add a when authenticated/registered hat block and a boolean one pls and then add settings for the ui

Thebloxers998:
I love how it saves my acc in this extension

Thebloxers998:
And make a block so we can add animationvto ui

Eaielectronic:
Ok

Thebloxers998:
> Ok

Yay

Thebloxers998:
> Ok

Also there's  a bug I found, in create account ui, there's a recover password thing it should be "Use Existing Account" thing,  especially recover password ui

Thebloxers998:
> this seems to be in the same boat as the captcha & security key extensions -- it's only client side so there's no actual security but people will still treat it as though there is security.

Thank God the pull request wasn't closed

Thebloxers998:
> > this seems to be in the same boat as the captcha & security key extensions -- it's only client side so there's no actual security but people will still treat it as though there is security.
> 
> Thank God the pull request wasn't closed

Also, don't take it too seriously

Thebloxers998:
> Ok

So have you added it yet?

Thebloxers998:
> I just wanted to create a simple interface to connect and then use an extension like https websockets or cloudlink for the actual connection between client and server.

That won't work, because it stores the account in the client instead, that could be unsafe

PR #1763, opened 2024-11-27
Lily/ClonesPlus: re-re fix performance for "when I start as w/ set to ..." block

SharkPool-SP:
Now it should work. It waits for variables to initialize

### DIFF
--- a/extensions/Lily/ClonesPlus.js
+++ b/extensions/Lily/ClonesPlus.js
@@ -450,22 +450,19 @@
       try {
         isCreatingCloneWithVariable = true;
 
-        // @ts-expect-error - ext_scratch3_control not typed yet
-        runtime.ext_scratch3_control._createClone(
-          '_myself_',
-          util.target
-        );
+        const original = util.target;
+        const clone = util.target.makeClone();
+        if (clone) {
+          Scratch.vm.runtime.addTarget(clone);
+          clone.goBehindOther(original);
 
-        const clones = util.target.sprite.clones;
-        const cloneNum = clones.length - 1;
-        const clone = clones[cloneNum];
-
-        const cloneVariable = clone.lookupVariableById(args.INPUTA);
-        if (cloneVariable) {
-          cloneVariable.value = args.INPUTB;
+          const cloneVariable = clone.lookupVariableById(args.INPUTA);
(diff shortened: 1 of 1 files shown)


CubesterYT (dismissed):
😔

GarboMuffin:
Fails on this project

https://static.muffin.ink/extensions-1763.sb3

image


production:

image

this pull request:

image

GarboMuffin:
Anything that isn't an edge-activated hat will differ at least a little bit. That could break projects. But maybe worth doing since this is probably rare and that it's not running right after the vanilla block is arguably a bug. Though in this project it doesn't even run at all so that's a pretty bad failure case.

Here is where the vanilla event gets fired. You could do it in the same spot.

https://github.com/TurboWarp/scratch-vm/blob/b900a375fa1c4a9574835c308baf5f8f16fd6cd9/src/sprites/rendered-target.js#L177

SharkPool-SP:
> Anything that isn't an edge-activated hat will differ at least a little bit. That could break projects. But maybe worth doing since this is probably rare and that it's not running right after the vanilla block is arguably a bug. (if you follow this logic expected output would be `when i start as clone` then `when i started as clone with` then `when timer` then `forever` then `end`).
> 
> Here is where the vanilla event gets fired. You could do it somewhere around. Might not be able to do this spot exactly since the variable won't be set yet.
> 
> https://github.com/TurboWarp/scratch-vm/blob/b900a375fa1c4a9574835c308baf5f8f16fd6cd9/src/sprites/rendered-target.js#L177

fixed

GarboMuffin:
btw the unrelated changes like Scratch.vm.runtime -> runtime probably didn't really need to be included

GarboMuffin:
!format

SharkPool-SP:
> btw the unrelated changes like Scratch.vm.runtime -> runtime probably didn't really need to be included

eh, ever so slightly faster + simpler

GarboMuffin:
at the cost of making the diffs bigger  for no reason and  increasing change of a typo  causing a regression

GarboMuffin:
ok will use this pr to test more things related to the formatting workflow being broken

GarboMuffin:
!format

PR #1769, opened 2024-11-30
Integers Extension

xFN10x:
this extension has some basic reporters, for with dealing with really large, or small numbers

### DIFF
--- a/extensions/_FN10_/integers.js
+++ b/extensions/_FN10_/integers.js
@@ -1,0 +1,209 @@
+/*
+   This extension was made with TurboBuilder!
+   https://turbobuilder-steel.vercel.app/
+*/
+(async function(Scratch) {
+    const variables = {};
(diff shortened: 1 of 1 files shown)


CubesterYT:
Ok first off, TurboBuilder is horrible when it comes to making extensions. There's already a lot of clear bloat from useless functions. Not just that, but it's very messy. I would much rather you try to make the extension yourself in the method extensions are normally made.

xFN10x:
> Ok first off, TurboBuilder is horrible when it comes to making extensions. There's already a lot of clear bloat from useless functions. Not just that, but it's very messy. I would much rather you try to make the extension yourself in the method extensions are normally made.

ok, It is a pretty simple extension, so I will make it myself

xFN10x:
i rewrote the extension, and updated the extensions.json

PR #1770, opened 2024-11-30
Add formatting workflow

CubesterYT:
Add a new workflow that should format a PR when a message is "!format"

Further testing by others is needed.


SharkPool-SP (approved):

Thebloxers998 (approved):

CST1229:
I still think this should be a `workflow_dispatch` workflow.

CubesterYT:
Are `workflow_dispatch`s accessible by normal people?

GarboMuffin:
If they enable GitHub actions on their fork. Do they need to do that anyways for this to work?

CubesterYT:
As it currently works, when the comment command is said, actions will auto format the PR, in theory. I haven't done extensive testing, so I'd like if others actually tried.

CST1229:
> As it currently works, when the comment command is said, actions will auto format the PR, in theory. I haven't done extensive testing, so I'd like if others actually tried.

i don't think actions would allow committing on a user's branch without them enabling actions (see also: sa's autoformat action which i'm pretty sure you do have to enable actions for)

GarboMuffin:
Ok, will just merge it and see what happens

For your reference there were a few problems with the original code

 - missing write contents permission (this entire organization is readonly by default)
 - ${{ }} inside shell script allows arbitrary shell code execution
 - not filtering to just pull request comments

CubesterYT:
Great, this will help so much!

Thebloxers998:
> Great, this will help so much!

Wait, but how do I use this

Thebloxers998:
> Great, this will help so much!

Wait, but how do I use this

PR #1796, opened 2024-12-21
Lily/TempVariables2: nitpicking

GarboMuffin:
Safari 13 and 14 have enough users that they show up in the top 100 user agents on clouddata.turbowarp.org so I'd like to keep as much alive there for as long as possible. Thus avoid using ??= (Safari 14) or Object.hasOwn (Safari 15.4).

Also includes ESLint rules, so that these are detected automatically moving forward

`ext_${extensionId}` is also a turbowarp convention not a Scratch convention

Probably at some point it makes sense to just adopt babel and transpile this stuff down to Safari 12 but for now, avoiding significant transformations is nice

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -112,6 +112,14 @@ module.exports = [
       'no-restricted-syntax': [
         'error',
         {
+          selector: 'AssignmentExpression[operator="??="]',
+          message: 'x ??= y syntax is too new; use x = x ?? y intead'
+        },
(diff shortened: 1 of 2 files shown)


GarboMuffin:
Other browsers added the features at roughly the same time. Problem with Safari is that its updates are tied to Apple OS updates and many people use old iOS, macOS, etc. versions while other browsers update independently

PR #1824, opened 2024-12-29
Use a service account PAT for format workflow

GarboMuffin:
Deploy key's write access is limited to TurboWarp/extensions, can't touch forks. Instead it now uses a PAT which I've tested to have the necessary permissions

Comment explains why the default workflow token isn't sufficient.


GarboMuffin:
!format

GarboMuffin:
Okay that workflow failed but that's expected because it's running the one on master which uses a deploy key that doesn't exist anymore

PR #1854, opened 2025-01-08
0832/rxFS: fix delete block, hide search block

GarboMuffin:
- Delete blocks leave holes in the arrays, breaking other blocks
 - The search block has been broken this entire time (`.indexOf(...) == undefined` is always false). It's not clear what it's even supposed to do, so just hide it.


GarboMuffin:
Ok I won't delist it yet, because there isnt a clear 1:1 replacement for it

PR #1860, opened 2025-01-09
Change some block labels

DNin01:
| Old | New |
|--------|--------|
| `without waiting` | and `continue` |
| broadcast `message` to `sprite` | send `message` to `sprite` |

### DIFF
--- a/docs/CubesterYT/KeySimulation.md
+++ b/docs/CubesterYT/KeySimulation.md
@@ -5,7 +5,7 @@ This extension allows you to simulate key presses and mouse clicks on the projec
 ## Press key
 
 ```scratch
-press (space v) for (0.1) seconds (without waiting v) :: #BF0000
+press (space v) for (0.1) seconds and (continue v) :: #BF0000
 ```
 
 This will trigger "key down?" and "when key pressed" blocks. It won't type text into text fields in the editor, for example.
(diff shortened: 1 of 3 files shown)


GarboMuffin:
im not opposed to it

CubesterYT:
This is great, the previous wording was bad

CubesterYT (approved):

SharkPool-SP:
I am opposed to the changes in More Events. The blocks are supposed to be based on this:
image

DNin01:
> I am opposed to the changes in More Events. The blocks are supposed to be based on this

I only changed "broadcast" to "send" on the sprite-specific message-sending blocks. The blocks that broadcast messages to all sprites still say "broadcast".

Blocks to broadcast messages or send messages to sprites

GarboMuffin:
"broadcast" in the parallel programing sense means "send to everyone" which those blocks don't do. MPI separates these operations into MPI_Bcast and MPI_Send

GarboMuffin:
Though it still is a broadcast since it goes to all the clones of that sprite too

CubesterYT:
I mean, if there is clones, you can think of it as a Carbon Copy type of send

Where it is still sending, but there are extra recipients getting the same information

NexusKitten (commented):
No issues with the changes in Key Simulation, looks like an overall good change.
Not a fan of the changes in More Events, though. Technically it's different from a broadcast, but it doesn't seem different enough to warrant creating new terminology. It's not exactly a negative change, I just don't see much of a point.

yuri-kiss:
I agree with nameless, and sharkpool

SharkPool-SP (approved):

DNin01:
Thanks

PR #1861, opened 2025-01-11
Complex Numbers Extension

BonesYT:
I couldn't find an extension dedicated to complex numbers, you know... sqrt(-1) and stuff like that, so I made one.
This is also my first extension, and it kinda works pretty well!

**All features**:
- Construct a complex number with real and/or imaginary component

- Get the real and imaginary components of that value
- Boolean to check if a number is real or not
- Addition, subtraction, multiplication, division and exponentiation. And modulo.
- Functions with single input, with extra: conjugate, flip, sign, atan2.
- Toggle between simple and advanced syntax, for perfomance reasons. Maybe simple mode is at least slightly faster.

**Examples**
1. Mandelbrot Set
image
2. Julia Set
image


extensions/BonesYT/ScratchComplex.js
```diff
@@ -467,4 +467,4 @@
     /* add methods for blocks */
   }
   Scratch.extensions.register(new ScratchComplex());
-})(Scratch);
\ No newline at end of file
+})(Scratch);
```
BonesYT, line 470: what do you mean format only works if the very last line is empty??????????????????????

penta-quark-neutro:
Interesting, I also made one of the same thing, but for personal use.
Your extension is quite good, but slow.
It takes almost 2.57 seconds to exponentiate 1000000 complex numbers.
Approximately 4 times slower than mine.
I think you should optimize it.

extensions/BonesYT/ScratchComplex.js
```diff
@@ -0,0 +1,473 @@
...
+// License: MPL-2.0
+(function (Scratch) {
+  /* Version 1.0 - 01/10/2025 MM/DD/YYYY */
+  "use strict";
+  if (!Scratch.extensions.unsandboxed) {
+    throw new Error("You're running an unsandboxed extension as sandboxed.");
+  }
+
+  class ScratchComplex {
+    constructor() {}
```
Brackets-Coder, line 14: if constructor is empty may I ask why it's here? Maybe just a clarification thing

extensions/BonesYT/ScratchComplex.js
```diff
@@ -0,0 +1,473 @@
...
+              "ln",
+              "e^",
+              "10^",
+              "sign",
+            ],
+          },
+          complexSyntaxMode: {
+            acceptReporters: false,
+            items: ["simple (faster)", "advanced (slower)"],
+          },
```
Brackets-Coder, line 465: Menus should be translated as well, use this structure for each item:

```
item: {
   text: Scratch.translate("simple"),
   value: "simple",
}
```

extensions/extensions.json
```diff
@@ -1,100 +1,101 @@
...
+  "godslayerakp/ws",
+  "Lily/CommentBlocks",
+  "veggiecan/LongmanDictionary",
+  "CubesterYT/TurboHook",
+  "Alestore/nfcwarp",
+  "steamworks",
+  "itchio",
+  "gamejolt",
+  "obviousAlexC/newgroundsIO",
+  "BonesYT/ScratchComplex",
```
Brackets-Coder, line 99: Why'd you put this at the bottom? It doesn't really matter but you could put it near other similar extensions

images/BonesYT/ScratchComplex.svg
Brackets-Coder, line 1: Images aren't supposed to have text because of translation but I guess this is okay...

extensions/extensions.json
Brackets-Coder, line 1: Github did something weird with displaying the diff... but sure

Brackets-Coder (changes requested):
Haven't tested this fully yet, but it loads and seems to work
If this is your first extension, nice job! Here's just a few tips to make it better:

Avoid using the master branch, create a new branch for each pull request you'd like to submit for organization purposes.

Kenay555:
<casually breaks window>
Hi @BonesYT ! 
I may or may not made something similar (like, with more blocks and such...) see #2113 for more :D
Anyways, I'm working on updating my extension, because it's wierd with every letter. Plus, I found how to implement Quaternions, and maybe have more optimization ideas for it, maybe too many...
I liked how you tackled some problems in the file. But that's not why I am here.

I was recently told about your existence, and looks like your extension has more power than mine (altho I have more blocks >:D but that's not the point either). Wanna work together?
image

(or maybe we should call @penta-quark-neutro to see what optimizations he used!)

PR #1863, opened 2025-01-13
Add pull request labeling workflow

CubesterYT:
Simple workflow to automatically add PR tags, we do end up forgetting to add tags, so this will do.

I'm still really bad at writing workflows, so I wouldn't be surprised if this is pretty bad and/or can be better. But this is a start.


CubesterYT:
Ok, this is really problematic, this is going to need help from more experienced workflow writers.

Brackets-Coder:
This sounds intriguing, but I'm not that familiar with the technical automation of GitHub repositories, so unfortunately I can't help you. Probably GarboMuffin?

PR #1864, opened 2025-01-13
veggiecan/mobilekeyboard: fix typed text block returning undefined

### LINKED ISSUE #1765: Mobile Keyboard: typed text is empty, but "length of (typed text)" returns 9 somehow
I'm trying to make a custom text box for myself.
I used the following script:
block_11_27_2024-8_04_27 AM
As you can see, I absolutely make sure that the text box is completely empty before I do any actions on it.
And yet, when this script runs, the text shows "undefined", and the length of the inputted text somehow immediately jumps up to 9.
image
I'm not sure if this is supposed to happen, but my guess is that it *shouldn't happen*.

CubesterYT:
Resolves #1765

Simple fix, really. I kept forgetting to make this change.

### DIFF
--- a/extensions/veggiecan/mobilekeyboard.js
+++ b/extensions/veggiecan/mobilekeyboard.js
@@ -242,7 +242,7 @@
     }
 
     typedTextSinceKeyboardOpened() {
-      return this.typedText;
+      return this.typedText ?? "";
     }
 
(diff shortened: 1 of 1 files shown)


CubesterYT:
Oh right, yeah true 😅

PR #1880, opened 2025-01-21
NexusKitten/moremotion: Replace atan with atan2

DNin01:
Advantages:
- Does essentially the same thing but with less lines of code
- Reports values between -180 and 180 instead of between -90 and 270

Kind of like #1460 but for a different extension.

Of course, the old blocks are still there for backwards compatibility.

### DIFF
--- a/extensions/NexusKitten/moremotion.js
+++ b/extensions/NexusKitten/moremotion.js
@@ -65,6 +65,24 @@
               },
             },
             extensions: ["colours_motion"],
+            hideFromPalette: true,
+          },
+          {
+            filter: [Scratch.TargetType.SPRITE],
+            opcode: "pointto_v2",
+            blockType: Scratch.BlockType.COMMAND,
+            text: Scratch.translate("point towards x: [X] y: [Y]"),
+            arguments: {
+              X: {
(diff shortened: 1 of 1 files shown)


GarboMuffin:
Might not need to do a v2 for this

GarboMuffin:
Autolaber failure unrelated to your changes

GarboMuffin:
"direction to" needs a v2 but "point to" does not

GarboMuffin (approved):

CubesterYT (approved):

extensions/NexusKitten/moremotion.js
```diff
@@ -65,6 +65,7 @@
               },
             },
             extensions: ["colours_motion"],
+            hideFromPalette: true,
```
DNin01, line 68: @GarboMuffin Oops? I'm pretty sure this block should be visible on the block palette.

GarboMuffin: oops

GarboMuffin: think i have a comment in there that's also wrong, i said old version was 0-360 when its -90-270

Issue #1787, opened 2024-12-14
Pause button doesn't affect Tween for whatever reason


SaltGrain347:
I was making music player with tweening using Sound Expanded Extension and Tween Extension but for whatever reason, when I hit pause, the tween is left uninterrupted/is not affected by pausing.

https://github.com/user-attachments/assets/7ec15d0e-3c5c-46ba-9df8-8d4a9b45d7e2

SaltGrain347:
hello???

GarboMuffin:
Sorry we are very busy volunteers

fath11:
The extension uses `Scratch.vm.runtime.currentMSecs` to tell time while tweening which doesn't stop ticking when the project is paused or stopped. Is this intended?

GarboMuffin:
no

GarboMuffin:
it would be reasonable to convert it to use the builtin vanilla stack timer, then pause support would be automatic

PR #1887, opened 2025-01-25
JeremyGamer13/tween: Fix Pause button not suspending Tweening

fath11:
Resolving #1787

### DIFF
--- a/extensions/JeremyGamer13/tween.js
+++ b/extensions/JeremyGamer13/tween.js
@@ -553,7 +553,8 @@
         // Still running
         util.yield();
 
-        const progress = util.stackFrame.timer.timeElapsed() / util.stackFrame.duration;
+        const progress =
+          util.stackFrame.timer.timeElapsed() / util.stackFrame.duration;
         const tweened = state.easingFunction(progress, state.easeDirection);
         return interpolate(tweened, state.start, state.end);
       }
--- a/extensions/JeremyGamer13/tween.js
+++ b/extensions/JeremyGamer13/tween.js
@@ -250,8 +250,6 @@
     bounce,
   };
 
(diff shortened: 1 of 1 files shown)


SharkPool-SP (approved):

GarboMuffin (approved):

CST1229:
wouldn't this break when using the reset timer block?

GarboMuffin (changes requested):
yes it would

SharkPool-SP:
> wouldn't this break when using the reset timer block?

didnt think of that.

GarboMuffin:
There's also the edge case where the tweening block can be started while the project is already paused, in which case it should play as if the project were not paused

So this is probably going to be a good bit more complicated

It's 1:30AM so you probably don't want me to suggest something right now

SharkPool-SP:
Its 11:30pm for me rn, best I can think of is set up our own timer for tween that iterates using before execute and pauses when runtime is paused.

CST1229:
> Its 11:30pm for me rn, best I can think of is set up our own timer for tween that iterates using before execute and pauses when runtime is paused.

yea, that is also exactly how the glide block works

fath11:
since the glide block doesn't have all those issues its prob best to just copy it which shouldn't be hard

i will push the fix tomorrow

SharkPool-SP:
> since the glide block doesn't have all those issues its prob best to just copy it which shouldn't be hard
> 
> i will push the fix tomorrow 

It's been tomorrow

fath11:
ive been a little busy but i should be free today

GarboMuffin (approved):

CubesterYT (approved):

CubesterYT:
!format

PR #1898, opened 2025-01-29
Lily/MoreTimers: Don't rely on system clock

DNin01:
This prevents the More Timers extension from being thrown off by system clock changes.

Tested, no breaking changes.

### DIFF
--- a/extensions/Lily/MoreTimers.js
+++ b/extensions/Lily/MoreTimers.js
@@ -25,7 +25,8 @@
    */
   const timerValue = (timer) => {
     return (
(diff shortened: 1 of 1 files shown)


DNin01:
!format

CubesterYT:
Maybe we could look into stack timer?

GarboMuffin:
Stack timer great for within a block but the timers outlive each individual block here

GarboMuffin:
Wouldn't be opposed to removing the Math.floor()

GarboMuffin (approved):

CubesterYT (approved):

PR #1904, opened 2025-02-03
Lily/Video: added "play video until done" and "loop video" block

### LINKED ISSUE #1903: Lily/Video: add looping videos and "play until done"

fath11:
Resolves #1903 

Should we put loop in `getAttribute` or `getState`?

### DIFF
--- a/extensions/Lily/Video.js
+++ b/extensions/Lily/Video.js
@@ -263,10 +263,12 @@
               },
             },
           },
-		  {
+          {
             opcode: "startVideoAndWait",
             blockType: Scratch.BlockType.COMMAND,
-            text: Scratch.translate("start video [NAME] at [DURATION] seconds and wait until done"),
+            text: Scratch.translate(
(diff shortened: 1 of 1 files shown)


fath11:
!format

SharkPool-SP:
this implementation will break compatibility

SharkPool-SP:
It would be best to have a toggle loop block and a separate play video until done block

SharkPool-SP:
additionally, the "play until done" block should actually make the block wait

fath11:
right, i forgot about compatibility

fath11:
!format

fath11:
I can't decide if toggle loop block should be separate from play until done. Because if they are separate then play until done will never finished running. One way to combat that is to make play until done block disable looping which will make more sense if loop is toggled from the same block.

fath11:
!format

fath11:
!format

SharkPool-SP (approved):

PR #1907, opened 2025-02-05
Skyhigh173/json: add analysis block

yuri-kiss:
this adds the same analysis support as data-analysis but into JSON (to get rid of the need to do stuff like the image below)
image


yuri-kiss:
!format

yuri-kiss:
!format

SharkPool-SP:
can you just add this to the other JSON. v2.6 PR. Would be alot easier

yuri-kiss:
> can you just add this to the other JSON. v2.6 PR. Would be alot easier

No

CubesterYT:
> can you just add this to the other JSON. v2.6 PR. Would be alot easier

That PR is stale, its best to keep this separate.

CubesterYT:
!format

yuri-kiss:
> !format

Doesn't work

CubesterYT (approved):

extensions/Skyhigh173/json.js
```diff
@@ -560,6 +576,39 @@
               },
             ],
           },
+          analysis: {
+            acceptReporters: true,
+            items: [
+              {
+                text: Scratch.translate("minimum value"),
```
NexusKitten, line 583: you could probably get away with just making this "minimum" as opposed to "minimum value," takes up less space and matches more with the other inputs (i.e. "average" or "median")

extensions/Skyhigh173/json.js
```diff
@@ -560,6 +576,39 @@
...
+                value: "minimum",
+              },
+              {
+                text: Scratch.translate("maximum value"),
```
NexusKitten, line 587: same thing as previous comment, this could probably just be "maximum"

extensions/Skyhigh173/json.js
```diff
@@ -1008,6 +1057,58 @@
...
+          }
+          return mode[0];
+        }
+        case "variance": {
+          const average = list.reduce((a, b) => a + b, 0) / listLength;
+          const list2 = list.map((a) => (a - average) ** 2);
+          return list2.reduce((a, b) => a + b, 0) / listLength;
+        }
+      }
+      return 0;
```
NexusKitten, line 1110: it's probably a good idea to return with an empty string rather than a zero; from the looks of it, that's what the rest of the blocks in this extension do

extensions/Skyhigh173/json.js
```diff
@@ -481,6 +481,22 @@
...
           },
+          {
+            opcode: "json_array_analysis",
+            blockType: Scratch.BlockType.REPORTER,
+            text: Scratch.translate("[analysis] of array [list]"),
+            disableMonitor: true,
+            arguments: {
+              list: {
+                type: Scratch.ArgumentType.STRING,
+                defaultValue: "[17, 1, 2017, 0, 120, 14]",
```
NexusKitten, line 492: any reason for these specific numbers? it might be cleaner just to do "[1, 2, 3, 4, 5]"

NexusKitten (commented):
most of these are just stylistic changes, considering most of this code is similar/the same as https://github.com/TurboWarp/extensions/pull/352 and that was already reviewed and accepted, I'd doubt this would cause any new issues

extensions/Skyhigh173/json.js
yuri-kiss: well this block specifically returns a number so im going to keep it as 0

extensions/Skyhigh173/json.js
yuri-kiss: its just a bunch of random numbers, 1,2,3,4,5 doesnt do anything with mode

extensions/Skyhigh173/json.js
yuri-kiss: this was also a stylistic choice, it reads better as value

extensions/Skyhigh173/json.js
yuri-kiss: this was also a stylistic choice, it reads better as value

yuri-kiss:
> most of these are just stylistic changes, considering most of this code is similar/the same as #352 and that was already reviewed and accepted, I'd doubt this would cause any new issues

I wrote the code myself I just wanted to port the blocks 🤷

extensions/Skyhigh173/json.js
yuri-kiss: also I will say in my previous JSON PR the random numbers weren't an issue when they easily could have just been that array reversed so its not like this is out of the question just in case you are wondering lol

NexusKitten:
> > most of these are just stylistic changes, considering most of this code is similar/the same as #352 and that was already reviewed and accepted, I'd doubt this would cause any new issues
> 
> I wrote the code myself I just wanted to port the blocks 🤷

Yeah, I noticed the code was different, however the way you two approached it was generally the same (specifically for the more simple blocks with built-in JS functions or well-known workarounds)

extensions/Skyhigh173/json.js
```diff
@@ -1008,6 +1057,58 @@
...
+        return 0;
+      }
+      if (!Array.isArray(list)) {
+        return 0;
+      }
+      list = list.map(Scratch.Cast.toNumber);
+      const listLength = list.length;
+      switch (Scratch.Cast.toString(args.analysis)) {
+        case "maximum":
+          return Math.max(...list);
```
GarboMuffin, line 1074: Using spread syntax like this collapses when list gets too long. Should be a boring for loop. Feel free to submit another PR fixing the original. Don't try to cram it in here.

yuri-kiss: Okay!

SharkPool-SP (approved):

yuri-kiss:
Do not merge till garbo says it's ok please

PR #1921, opened 2025-02-08
SharkPool/Camera.js: fix speech bubble positions

SharkPool-SP:
Pointed out by Nameless. Thanks!

### DIFF
--- a/extensions/SharkPool/Camera.js
+++ b/extensions/SharkPool/Camera.js
@@ -161,7 +161,7 @@
     );
 
     bounds.top = realTopLeft[1];
-    bounds.left =  realTopLeft[0];
+    bounds.left = realTopLeft[0];
     bounds.bottom = realBottomRight[1];
     bounds.right = realBottomRight[0];
     return bounds;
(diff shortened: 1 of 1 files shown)


SharkPool-SP:
!format

extensions/SharkPool/Camera.js
```diff
@@ -142,6 +142,31 @@
...
+      [bounds.left, bounds.top],
+      true,
+      camSystem
+    );
+    const realBottomRight = translatePosition(
+      [bounds.right, bounds.bottom],
+      true,
+      camSystem
+    );
+    return {
```
GarboMuffin, line 162: It used to return a scratch-render Rectangle now you're making it return a  plain object so the contract is being broken

GarboMuffin: What you can do is take the original object you get returned and then modify it in-place instead of making something new

SharkPool-SP:
!format

extensions/SharkPool/Camera.js
SharkPool-SP: done

GarboMuffin (approved):

CubesterYT (approved):

PR #1922, opened 2025-02-08
[Automated] Update translations 2025-02-08

DangoCat:
This pull request was made by a robot.


NexusKitten:
How come we're removing the translations for the old camera extension? Obviously it's deprecated at this point but I still think it's worth leaving the translations in for users that may already have projects using it, among other reasons.

SharkPool-SP:
Idk. It's the way it's setup. A lot of hidden or unused stuff isn't translated

GarboMuffin:
The translation generation only looks at extensions.json right now

SharkPool-SP:
Yeah add mcutils to that list

PR #1923, opened 2025-02-08
Add `client-metadata.json`, `bskycallback.html` for Bluesky's OAuth System

hammouda101010:
### Changes Made:
- Added `client-metadata.json` inside the `website` folder
- Added `bskycallback.html` inside the `website` folder

### Notes:
This doesn't add an new extension, but it has to be done for support for BlueSky's OAuth system (because it uses `IndexedDB`, which is not shared by origin), so that [TurboButterfly's new OAuth rework](https://github.com/hammouda101010/turbowarp-bsky-api) can work properly.

I will create a pull request of the extension when it will be done.
This won't conflict with any extension anyways.

### DIFF
--- a/website/bskycallback.html
+++ b/website/bskycallback.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8"/>
+    <title>TurboWarp OAuth Callback for BlueSky</title>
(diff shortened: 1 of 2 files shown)


yuri-kiss:
This should be added along with the extension and preferably added to its own bluesky specific folder on the site.

PR #1959, opened 2025-02-23
true-fantom/math: Add true sin/cos/tan block.

NexusKitten:
sorry mio

In Scratch/Turbowarp, the sin/cos/tan/asin/acos/atan blocks don't return as expected-- they're run through several other equations to deal with rounding, and to make the blocks compatible with degrees. 

chrome_eTFJtRFaWZ

While this makes the blocks more user-friendly, it can be annoying for people who want the original functions to have to essentially undo the extra calculations. This PR adds a new block to the Math extension which introduces a new block, featuring all the same trig functions, but returning exactly what their Javascript Math counterpart does.

chrome_zkk2vNonX3

### DIFF
--- a/extensions/true-fantom/math.js
+++ b/extensions/true-fantom/math.js
@@ -699,8 +699,9 @@
           return Math.acos(n);
         case "atan":
           return Math.atan(n);
+        default:
+          return 0;
       }
-      return 0;
     }
     pi_block() {
       return Math.PI;
--- a/extensions/true-fantom/math.js
+++ b/extensions/true-fantom/math.js
@@ -497,6 +497,21 @@
             },
             extensions: ["colours_operators"],
           },
+          {
+            opcode: "true_math_op",
+            blockType: Scratch.BlockType.REPORTER,
+            text: Scratch.translate("true [OPERATOR] [NUM]"),
+            arguments: {
+              OPERATOR: {
+                type: Scratch.ArgumentType.STRING,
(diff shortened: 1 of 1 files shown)


NexusKitten:
!format

CubesterYT:
What happened to my boy `false` sin/cos/tan/asin/acos/atan?

SharkPool-SP (approved):
I dont think these menu items need to be translated, so I think its good

extensions/true-fantom/math.js
```diff
@@ -662,6 +683,25 @@
...
+          return Math.cos(n);
+        case "tan":
+          return Math.tan(n);
+        case "asin":
+          return Math.asin(n);
+        case "acos":
+          return Math.acos(n);
+        case "atan":
+          return Math.atan(n);
+      }
```
CubesterYT, line 702: Maybe have a `default` branch that returns nothing or 0 for safety

CubesterYT (changes requested):

extensions/true-fantom/math.js
KyleKart: This makes total sense, doesn't Scratch do this itself?

NexusKitten: technically the old code had a default (returning zero if nothing happened in the switch case) but for consistency sake, i've gone ahead and turned it into a default branch

CubesterYT (approved):

PR #1972, opened 2025-02-25
Add new extension: NGL anonymous message sender extension

matu6968:
This is a extension which allows you to send a message through anybody over NGL's API.
Enjoy!


WAYLIVES:
What is the NGL API?

matu6968:
> What is the NGL API?
It is a anonymous messaging service at https://ngl.link

matu6968:
!format

WAYLIVES:
> It is a anonymous messaging service at

Ah, I see. It looks very cool, although I don't understand it)))

matu6968:
!format

PPPDUD:
@CubesterYT @SharkPool-SP @GarboMuffin Do any of you know why validation failed? I can't figure out what's wrong with the extension.

matu6968:
> @CubesterYT @SharkPool-SP @GarboMuffin Do any of you know why validation failed? I can't figure out what's wrong with the extension.
It said something was wrong with validating the URL, nothing to worry about since i had to locally test it on my machine

Brackets-Coder:
@CubesterYT correct me if I'm wrong I don't want to make any mistakes or cause extra problems here

PPPDUD:
> I would close this. It's not about suppressing communication, it's about future compatibility and security for our users. We can't have unmoderated communication for obvious reasons (notably child safety, security, other malicious reasons, etc.), the domain or external server could fail at any given moment, the domain could be inaccessible on a whitelisted network, and a slew of other reasons. It's been a hard ground rule in the past that TW not rely on external dependencies, so that automatically rules this out, and when combined with uncensored communication, I'd say probably close to 0 chance of being merged.

I don't want to be enemies with you, but "child safety" and "security" are no excuse for limiting freedom of speech. If we educate users to use pseudonyms, never reveal personal info, and stop chatting if they feel endangered or offended, we shouldn't need censorship.

PPPDUD:
@Brackets-Coder I would not call that mental illness; it does not incapacitate the afflicted but instead induces urges that may result in harm to others. I will point out that this medium is also both censored and moderated, almost to an absurd extent.

@GarboMuffin Would you please lock this before I start a flame war?

Brackets-Coder:
@PPPDUD I hope you understand my perspective as I understand yours. Both arguments have their pros and cons and it's outside the scope of what can be discussed here. My job is to speak the truth loving kindness, not to create bitterness between people. If I've done what I was trying to avoid, I entirely apologize

PR #1980, opened 2025-02-26
utilities: Port "normalize" block from Lily's Toolbox

NexusKitten:
Technically isn't a 1:1 port (I cleaned up the code a little and added an extra Scratch.Cast call just to be safe), however it functions practically identically.

chrome_spL57aYLcY

### DIFF
--- a/extensions/utilities.js
+++ b/extensions/utilities.js
@@ -208,6 +208,17 @@
             },
           },
           {
+            opcode: "normalize",
+            blockType: Scratch.BlockType.REPORTER,
+            text: Scratch.translate("normalize [INPUT]"),
+            arguments: {
+              INPUT: {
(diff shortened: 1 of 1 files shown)


NexusKitten:
!format

yuri-kiss:
"it's arguable that this would be better fit for the math extension, but i feel like it's general use enough to fit into utilities :idrk:" -Nameless on discord

CubesterYT:
I think it's suited for Math better, considering the Math extension has some of the same blocks utilities has. So we can just have it more centralized over expanding a general extension.

GarboMuffin:
This is not a normalize block it is a sign function block. Available in browsers as Math.sign

One of the math extensions makes more sense

NexusKitten:
superseded by https://github.com/TurboWarp/extensions/pull/1981

PR #1990, opened 2025-02-28
Method Fixes

CubesterYT:
Self explanatory.

I want a proper in depth review (Looking at you, SharkPool), so I can make sure I didn't miss anything or nothing is broken.

### DIFF
--- a/extensions/-SIPC-/time.js
+++ b/extensions/-SIPC-/time.js
@@ -387,16 +387,17 @@
       const totalSeconds = Scratch.Cast.toNumber(args.VALUE);
       const seconds =
         args.ROUND === "rounded"
-          ? Math.round(totalSeconds % 60)
-              .toString()
-              .padStart(2, "0")
+          ? Scratch.Cast.toString(Math.round(totalSeconds % 60)).padStart(
+              2,
+              "0"
+            )
           : (totalSeconds % 60).toFixed(3).padStart(6, "0");
-      const minutes = Math.floor((totalSeconds / 60) % 60)
-        .toString()
-        .padStart(2, "0");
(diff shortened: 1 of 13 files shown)


extensions/-SIPC-/time.js
```diff
@@ -387,16 +387,17 @@
       const totalSeconds = Scratch.Cast.toNumber(args.VALUE);
       const seconds =
         args.ROUND === "rounded"
-          ? Math.round(totalSeconds % 60)
-              .toString()
-              .padStart(2, "0")
+          ? Scratch.Cast.toString(Math.round(totalSeconds % 60)).padStart(
```
GarboMuffin, line 390: Changing toString() to Cast.toString() here has no value and if anything harms readability. You already know Math.round() is returning a number. Hope this isn't how the rest of the pull request looks.

GarboMuffin: Conceptually Scratch.Cast would be where any quirks in the anything-to-string conversion in Scratch-the-programming-language would live though it happens to be exactly the same as the default toString(). It doesn't need to be the same though.

Here, and many other places, you aren't dealing with a value from a Scratch project but rather some intermediate value you already know is a number. It's not benefiting at all to use Cast.

extensions/text.js
```diff
@@ -698,8 +700,8 @@
     }
 
     toCase(args, util) {
-      const string = args.STRING.toString();
-      const textCase = args.TEXTCASE.toString();
+      const string = Scratch.Cast.toString(args.STRING);
```
GarboMuffin, line 703: most of the changes in this extension do make sense though

CubesterYT:
Closed to be superseded by Miyo.

PR #1991, opened 2025-02-28
Window Settings

WAYLIVES:
# 🧱 Extension "Window Settings"
Window Settings2

HEY, friends! I've created a new extension for TurboWarp "Window Settings" :D

> Perhaps many of you'll say, "Oh, dude, why did you create this extension?! After all, such an extension already exists in the TurboWarp gallery!" Therefore, I will tell you in advance why I made this extension =)

There is such an extension as "Window Controls" in the TurboWarp gallery. With this extension, you can control the window where your project is located. But then why did I make a new similar extension? The problem is that this "Windows Controls" extension is not very convenient.

### For example:

> -

For example, I want to set the width and height of the scene: width = 640, height = 420. It is impossible to do this through this block, which you see in the screenshot, since the width and height of the window are set, not the stage, and the stage width and height will be smaller:
изображение

> -

This can be fixed through the following code, which you will see in the following screenshot. But the problem is that you will have to set a delay of more than 0 seconds. Which is not very convenient:
изображение

> -

Okay, then let's try to set the width and height through the next block, but again nothing worked out:
изображение

> -

But here everything is fast, everything is clear, the width and height of the stage are set without any delay, but at the same time the width and height of the window are changed:
изображение

https://github.com/WAYLIVES/TurboWARP/blob/main/extensions/WindowSettings/1.1.1.js


WAYLIVES:
How do I add a file here (.js)?

WAYLIVES:
!format

WAYLIVES:
!format

PR #2012, opened 2025-03-05
(W.I.P) New Extension: Pretty Blocks

hammouda101010:
## Pretty Blocks
This is an extension that lets you add formatting and format checking to your project much like Prettier.
It's also extensible (lets you create your custom rules), can import format plugins from URLs, can be used programaticaly and also has it's own custom "programming language" named "RegBool" that looks like this:
`if <true> /true/ else false`

Here are the features to finish:
- [ ] Format checking
- [ ] Formatting
- [ ] Expand "RegBool"'s Syntax
- [ ] Custom Plugins

I might note that this extension uses custom modal patches to show format errors and creating custom rules inside modals


Brackets-Coder:
Here a few seconds after this PR was made

extensions/Hammouda101010/prettyblocks.js
```diff
@@ -0,0 +1,880 @@
...
+
+  const vm = Scratch.vm;
+  const runtime = vm.runtime;
+  const Cast = Scratch.Cast;
+  /**Checks if the extension is in "penguinmod.com".*/
+  // @ts-ignore
+  const _isPM = Scratch.extensions.isPenguinMod;
+  /**Checks if the extension is in "unsandboxed.org".*/
+  const isUnSandBoxed =
+    JSON.parse(vm.toJSON()).meta.platform.url === "https://unsandboxed.org/";
```
LilyMakesThings, line 22: You can access the platforms of all 3 mods by doing `vm.runtime.platform`.

yuri-kiss (changes requested):
This is in the same boat as anything else that uses patches.

As'well as this is the TurboWarp gallery, the support for other mods is more work for US to maintain, it should be removed.

hammouda101010:
> This is in the same boat as anything else that uses patches.
> 
> As'well as this is the TurboWarp gallery, the support for other mods is more work for US to maintain, it should be removed.

so. technically, this extension won't be accepted due to the use of patches?
for the other mods support i can do that but i'll probably won't remove the patches (atleast the modal patches)

hammouda101010:
> so. technically, this extension won't be accepted due to the use of patches? for the other mods support removal i can do that but i'll probably won't remove the patches (atleast the modal patches)
>
i might possibly know that using patches might be buggy and also dangerous, but atleast you shoud've added some API to do this

you've also accepted @SharkPool-SP's Camera V2 extension, which (somewhat) uses patches to create new (builtin) modals.

like maybe add a way to create custom modals to add HTML elements to the modal, or add more types of modals.

SharkPool-SP:
> > so. technically, this extension won't be accepted due to the use of patches? for the other mods support removal i can do that but i'll probably won't remove the patches (atleast the modal patches)
> 
> >
> 
> i might possibly know that using patches might be buggy and also dangerous, but atleast you shoud've added some API to do this.
> 
> 
> 
> you've also accepted @SharkPool-SP's Camera V2 extension, which (somewhat) uses patches to create new (builtin) modals.
> 
> that's kind of unfair.
> 
> 
> 
> like maybe add a way to create custom modals to add HTML elements to the modal, or add more types of modals.

Camera V2 was merged to fix camera v1. And the only way to make a proper camera system was with patches.

Additionally, the modal in the camera v2 extension isn't patched. It just calls the function with the appropriate settings.

yuri-kiss:
> but atleast you shoud've added some API to do this.
no, thats not how that works, we dont cater to every single possible need some random developer may have
if its useful, and someone wants it as an API, it can be added (WITH TIME)

SharkPool-SP:
For the sake of things, this should be more of an Addon than an extension, this does nothing outside of the editor... 

This PR has also been stale for a good couple months (and its still a draft).

Im closing this, consider making it an addon.

PR #2029, opened 2025-03-11
Require Scratch.Cast in ESLint

Brackets-Coder:
This is the first time I've done something like this, and I was just following the patterns already in the file and in #1855, so I hope I did this right.

I don't think this will get merged, I probably did something dumb somewhere and I haven't tested this...

Anyway, I added a rule to eslint.config.js to return an ESLint error if using `Cast` instead of `Scratch.Cast` for all future extensions. Don't know if it will work.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -164,6 +164,16 @@ module.exports = [
               message: "Use Scratch.fetch() instead of window.fetch()",
             },
           ]),
+          "use-scratch-cast": createQueryRule([
+            {
+              selector: "MemberExpression[object.name=Cast]",
+              message: "Use Scratch.Cast instead of Cast",
+            },
+            {
+              selector: "MemberExpression[object.name=window][property.name=Cast]",
+              message: "Use Scratch.Cast instead of window.Cast",
+            },
+          ]),
           "use-scratch-open-window": createQueryRule([
             {
               selector: "CallExpression[callee.name=open]",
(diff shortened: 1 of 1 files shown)


Brackets-Coder:
Well, I guess it works, because I forgot to add ESLint ignores for the new rules to all of the old extensions and now they all return errors every time Cast is used instead of Scratch.Cast. Oops, that'll need to be fixed. I don't feel like going through all of the extensions right now so I'll probably get around to the Lint errors later.

This should make new extensions switch to `Scratch.Cast` for best practices.

GarboMuffin:
Aliasing a local copy of Cast is perfectly fine..

Brackets-Coder:
> Aliasing a local copy of Cast is perfectly fine..

I'm aware of this, and that's one of the biggest oversights of this PR, but is there a way we can check if a file contains just `Cast` but _not_ `Scratch.Cast` (and therefore does not use Cast properly) and return an error? Many extensions use something like 
`const Cast = Scratch.Cast` but if Scratch.Cast isn't present in the entire file but a `Cast` is then it should return an error.

GarboMuffin:
That is already an ESLint error because then Cast is not defined and ESLint should be able to detect this already. It will also be caught during any reasonable testing process because that block would encounter a ReferencError.

GarboMuffin:
In general if we're adding a new ESLint rule there should be an example somewhere of incorrect code that it produces a valid warning for. The change here finds errors in https://github.com/TurboWarp/extensions/blob/3dc6a5a76550346452beeccfd0df799f2d948156/extensions/Lily/Assets.js and https://github.com/TurboWarp/extensions/blob/3dc6a5a76550346452beeccfd0df799f2d948156/extensions/JeremyGamer13/tween.js both of which have `const Cast = Scratch.Cast;` which indicates that the problem of using Cast without defining it first doesn't really exist

PR #2030, opened 2025-03-12
WebSocket V2 Extension

BludIsAnLemon:
made this extension because why not allow more than one websocket connections


BludIsAnLemon:
!!format

BludIsAnLemon:
> here before somebody says "it should be merged with the original websocket extension"

This extension has a different system from the original

CubesterYT:
First off, if it's different entirely from the first one, just name this one Websockets v2

BludIsAnLemon:
> First off, if it's different entirely from the first one, just name this one Websockets v2
Alrighty

BludIsAnLemon:
Renamed it to WebSocket V2

Issue #2033, opened 2025-02-28
Project volume doesn't affect video volume


capyisslay:
When you change the project volume through the slider, the video volume (video extension) doesn't change at all

PR #2034, opened 2025-03-13
Lily/Video: Update volume functions to interact with the project volume

NexusKitten:
Resolves https://github.com/TurboWarp/extensions/issues/2033

Let the video's volume be affected by the project's volume (which can be manipulated through Scratch Addons and Sound Expanded).

The "volume of video" reporter has been updated to only return the video's personal volume, and does not take into account the project's volume.

https://github.com/user-attachments/assets/31dc0dca-23c2-488e-8ca0-a4725315e845

### DIFF
--- a/extensions/Lily/Video.js
+++ b/extensions/Lily/Video.js
@@ -151,13 +151,18 @@
       // Updated code to apply the project's volume to that of the video.
       // This allows the volume to interact with Scratch Addons and Sound Expanded.
       runtime.on("AFTER_EXECUTE", () => {
+        this.updateAudio();
+      });
+
+      this.updateAudio = function () {
         for (const skin of renderer._allSkins) {
           if (skin instanceof VideoSkin) {
             let projectVolume = runtime.audioEngine.inputNode.gain.value;
             skin.videoElement.volume = skin.videoVolume * projectVolume;
+            console.log(skin.videoElement.volume);
           }
         }
-      });
+      }
 
       runtime.on("RUNTIME_PAUSED", () => {
         for (const skin of renderer._allSkins) {
@@ -696,6 +701,7 @@
 
       const value = Cast.toNumber(args.VALUE);
       videoSkin.videoVolume = Math.min(1, Math.max(0, value / 100));
+      this.updateAudio();
     }
 
     setPlaybackRate(args) {
--- a/extensions/Lily/Video.js
+++ b/extensions/Lily/Video.js
@@ -42,6 +42,8 @@
       /** @type {string} */
       this.videoSrc = videoSrc;
 
+      this.videoVolume = 1;
+
       this.videoError = false;
 
       this.readyPromise = new Promise((resolve) => {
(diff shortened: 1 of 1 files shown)


SharkPool-SP:
I feel like it would be more efficient to iterate through all created videos rather than all skins

GarboMuffin:
It looks like right now the video element's volume doesn't actually change until end of frame. If the project's scripts get stuck for a while that might mean the volume blocks are now noticeably delayed, until the scripts are unstuck

SharkPool-SP:
> It looks like in this pull request the video element's volume doesn't actually change until end of frame. If the project's scripts get stuck for a while that might mean the volume blocks are now noticeably delayed, until the scripts are unstuck

wouldnt before execute work

GarboMuffin:
No that would another frame of delay

Old version would update element volume in middle of frame, not on either end.  Not sure if browsers actually process those changes mid frame or if they end up waiting until the end anyways

GarboMuffin:
The situation I described does not involve the addon at all

GarboMuffin:
Although yes without an event the slider will always be a tiny bit delayed

Another option is a making media source node and connecting it to the audio engine. That could help make autoplay more reliable

yuri-kiss:
> The situation I described does not involve the addon at all

I know. The point of using a event would mean you are not waiting for a frame to finish or start, and only when the volume actually updates.

NexusKitten:
!format

GarboMuffin:
yes that solves the project i was talking about once you remove the leftover console.log

extensions/Lily/Video.js
```diff
@@ -146,6 +148,21 @@
         }
       });
 
+      // Updated code to apply the project's volume to that of the video.
+      // This allows the volume to interact with Scratch Addons and Sound Expanded.
+      runtime.on("AFTER_EXECUTE", () => {
+        this.updateAudio();
+      });
+
+      this.updateAudio = function () {
```
GarboMuffin, line 157: is there a reason this is defined here and not in a normal class method

extensions/Lily/Video.js
```diff
@@ -146,6 +148,21 @@
         }
       });
 
+      // Updated code to apply the project's volume to that of the video.
+      // This allows the volume to interact with Scratch Addons and Sound Expanded.
+      runtime.on("AFTER_EXECUTE", () => {
```
GarboMuffin, line 153: I think this will slowly leak memory if you don't remove the listener later

yuri-kiss: it will

NexusKitten:
!format

extensions/Lily/Video.js
NexusKitten: i think i fixed this, i'm not entirely sure though

NexusKitten:
!format

extensions/Lily/Video.js
```diff
@@ -31,6 +31,17 @@
   elementContainer.ariaHidden = "true";
   document.body.appendChild(elementContainer);
 
+  // Updated code to apply the project's volume to that of the video.
+  // This allows the volume to interact with Scratch Addons and Sound Expanded.
+  const updateAudio = function() {
```
yuri-kiss: iterating every time even when you have the skin seems kinda of wasteful but other than that it looks "good" I guess. I still think an even should be made for this instead

GarboMuffin (approved):

yuri-kiss:
this is insanely resource in-effecient but since the idea for volume changes to become an event is not happening we may aswell merge this in its current state.

yuri-kiss (approved):

GarboMuffin:
yes its not ideal but we already have this O(n) loop so its not regressing much

PR #2040, opened 2025-03-18
Lily/MoreEvents: added "whenProjectPaused" and "whenProjectUnPause"

The11thPencil:


fath11:
!format

extensions/Lily/MoreEvents.js
```diff
@@ -205,6 +212,33 @@
...
       };
+
+      let pauseHandler = () => {
+        runtime.allScriptsByOpcodeDo(
+          "lmsMoreEvents_whenProjectPaused",
+          (script, target) => {
+            const topBlockId = script.blockId;
+            for (let j = 0; j < runtime.threads.length; j++) {
+              if (
+                runtime.threads[j].topBlock === topBlockId &&
```
GarboMuffin, line 223: This does not account for clones properly. The script is only going to run in one of the clones at a time since theyll have the same top block ID

extensions/Lily/MoreEvents.js
```diff
@@ -205,6 +212,33 @@
...
+              }
+            }
+            runtime._pushThread(topBlockId, target);
+          }
+        );
+        runtime.removeListener("BEFORE_EXECUTE", pauseHandler);
+      };
+
+      runtime.on("RUNTIME_PAUSED", () => {
+        runtime.on("BEFORE_EXECUTE", pauseHandler);
```
GarboMuffin, line 236: You can use .once instead of .on and it will only fire once. Easier to understand than using .on with .removeListener

extensions/Lily/MoreEvents.js
```diff
@@ -205,6 +212,33 @@
...
+      });
+
+      runtime.on("RUNTIME_UNPAUSED", () => {
+        runtime.startHats("lmsMoreEvents_whenProjectUnPaused");
```
GarboMuffin, line 240: It's possible for someone to pause then unpause in the same frame (eg.  if they lower the framerate).

It seems like what will happen here is that because RUNTIME_PAUSED handler until BEFORE_EXECUTE to start threads while the RUNTIME_UNPAUSED handler does startHats immediately, the when unpaused scripts will run before the when paused scripts, which is backwards. I guess they should both be in BEFORE_EXECUTE to ensure consistent ordering

yuri-kiss:
why was this closed

fath11:
wait why is this closed, i didnt meant to close it lol

extensions/Lily/MoreEvents.js
fath11: i am not quite sure how to fix this..

LilyMakesThings:
I'm not a big fan of doing this as it is interfacing with a feature that is introduced through an addon and nothing internally. I'd rather it be introduced into a separate extension like Runtime Options, not here.

Brackets-Coder:
!format

Brackets-Coder:
@fath11 this is similar to (though not an exact duplicate of) #1245. According to #2271, we should do our best to avoid similar/duplicate pull requests by encouraging collaboration instead of "one is better than the other". Since @LilyMakesThings would prefer this kind of functionality in a separate extension and #1245 already has this functionality, @fath11 are you okay if I close this? You could always collaborate with @SharkPool-SP to get this extra pause functionality to Turbowarp

The11thPencil:
Yes I am okay if this is closed

Brackets-Coder:
Okay, hope this won't upset you

PR #2045, opened 2025-03-22
Lily/AllMenus: mark as scratch-compatible

GarboMuffin:


hammouda101010:
excuse me **w h a t**?

CST1229:
> excuse me **w h a t**?

the All Menus extension doesn't add any blocks, therefore it doesn't need the "incompatible with scratch" prompt because it doesn't actually save into the project (since it doesn't add any blocks!)

CST1229:
> > > excuse me **w h a t**?
> > 
> > 
> > the All Menus extension doesn't add any new blocks, therefore it doesn't need the "incompatible with scratch" prompt because it doesn't actually save into the project (since it doesn't add any blocks!)
> 
> but is it allowed in Scratch or not?

technically [TurboWarp itself is not allowed at all](https://docs.turbowarp.org/4.4), but the ST doesn't seem to care and you can achieve the same results by editing the project.json file anyways

PR #2057, opened 2025-03-29
SharkPool/Camera: Clones should inherit camera from parent, add epsilon

SharkPool-SP:
Requested by Cubester
image

I also added a epsilon for float point issues and some minor optimizations

### DIFF
--- a/extensions/SharkPool/Camera.js
+++ b/extensions/SharkPool/Camera.js
@@ -256,6 +256,7 @@
   const ogInitDrawable = vm.exports.RenderedTarget.prototype.initDrawable;
   vm.exports.RenderedTarget.prototype.initDrawable = function (layerGroup) {
     ogInitDrawable.call(this, layerGroup);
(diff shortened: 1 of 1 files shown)


SharkPool-SP:
!format

GarboMuffin:
Would it make sense to wrap most of the new function in `if (!this.isOriginal) {`, otherwise it is doing some unnecessary works when the original sprites load

SharkPool-SP:
> Would it make sense to wrap most of the new function in `if (!this.isOriginal) {`, otherwise it is doing some unnecessary works when the original sprites load

true

SharkPool-SP:
> Would it make sense to wrap most of the new function in `if (!this.isOriginal) {`, otherwise it is doing some unnecessary works when the original sprites load

fixed

SharkPool-SP:
!format

GarboMuffin (approved):

CubesterYT (approved):

PR #2061, opened 2025-03-31
Docs markdown alerts: case insensitive

Brackets-Coder:
title says it all

Also, putting HTML inside them kinda works, but not as intended. I'll fix that eventually


Brackets-Coder:
Oops wrong hang on

PR #2074, opened 2025-04-08
Vectors

unknown07724:
Vectors is a extension for [Vectors](https://en.wikipedia.org/wiki/Vector_%28mathematics_and_physics%29)
Blocks: Create vector with ID,direction and magnitude: creates a vector, vector direction, returns direction of the vector based on ID, vector magnitude, returns magnitude of the vector based on ID, Change Magnitude and direction, changes magnitude and direction of the vector based on ID, and List of IDs, lists all elements in IDs


fath11:
there is no block for deleting vectors?

fath11:
also please run !format

unknown07724:
!format

unknown07724:
> there is no block for deleting vectors?

Yes, just stop using a vector

unknown07724:
!format

unknown07724:
!format

unknown07724:
!format

unknown07724:
@fath11 there is now

extensions/unknown/vectors.js
```diff
@@ -0,0 +1,169 @@
+(function (Scratch) {
```
yuri-kiss, line 7: you're missing the header data

extensions/unknown/vectors.js
```diff
@@ -0,0 +1,169 @@
...
+      return {
+        id: "unknownvectors",
+        name: Scratch.translate("Vectors"),
+        color1: "#3495eb",
+        blocks: [
+          {
+            opcode: "vectorCreate",
+            blockType: Scratch.BlockType.COMMAND,
+            text: Scratch.translate(
+              "create vector with ID [ID], direction [DIRECTION] and magnitude [MAGNITUDE]"
```
yuri-kiss: you have inconsistent capitalization across the blocks, the blocks should be written starting with lowercase. they are NOT sentences.

extensions/unknown/vectors.js
```diff
@@ -0,0 +1,169 @@
...
+            arguments: {
+              ID: { type: Scratch.ArgumentType.STRING, defaultValue: "vec1" },
+            },
+          },
+        ],
+      };
+    }
+
+    vectorCreate(args) {
+      const id = args.ID;
```
yuri-kiss: you don't cast anything, this is poor practice

extensions/unknown/vectors.js
```diff
@@ -0,0 +1,169 @@
+(function (Scratch) {
+  "use strict";
+
+  // Ensuring the extension is unsandboxed
+  if (!Scratch.extensions.unsandboxed) {
+    throw new Error("This extension must run unsandboxed");
+  }
+
+  // Vectors storage
+  const ids = [];
```
yuri-kiss, line 18: why can't this be a set

extensions/unknown/vectors.js
```diff
@@ -0,0 +1,169 @@
...
+  const directions = {};
```
yuri-kiss, line 19: these should be Object.create(null); or a Map

extensions/unknown/vectors.js
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    vectorChangeMag(args) {
+      const id = args.ID;
+      magnitudes[id] = args.MAGNITUDE;
+    }
+
+    vectorMag(args) {
+      const id = args.ID;
+      return magnitudes[id] ?? Scratch.translate("vector not found");
```
yuri-kiss: you should return an empty string or 0 here

extensions/unknown/vectors.js
```diff
@@ -0,0 +1,169 @@
...
+      if (index !== -1) {
+        ids.splice(index, 1);
+        delete directions[id];
+        delete magnitudes[id];
+      }
+    }
+
+    vectorDir(args) {
+      const id = args.ID;
+      return directions[id] ?? Scratch.translate("vector not found");
```
yuri-kiss: same here

yuri-kiss (changes requested):
this seems to literally just be a weird way of storing 2 dictionary's, what is the point of this extension, you can't even do any operations?

unknown07724:
Maps are bullshit, it does what it needs to do very well

extensions/unknown/vectors.js
unknown07724: shit, Imma re-add that Ig

unknown07724:
!format

GarboMuffin:
it would be wonderful if you would be more reasonable with your language on this issue tracker

unknown07724:
it already does what it needs to do, I don't want to spend multiple hours rewriting the entire extension

extensions/unknown/vectors.js
unknown07724: do I want people to wonder why their shit isn't working? if you said yes then you are wrong, the answer is no.

extensions/unknown/vectors.js
unknown07724: aren't they numbers, you can not put strings in there, and if you could, who would?

extensions/unknown/vectors.js
Brackets-Coder: > do I want people to wonder why their shit isn't working? if you said yes then you are wrong, the answer is no.

@GarboMuffin said to chill the language

extensions/unknown/vectors.js
unknown07724: I see now, attempted to fix, fucked up the code

unknown07724:
!format

extensions/unknown/vectors.js
unknown07724: fixed the fuck up, now works better

extensions/unknown/vectors.js
unknown07724: decided upon making it "null"

extensions/unknown/vectors.js
```diff
@@ -0,0 +1,180 @@
...
+    vectorChangeMag(args) {
+      const id = Scratch.Cast.toString(args.ID);
+      magnitudes[id] = Scratch.Cast.toNumber(args.MAGNITUDE);
+    }
+
+    vectorMag(args) {
+      const id = Scratch.Cast.toString(args.ID);
+      return magnitudes[id] !== undefined
+        ? magnitudes[id]
+        : Scratch.translate("null");
```
CST1229: translating block return values is usually not a good idea; people may compare the result of this block to the string `null` (or whatever it is in their language) and then have it break in other languages that have a different string for it.
(`null` specifically is a fairly language-neutral word so it should be fine untranslated. or you could follow Scratch's precedent and return an empty string; this is what lists also do for example so I don't think it would be confusing)

unknown07724:
!format

unknown07724:
!format

unknown07724:
> this seems to literally just be a weird way of storing 2 dictionary's, what is the point of this extension, you can't even do any operations?

you can now

extensions/unknown/vectors.js
unknown07724: fixed

PR #2082, opened 2025-04-18
Lily/ClonesPlus.js -- Use name-based system to prevent showing variable Ids

SharkPool-SP:
stops that weird text jumble from the variable ID when it no longer exists or is duplicated

### DIFF
--- a/extensions/Lily/ClonesPlus.js
+++ b/extensions/Lily/ClonesPlus.js
@@ -447,7 +447,9 @@
       );
       const clones = util.target.sprite.clones;
       const cloneNum = clones.length - 1;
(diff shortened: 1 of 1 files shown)


SharkPool-SP:
!format

GarboMuffin:
Then it breaks when a variable gets renamed

SharkPool-SP:
> Then it breaks when a variable gets renamed

Well there's nothing really else we can do. Pick your poison

GarboMuffin:
I will fix it myself later then

yuri-kiss:
> I will fix it myself later then

preferably explain how because this is a much larger issue at hand...

LilyMakesThings:
You're also going to break literally every project that uses clonesplus by doing something like this.

PR #1935, opened 2025-02-14
Create indexeddb.js

kilgorezer:

PR #1613, opened 2024-07-15
SQLSnap! Fileshare

Bitter130:
An extension that allows for file storage online.
image

PR #1341, opened 2024-03-02
Add IndexedDB Extension

ffghdfh:
lemme in please 🥺

PR #2087, opened 2025-04-21
Create firebasedb.js

Logise1123:


extensions/firebasedb.js
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        getInfo() {
+            return {
+                id: "FirebaseDB",
+                name: "Firebase DB",
+                color1: "#fea631",
+                menuIconURI: icon,
+                blocks: [
+                    {blockType: Scratch.BlockType.LABEL, text: "Made by @logise on Discord"},
```
Brackets-Coder, line 25: Just for consistent formatting why can't you open this structure up like the other blocks?

extensions/firebasedb.js
```diff
@@ -0,0 +1,187 @@
...
+
+    class FirebaseDB {
+        constructor() {
+            this.api = "https://guessthepin-2fe64-default-rtdb.europe-west1.firebasedatabase.app";
+        }
+
+        getInfo() {
+            return {
+                id: "FirebaseDB",
+                name: "Firebase DB",
```
Brackets-Coder, line 21: To fix all of the lint errors you can just use `Scratch.translate()` on the extension and block names.

extensions/firebasedb.js
```diff
@@ -0,0 +1,187 @@
...
+                color1: "#fea631",
+                menuIconURI: icon,
+                blocks: [
+                    {blockType: Scratch.BlockType.LABEL, text: "Made by @logise on Discord"},
+                    {
+                        opcode: "setKey",
+                        blockType: Scratch.BlockType.COMMAND,
+                        text: "set key [KEY] to value [VALUE]",
```
Brackets-Coder, line 29: you should also translate the block text

extensions/firebasedb.js
```diff
@@ -0,0 +1,187 @@
...
+                        arguments: {
+                            PASSWORD: { type: Scratch.ArgumentType.STRING, defaultValue: "password" },
+                            KEY: { type: Scratch.ArgumentType.STRING, defaultValue: "key" }
+                        }
+                    }
+                ]
+            };
+        }
+
+        async delay() {
```
Brackets-Coder, line 76: Lint is returning `Async method 'delay' has no 'await' expression`. Avoid putting `async` before the extension's block functions, rather just put an asynchronous function inside them

extensions/firebasedb.js
```diff
@@ -0,0 +1,187 @@
...
+            const d = Math.random() * 500;
+            return new Promise(r => setTimeout(r, d));
+        }
+
+        async setKey(args) {
+            await this.delay();
+            const { KEY, VALUE } = args;
+            if (VALUE.length > 8000) return;
+            await fetch(`${this.api}/pin/${encodeURIComponent(KEY)}.json`, {
```
Brackets-Coder, line 85: Use Scratch.fetch() instead of fetch()

extensions/firebasedb.js
```diff
@@ -0,0 +1,187 @@
...
+                    salt,
+                    iterations: 100000,
+                    hash: "SHA-256"
+                },
+                keyMaterial,
+                { name: "AES-GCM", length: 256 },
+                false,
+                ["encrypt", "decrypt"]
+            );
+        }
```
Brackets-Coder, line 116: It seems like all of the block functions are async for interacting with the database server. Is it necessary to do it like this or can you put an asynchronous function inside all of them? Also, you might want to include documentation noting this because I don't remember if the script pauses until each async block finishes executing or continues while the asyncs are still executing

extensions/firebasedb.js
```diff
@@ -0,0 +1,187 @@
...
+
+            await fetch(`${this.api}/cypher/${encodeURIComponent(KEY)}.json`, {
+                method: "PUT",
+                body: JSON.stringify(fullPackage)
+            });
+        }
+
+        async getKeyWithPassword(args) {
+            await this.delay();
+            const { KEY, PASSWORD } = args;
```
Brackets-Coder, line 145: Generally speaking you should use Scratch.Cast() on all of your arguments to avoid weird quirks... this goes for all blocks too

Brackets-Coder (commented):
This is just a quick review going over TW extension best practices, I don't have time to do a full in-depth functionality review. To wrap up: use `Scratch.translate` and `Scratch.Cast` where necessary.

Brackets-Coder:
!format

CubesterYT:
!format

extensions/firebasedb.js
```diff
@@ -5,183 +5,230 @@
...
+
+  class FirebaseDB {
+    constructor() {
+      this.api =
+        "https://guessthepin-2fe64-default-rtdb.europe-west1.firebasedatabase.app";
```
Brackets-Coder, line 18: If this references an external domain, it likely won't be merged. Here's a few reasons:

1. If the domain ever goes down, real projects don't work any more -- there's no guarantee of future compatibility
2. Some users may be on a whitelisted network that blocks this anyway.

I'd recommend closing this PR and working with the already open #1341, which uses a browser API for a more compatible permanent storage solution.

SharkPool-SP:
As nice of an extension this is, Im closing it for several reasons:
- The API url is from an external domain, which I presume is owned by you. If the domain ever goes down, the extension doesnt work -- there's no guarantee of future compatibility.
- Theres other PRs for storage including Local Storage, [IndexDB](https://github.com/TurboWarp/extensions/pull/1935), and [SQL Snap! FileShare](https://github.com/TurboWarp/extensions/pull/1613) (Which is hosted and maintained by Snap!, which is arguably more trustworthy of not going down anytime soon)
- Theres no option for custom domains (if the user wants to put data in their own FireBase DB), but even then that risks the possibility of leaking API keys in public projects.

PR #2110, opened 2025-05-05
Lily/ClonesPlus: revert breaking change

TheShovel:
https://github.com/TurboWarp/extensions/commit/5825ea724fe303963adb8aa4af75ce75094bd38f
This change breaks projects made with previous versions of the extension. This is was a change in functionality!

One of my projects is affected by this change, because it is built with clones starting with empty local variables in mind. This should be handled differently. I dont know how this was approved. You cant change stuff like this after such a long time of them working this way, without making sure that old projects are unaffected

### DIFF
--- a/extensions/Lily/ClonesPlus.js
+++ b/extensions/Lily/ClonesPlus.js
@@ -442,7 +442,7 @@
     createCloneWithVar(args, util) {
       // @ts-expect-error - not typed yet
       Scratch.vm.runtime.ext_scratch3_control._createClone(
-        "_myself_",
+        util.target.sprite.name,
         util.target
       );
       const clones = util.target.sprite.clones;


TheShovel:
There are rules put in place for these things
https://docs.turbowarp.org/development/extensions/compatibility#what-if-you-need-to-break-compatibility
https://docs.turbowarp.org/development/extensions/compatibility#dont-significantly-change-block-behavior

SharkPool-SP:
Well now this ‘breaks’ projects using the new change

GarboMuffin:
this is a very unfortuante situation but probably since there hasnt been a desktop app update with this, it can be feasibly reverted, ...

yuri-kiss (approved):
I don't see why not, we can make a separate pull request to make this a new block :)

GarboMuffin (approved):

PR #2126, opened 2025-05-21
Add "Blocos de Espera" extension for advanced timing

DJ-15-DJS:
This extension is called "Blocos de Espera" and adds new wait blocks with support for milliseconds, seconds, minutes, and hours.

✨ Features:
- Wait for milliseconds
- Wait for seconds
- Wait for minutes
- Wait for hours

It helps projects that require more precise or extended timing beyond the default wait block.

Tested and working correctly.

Thank you for reviewing this!


yuri-kiss:
this is redundant. you can just use simple math.

Issue #2127, opened 2025-05-23
Skins has horrible state management


yuri-kiss:
there is a few race conditions that lead to bad state mangement that breaks the whole extension

![Image](https://github.com/user-attachments/assets/adea7875-7a79-4e59-b6e1-f20adad1d549)

![Image](https://github.com/user-attachments/assets/2e150f64-63e2-4444-8fce-0f97e9075145)

PR #2128, opened 2025-05-23
Lily/Skins: Fix some fatal loading bugs.

yuri-kiss:
This causes a lot of errors if it happens repeatedly and can lead to project corruption if handled incorrectly.
This fixes some of thoes loading bugs. fixes: #2127

The issue arises from the fact that you can rename / delete / cancel the loading of a costume / skin while its still loading, and the renderer throws and error if that happens at a bad time, Skins doesnt handle this, this aims to fix that issue

### DIFF
--- a/extensions/Lily/Skins.js
+++ b/extensions/Lily/Skins.js
@@ -17,30 +17,60 @@
     return true;
   };
 
-  /**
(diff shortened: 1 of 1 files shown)


yuri-kiss:
!format

PR #2142, opened 2025-06-04
Xeltalliv/clippingblending: Fix canvas alpha compatibility and other changes

Xeltalliv:
This:
* Fixes compatibility with canvas having alpha, without changing behavior of  _any_ old projects compared to how they used to work (at least in theory).
* Sorts out long standing issue of some blending modes working incorrectly, by deprecating old block and making a new one, where everything is done in a corrected and clean way.
* Adds new blending modes, to provide proper solutions to what was done through bugs.
* Adds API, so that [Shaded](https://pen-group.github.io/extensions/) extension can be compatible with this extension, without the need for maintaining a custom fork of it.
* Adds a label explaining why there is sometimes no blocks in the palette.
image

### DIFF
--- a/extensions/Xeltalliv/clippingblending.js
+++ b/extensions/Xeltalliv/clippingblending.js
@@ -383,6 +383,7 @@
               },
             },
             filter: [Scratch.TargetType.SPRITE],
+            hideFromPalette: true,
             extensions: ["colours_looks"],
           },
           {
--- a/extensions/Xeltalliv/clippingblending.js
+++ b/extensions/Xeltalliv/clippingblending.js
@@ -40,6 +40,23 @@
   let scratchUnitHeight = 360;
   let penDirty = false;
 
+  // prettier-ignore
+  const Blendings = Object.assign(Object.create(null), {
+    "default": [gl.ONE, gl.ONE_MINUS_SRC_ALPHA, gl.ONE, gl.ONE_MINUS_SRC_ALPHA, gl.FUNC_ADD],
+    "default behind": [gl.ONE_MINUS_DST_ALPHA, gl.ONE, gl.ONE_MINUS_DST_ALPHA, gl.ONE, gl.FUNC_ADD],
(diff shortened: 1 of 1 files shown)


SharkPool-SP:
Instead of the function for getting the drawable you could just use the Drawable class in renderer.exports

Xeltalliv:
Um, that's exactly what I did

image

https://github.com/TurboWarp/extensions/blob/e694b07487d2d22ef96c0cd86a22ebc08c987e6d/extensions/Xeltalliv/clippingblending.js#L37
but I kept the old one as a fallback, in case this extension was used in some obscure outdated turbowarp mods without renderer.exports.

extensions/Xeltalliv/clippingblending.js
```diff
@@ -32,8 +32,9 @@
   const runtime = vm.runtime;
   const renderer = vm.renderer;
   const _drawThese = renderer._drawThese;
-  const gl = renderer._gl;
+  const gl = renderer.gl;
   const canvas = renderer.canvas;
+  let Drawable = renderer?.exports?.Drawable || getDrawable();
```
yuri-kiss: you shouldnt use `?.` btw (see garbos nitpick pr on tempvars2)

yuri-kiss (dismissed):

extensions/Xeltalliv/clippingblending.js
Xeltalliv: After considering that this code now also uses `BlockType.LABEL`, I decided to remove support for those old versions of TurboWarp, even though the original [merge](https://github.com/TurboWarp/extensions/commit/72f732ed517aec343bc34103a5717845aec1b268) of this extension predates both [renderer.exports](https://github.com/TurboWarp/scratch-render/commit/19f057065d99624a0c8b76f4461d08b6341ad630) and [`BlockType.LABEL`](https://github.com/TurboWarp/scratch-vm/commit/8a7ae3f7741bd12da9899ff00d48afb9b2b37c00) by 2 months.

SharkPool-SP:
Yuri has to approve because she made requests. I can’t merge :(

PR #2148, opened 2025-06-04
HTTP

wiktorlaskowski:
NEW HTTP ICON✅
http


yuri-kiss:
I actually kind of like this

yuri-kiss:
could you find a way to make it look like it has smth to do with well networking, and update the image credits please

SharkPool-SP (approved):

PR #2164, opened 2025-06-13
cloudlink: make every (non-localhost) server ID behave like 7

CST1229:
this is a bit of a "PR made slightly out of anger".

"use server 7" is almost memetic in the TW server, and some people think CL is broken because people try to connect to e.g server ID 1 and it doesn't work. 7 is the only server left anyways.

(0 still remains as localhost though)

### DIFF
--- a/extensions/cloudlink.js
+++ b/extensions/cloudlink.js
@@ -2178,6 +2178,10 @@
         console.warn("[CloudLink] Already connected to a server.");
         return;
       }
+      // This is the only server that's listed and works.
+      if (Scratch.Cast.toNumber(args.ID) >= 1) {
+        args.ID = 7;
+      }
       if (
         !Object.prototype.hasOwnProperty.call(
           clVars.serverList,


GarboMuffin:
I was hoping making 7 be the default value and only server in the list would be enough but I think this is a good idea too

Brackets-Coder:
is there a way to connect to wss://clouddata.turbowarp.org? I'm assuming it's a problem with the server's handshake rather than the extension but it's returning a 4000 error.

CST1229:
> is there a way to connect to wss://clouddata.turbowarp.org? I'm assuming it's a problem with the server's handshake rather than the extension but it's returning a 4000 error.

i don't think you can connect to cloud variable servers with cloudlink (only cloudlink servers with a cloud variable client)

Brackets-Coder:
> > is there a way to connect to wss://clouddata.turbowarp.org? I'm assuming it's a problem with the server's handshake rather than the extension but it's returning a 4000 error.
> 
> i don't think you can connect to cloud variable servers with cloudlink (only to cloudlink servers with a cloud variable client)

Will something like this be possible with two-way Web Socket connections that were discussed as a possibility in other pull requests and issues?

PR #1058, opened 2023-09-24
Add Sound Waves Extension

SharkPool-SP:
Play oscillators in your project!

Screenshot 2023-09-24 at 10 08 28 AM

Update:
Screenshot 2024-01-28 at 1 05 16 AM

PR #2187, opened 2025-07-01
Add oscillators extension

alctd:
Added the oscillators extension

### DIFF
--- a/extensions/Amp/oscillators.js
+++ b/extensions/Amp/oscillators.js
@@ -1,0 +1,87 @@
+// Name: Oscillators
+// ID: oscillator
+// Description: Synthensize simple sound waves.
+// By: AmpElectrecuted <https://scratch.mit.edu/users/AmpElectrecuted/>
+// License: MPL-2.0
+
+(function (Scratch) {
+  'use strict';
+  let context = new AudioContext();
+  let oscillator = new OscillatorNode(context);
+  oscillator.connect(context.destination);
+  let isStart = false;
+  let isPaused = true;
+
+  class OscillatorExtension {
+    getInfo() {
+      return {
+        id: 'oscillator',
+        name: Scratch.translate('Oscillators'),
+        blocks: [
+          {
+            opcode: 'startFreq',
+            blockType: Scratch.BlockType.COMMAND,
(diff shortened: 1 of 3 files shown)


alctd:
!format

Brackets-Coder:
> @Brackets-Coder

Is there something you would like me to do?

extensions/Amp/oscillators.js
Brackets-Coder, line 1: Looks good, you're consistent with IDs and translating so that's great. Just need a few casting things and I guess you're done.

It's a very small extension, but as @SharkPool-SP mentioned earlier there's already an open pull request for a very similar extension. It would be better if both could be merged into one extension or if one were chosen over the other.

extensions/Amp/oscillators.js
```diff
@@ -0,0 +1,87 @@
...
+                value: "triangle",
+              },
+            ],
+          },
+        },
+      };
+    }
+
+    startFreq(args) {
+      oscillator.type = args.WAVEFORM;
```
Brackets-Coder: It's a menu, but it accepts reporters so you should still cast to string
```suggestion
      oscillator.type = Scratch.Cast.toString(args.WAVEFORM);
```

extensions/Amp/oscillators.js
```diff
@@ -0,0 +1,87 @@
...
+      oscillator.frequency.setValueAtTime(args.FREQ, context.currentTime);
```
Brackets-Coder: You should cast `freq` to a number

```suggestion
      oscillator.frequency.setValueAtTime(Scratch.Cast.toNumber(args.FREQ), context.currentTime);
```

Brackets-Coder:
I don't think that review will help you much because I'm not a moderator so I can't approve anything

extensions/Amp/oscillators.js
Brackets-Coder, line 1: Looks fine to me

Brackets-Coder (commented):
Yep, all looks good. The one thing I'd say is don't use PenguinFork:master branch, you should have different branches for each Pull Request/Extension to help separate commits and organize things better 😉 

Other than that, I'd say it's good enough. But you might want to collab with #1058, as it's a duplicate PR and you and @SharkPool-SP both want the same extension on the gallery

PR #2202, opened 2025-08-04
Add min and max block to true-fantom/math.js

LSPECTRONIZTAR:
Not only do we need this, but I literally don't know an official extension in the gallery that has this block.

block_8_4_2025-2_30_53 PM

### DIFF
--- a/extensions/true-fantom/math.js
+++ b/extensions/true-fantom/math.js
@@ -639,7 +639,11 @@
       return 0 - cast.toNumber(A);
     }
     min_or_max_block(args) {
-      if (args.MIN_OR_MAX === "min") { return Math.min(args.A, args.B); } else { return Math.max(args.A, args.B); }
+      if (args.MIN_OR_MAX === "min") {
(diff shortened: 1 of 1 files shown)


extensions/true-fantom/math.js
```diff
@@ -614,6 +638,13 @@
     negative_block({ A }) {
       return 0 - cast.toNumber(A);
     }
+    min_or_max_block(args) {
+      if (args.MIN_OR_MAX === "min") {
+        return Math.min(args.A, args.B);
+      } else {
+        return Math.max(args.A, args.B);
+      }
+    }
```
Brackets-Coder, line 647: Just do this:

```suggestion
    min_or_max_block(args) {
        return Math[args.MIN_OR_MAX](args.A, args.B);
    }
```

oh wait just realized accept reporters was true never mind it won't work

This means you should use Scratch.Cast.toString()

LSPECTRONIZTAR:
Forget it, doesn't look like it's ever gonna happen

PR #2204, opened 2025-08-07
Update controlcontrols.svg

wiktorlaskowski:


SharkPool-SP (approved):

Brackets-Coder (changes requested):
I can't decide which one I like better but I can merge this if you want

SharkPool-SP:
> The old one has bigger, clearer icons and the solid non-gradient background distracts less...

we can just remove the gradient

Brackets-Coder:
> > The old one has bigger, clearer icons and the solid non-gradient background distracts less...
> 
> we can just remove the gradient

If we do that I'd be cool but honestly it really doesn't matter to me

SharkPool-SP:
@Brackets-Coder how does it look now

Brackets-Coder (approved):
Much better

PR #2206, opened 2025-08-11
New Extension: SK17

SkyBuilder1717:
SK17 is an encrypting extension, for example, saving games and loading private data in compiled games!
browser_PLyv0lOyRj


extensions/SkyBuilder1717/SK17.js
```diff
@@ -0,0 +1,377 @@
+// Name: SK17
```
Brackets-Coder: Maybe make the name less vague and more topical, like "Encryption"

extensions/SkyBuilder1717/SK17.js
```diff
@@ -0,0 +1,377 @@
...
+// ID: SkyBuilder1717Encryption
+// Description: Adds new encryption format: SK17. Useful for saving games and loading private data in compiled games.
+// By: SkyBuilder1717
+// License: MIT
```
Brackets-Coder, line 5: Why?

extensions/extensions.json
```diff
@@ -98,4 +99,4 @@
   "gamejolt",
   "obviousAlexC/newgroundsIO",
   "Lily/McUtils" // McUtils should always be the last item.
```
Brackets-Coder, line 102: You shouldn't remove the comment...?

images/SkyBuilder1717/SK17.svg
Brackets-Coder, line 1: Images can't contain text.

Brackets-Coder (changes requested):
I tested it and it seems to work. Before I give my approval, I'd like to clarify a few things...

First, you shouldn't use the master branch; try to leave it blank and have different branches for each pull request/extension. This keeps things organized for future contributions. If you're confused and need more information just ask.

Second, see the few short notes below.

extensions/extensions.json
SkyBuilder1717: It's not removing, it's github formatting

images/SkyBuilder1717/SK17.svg
SkyBuilder1717: Why? What about Simple 3D or Turbohook?

extensions/SkyBuilder1717/SK17.js
SkyBuilder1717: Because why not?

extensions/SkyBuilder1717/SK17.js
SkyBuilder1717: It's wrote in ID

images/SkyBuilder1717/SK17.svg
Brackets-Coder: > Why? What about Simple 3D or Turbohook?

text in images can't be translated and it's not good for typography. I think Turbohook was added fairly before that soft requirement and I think there was an argument for keeping it for Simple3D (the banner was rendered with the extension)...

extensions/SkyBuilder1717/SK17.js
Brackets-Coder: > Because why not?

Sorry I was just asking about why you chose MIT instead of MPL-2.0

extensions/SkyBuilder1717/SK17.js
Brackets-Coder: > It's wrote in ID

The name is what appears to users on the gallery, "Encryption" may be less confusing.

```suggestion
// Name: Encryption
```

GarboMuffin:
not opposed to the idea of having blocks that wrap crypto primitives but the description "Useful for saving games and loading private data in compiled games." is probably very misleading since the keys are in the project, so anyone can just unpackage it and get the key out. claims about the security of a system should have a strong basis

PR #2235, opened 2025-09-04
tweaked the background color of search-base in dark mode

The11thPencil:
problem:
image

solution:
image

i am just _slightly_ annoyed

### DIFF
--- a/development/homepage-template.ejs
+++ b/development/homepage-template.ejs
@@ -264,6 +264,12 @@
         transition: box-shadow .15s ease-in-out, box-shadow .15s ease-in-out;
       }
 
+      @media (prefers-color-scheme: dark) {
+        .search-base {
+          background-color: #111;
+        }
+      }
+
       .search-base:focus {
         outline: none;
         box-shadow: 0 0 0 .25rem rgba(236, 90, 84, .25);


development/homepage-template.ejs
```diff
@@ -264,6 +264,12 @@
         transition: box-shadow .15s ease-in-out, box-shadow .15s ease-in-out;
       }
 
+      @media (prefers-color-scheme: dark) {
+        .search-base {
+          background-color: #111;
+        }
+      }
+
```
Brackets-Coder, line 272: What we could do instead (although I don't know how much this will affect other things) is instead of only selecting the one class, delete this whole block and modify the one at the top:


```suggestion
@media (prefers-color-scheme: dark) {
        :root {
          color: #eee;
          background: #111;
        }
        * {
          background: #111;
        }
        a {
          color: #56b2ff;
        }
        a:visited {
          color: #f192ff;
        }
        a:active {
          color: red;
        }
        .invert-dark {
          filter: invert(100%);
        }
      }
```

Brackets-Coder: but I don't know if the asterisk selector works like that or how much it will break other things, it's just a thought

PR #2241, opened 2025-09-14
Simple3D: Add translations and make a small refactor

### LINKED ISSUE #2184: Add multi-language support for the Simple 3D extension
The "Simple 3D" extension hasn't added a "Scratch.translate.setup" part so far, so I can't translate this extension on Transifex. Can you chaps add it?

Brackets-Coder:
Fixes #2184

Didn’t update this:
```
      powersOfTwo: {
        acceptReporters: true,
        items: ["1", "2", "4", "8", "16"],
      },
```
I didn’t think arabic numerals had to be translated.
I haven’t yet translated the function-based (I think they’re dynamic) menus at the bottom of the code.

**I also found one of the block definitions had a chain of if statements that could be refactored into a switch, so I did that too-- let me know if I wasn't supposed to or if it broke anything.**

I haven't tested this, so if you guys find anything I missed please let me know because I did this really quickly and kinda skimmed it.

### DIFF
--- a/extensions/Xeltalliv/simple3D.js
+++ b/extensions/Xeltalliv/simple3D.js
@@ -1852,7 +1852,7 @@ void main() {
   const definitions = [
     {
       blockType: BlockType.BUTTON,
-      text: "Open extra resources",
+      text: Scratch.translate("Open extra resources"),
       func: "openSite",
       def: function () {
         // Exempted from Scratch.openWindow as initiated by user gesture.
@@ -1863,7 +1863,7 @@ void main() {
     },
     {
       blockType: BlockType.BUTTON,
-      text: "Open sample project",
+      text: Scratch.translate("Open sample project"),
       func: "getSampleProject",
       def: function () {
         const url = new URL(location.href);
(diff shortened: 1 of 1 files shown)


Brackets-Coder:
!format

PPPDUD (approved):
Looks good. I will note that most of the text is still in English even after switching languages, but I think that's on the translation team to figure out. I approve.

Brackets-Coder:
@Xeltalliv is the switch case refactor okay?

Xeltalliv:
Firstly, the lack of translations was intentional:
Screenshot_20250915-011041_Discord
All the documentation is also purely in English: https://xeltalliv.github.io/simple3d-extension/

Secondly, due to the unusual way how extension is structured, the object defining all blocks gets initialized only once when extension is loaded. This means Scratch.translate will only get called once and switching language afterwards without reloading the whole editor wouldn't work. Other extensions reinitialize the whole object, so they can switch language without restarting the editor.

Thirdly, the whole point of this unusual structure is to improve cleannes and readability of the code, while also improving performance by not reinitializing so many blocks. The "def" of blocks already gets additionally processed with a loop. So I think the only right approach in this case would be to also build some system with loops to apply Scratch.translate on everything automatically without all that boilerplate, and reapply it when the language is changed without reconstructing the whole object.

All this to say, adding translations here isn't as simple as in other extensions. I do not approve this PR in it's current state and I'd like to eventually add it myself properly.

> @Xeltalliv is the switch case refactor okay?

Yes, that change is fine.

Brackets-Coder:
Noted, I wasn't aware of this.

Brackets-Coder:
It's still okay to leave the translations in the block text though right? They won't work without l10n so they can just stay there until support is added...?

PR #2252, opened 2025-09-24
Added Project Lighthouse extension

Sesquil:
Added extension for Project Lighthouse client, enabling students to display images on the highrise building at Kiel University


Brackets-Coder:
This won't be merged:

1. It seems to display images on the side of a real-life building. This is a bloat to the gallery and 99% of scratch users will not find it useful.
2. It relies on an external domain that could be whitelisted or go down at any point, and future compatibility is not guaranteed
3. You're failing every single one of the merge requirements; even if they passed it probably wouldn't be merged.
4. External APIs (especially those that require some kind of key or authentication token) are explicitly forbidden in the [contributing guidelines](https://github.com/TurboWarp/extensions/blob/37ec92186625f2c4b5c341f51adae707fd02206d/CONTRIBUTING.md); it's not good for security practices.
5. This extension relies on server-side interaction and does not work offline...

PR #2258, opened 2025-09-28
feat: add securelocalstorage (Local Storage V2)

ampelectrecuted:
Basically like the current Local Storage extension but less hackable.

### DIFF
--- a/extensions/8to16/securelocalstorage.js
+++ b/extensions/8to16/securelocalstorage.js
@@ -14,7 +14,9 @@
   }
 
   // If no crypto.randomUUID support, error out.
-  
+  if (typeof crypto.randomUUID !== "function") {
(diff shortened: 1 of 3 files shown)


GarboMuffin:
i'm not opposed to making the existing local storage extension store the namespace in extensionStorage so that people don't need to run that block all that time, and you can even generate a UUID on first launch if you want, that's fine

but calling this scheme "secure" is definitely misleading. you shouldn't be doing that

GarboMuffin (changes requested):
is there a reason you can't add the autogenerated id + persistence to the original extension?

GarboMuffin:
it shouldn't cause compatibility issues unless the project is showing the warnings about running blocks in the wrong order, in which case it is already broken and we're not breaking it any more

GarboMuffin:
https://github.com/TurboWarp/extensions/pull/2270

Issue #880, opened 2023-08-13
[IDEA] port of scratch labs face sensing extension to turbowarp


idoinfactexist:

NexusKitten:
There's a few issues with this;
1. It wouldn't be allowed to be a direct port. Like Animated Text, it would have to be rebuilt from the ground up for copyright reasons.
2. From what I know of, the Scratch Labs face sensing extension doesn't use an API or library; essentially, Garbo would have to program the entire extension, not just piggyback off pre-existing APIs.
3. The reason Face Sensing hasn't been added to Scratch yet is due to the AI not being advanced enough to detect certain races/people in certain lightings. Not sure if Garbo cares much about that, but it's an issue nonetheless.

With that being said, I've been hoping for this for a while, hope somebody makes it!

PPPDUD:
> There's a few issues with this;
> 
> 1. It wouldn't be allowed to be a direct port. Like Animated Text, it would have to be rebuilt from the ground up for copyright reasons.
> 2. From what I know of, the Scratch Labs face sensing extension doesn't use an API or library; essentially, Garbo would have to program the entire extension, not just piggyback off pre-existing APIs.
> 3. The reason Face Sensing hasn't been added to Scratch yet is due to the AI not being advanced enough to detect certain races/people in certain lightings. Not sure if Garbo cares much about that, but it's an issue nonetheless.
> 
> With that being said, I've been hoping for this for a while, hope somebody makes it!

Why was this marked as spam?? They're just providing their opinions and a few facts!

artem21072015:
> 3. The reason Face Sensing hasn't been added to Scratch yet is due to the AI not being advanced enough to detect certain races/people in certain lightings. Not sure if Garbo cares much about that, but it's an issue nonetheless.
it has been added so it's outdated

PR #2264, opened 2025-10-04
Face sensing

GarboMuffin:
Closes https://github.com/TurboWarp/extensions/issues/880

Still a lot to do.

 - I'm developing against the Scratch Lab version. Not sure if they changed anything in the version that will get released.
 - Still need to figure out how large external libraries will be handled. Being forced to figure that out now.
 - Need to verify that performance impact is minimal, probably add some throttling to keep in line with the Scratch Lab version.
 - Blazeface doesn't directly tell us some things like how titled the head is or where the top is after accounting for rotation. I wrote some of my own estimators for those that seem to closely match Scratch Lab, at least on me.

### DIFF
--- a/extensions/lab/face-sensing.js
+++ b/extensions/lab/face-sensing.js
@@ -1,0 +1,346 @@
+// Name: Face Sensing
+// ID: faceSensing
+// Description: Compatible with Scratch Lab's Face Sensing experiment.
+// License: MPL-2.0
+
+(async function (Scratch) {
+  "use strict";
+
+  const initializeDetector = async () => {
+    // TODO: this is awful
+    await import('https://cdn.jsdelivr.net/npm/@mediapipe/face_detection');
+    await import('https://cdn.jsdelivr.net/npm/@tensorflow/tfjs-core');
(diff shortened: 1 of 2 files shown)


extensions/lab/face-sensing.js
```diff
@@ -0,0 +1,434 @@
+// Name: Face Sensing
+// ID: faceSensing
+// Description: Compatible with Scratch Lab's Face Sensing experiment.
```
ampelectrecuted: Description idea: `An easy way to detect faces. Compatible with Scratch Lab's Face Sensing experiment.`
When this is actually added it can be changed to `An easy way to detect faces. Compatible with Scratch's Face Sensing extension.`

GarboMuffin:
For the extra non-script assets that blazeface loads - Not clear what we'll do about those since #2267 doesn't plan to include anything. For now I'll probably put a copy of them on packagerdata.turbowarp.org and hardwire the desktop app to make this work offline.

GarboMuffin:
it's fine to keep it here. packaging everything so that it can work offline is easier outside of that repository

PR #2266, opened 2025-10-04
Convert development tools to ESM and update dependencies

GarboMuffin:
overdue

problem last time was that the desktop app used to import this repo as a submodule and wouldn't have the dependencies. it since was changed to be an npm dependency which fixes that


Brackets-Coder:
This is fantastic thanks so much

PR #2272, opened 2025-10-07
lab/face-sensing: is compatible with scratch 

GarboMuffin:


GarboMuffin:
yes we're not using their banner

if there is one in the art issue or something, you can put that in, i personally am not going to make anything for it

PR #2286, opened 2025-10-18
Add Browser Bridge extension

LeDavax:
## Summary

Adds the unsandboxed extension **"Browser Bridge"**, providing access to browser APIs such as:
- `localStorage` (set/get/delete)
- `console.log`
- `alert`, `confirm`, and `prompt` dialogs
- `open URL` in a new tab

## Details

This extension is intended to help users interact with the browser environment directly from TurboWarp projects.
All actions that read, modify, or delete data, as well as access dialogs or the console, require user consent.
It can be useful for experiments, debugging, or creating web-based Scratch projects that store data locally.

**Author:** Startoto1007  
**License:** MPL-2.0  

---

Browser Bridge extension example

### DIFF
--- a/extensions/Startoto1007/browser-bridge.js
+++ b/extensions/Startoto1007/browser-bridge.js
@@ -7,6 +7,79 @@
 (function (Scratch) {
   "use strict";
 
+  if (!Scratch.extensions || !Scratch.extensions.unsandboxed) {
+    console.warn(
+      Scratch.translate(
+        "Browser Bridge: running in sandboxed mode or Scratch extensions API not available. Some features may be limited."
+      )
+    );
+  }
+
+  const consent = { storage: null, misc: null };
+
+  function askConsent(type, message) {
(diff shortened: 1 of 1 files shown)


Startoto1007:
> 90% chance this won't be merged

Why?

PPPDUD:
> > All of the LLM detectors I used claim that the description was written by a human, but I have my doubts that a real person would follow the rule of threes so accurately.
> > Please do not slop on TurboWarp.
> 
> I wouldn't be so critical; not everyone is unintelligent or uses AI irresponsibly It's not something you can guarantee and generated descriptions are nothing to worry about
> 
> However, if the code were obviously AI, there would be a problem

I agree that it's nothing to worry about _in its current state_ (well-meaning new contributors), but it's easy enough to just make 10 spam accounts and make convincing-enough LLM slop PRs with slight security vulnerabilities in them.

PR #2293, opened 2025-10-19
obviousAlexC/newgroundsIO: use Scratch.external

GarboMuffin:
Use the new imports instead of pasting the entire library in there.

In development, requires https://github.com/TurboWarp/scratch-vm/commit/55674f82eef8d901affdb2e6061d5127020be489

But when using the final version that gets published on extensions.turbowarp.org, we inline the entire NGIO script in there so there is no dependency on any VM feature


GarboMuffin:
@David-Orangemoon do you have any complaints here?


development version needs VM changes to run, but final built version inlines the script so it works the same as it currently does

PR #2306, opened 2025-10-22
text.js -- Add array of regex matches block

ScratchFakemon:
Added modified versions of `split` and `matchRegex` that both return an array instead of a specific item, which can be useful in more complex projects. I also moved `testRegex` so all of the match blocks could be together in the palette.

Pretty quick, so there isn't that much polish. It was mostly an easy copy-and-paste job and things are still left in.

### DIFF
--- a/extensions/text.js
+++ b/extensions/text.js
@@ -571,29 +571,6 @@
       }
       return splitCache.arr[item - 1] || "";
     }
-    splitJSON(args, util) {
-      // split but it returns an array
-      const string = Scratch.Cast.toString(args.STRING);
-      const split = Scratch.Cast.toString(args.SPLIT);
-
-      // Cache the last split
-      if (
-        !(
-          splitCache &&
-          splitCache.string === string &&
-          splitCache.split === split
-        )
-      ) {
-        const regex = this._caseInsensitiveRegex(split);
-
-        splitCache = {
-          string,
-          split,
-          arr: string.split(regex),
-        };
-      }
-      return JSON.stringify(splitCache.arr) || "[]";
-    }
 
     count(args, util) {
       // Fill cache


ScratchFakemon:
Converted to draft so I could fix some things

Brackets-Coder (commented):
Syntax looks fine I guess, nothing jumps out just a quick skim

For the split JSON block, is the intended functionality to leave blank elements where the split was or remove everything except the split 

e.g., "apple" split by "p" returns ["a","","le"], did you mean to remove the blank middle item?

Another concern I have—do the "regex" inputs accept RegEx blocks from the RegEx extension? Those return strings, so you might want to find ways to parse the strings back into actual JS regex

Brackets-Coder:
> I just copied the existing `split` and `matchRegex` blocks and just made them return the whole array, so the functionality should match the existing blocks. There isn't anything special going on. Like I said, it's unpolished and kind of hacked together.
> 
> (I also have no idea how regexes work sooooo time to go learn stuff ig)

https://www.w3schools.com/js/js_regexp.asp

Brackets-Coder:
Wait so if the functionality is the same what actually changed?

ScratchFakemon:
> Wait so if the functionality is the same what actually changed?

I only changed it so it returned the whole (stringified) array instead of a single item.

ScratchFakemon:
@Brackets-Coder using what I said, could you re-review and see if there are any changes that I actually need to make?




Also (assuming the location on your Scratch account is correct), happy Thanksgiving 🦃 :)

Brackets-Coder:
Hello there, 
I understand your frustration at our tardiness for getting such a small pull request merged. While it has been several months, December in particular is quite a busy month for most people and Turbowarp isn't a full-time gig for us.

Since you already have my approval and I can't give you another, @PPPDUD or @SharkPool-SP  can you review this small one?

SharkPool-SP:
Sure I'll look later

extensions/text.js
```diff
@@ -296,10 +311,10 @@
             },
           },
           {
-            opcode: "countRegex",
+            opcode: "matchRegexJSON",
             blockType: Scratch.BlockType.REPORTER,
             text: Scratch.translate({
-              default: "count regex /[REGEX]/[FLAGS] in [STRING]",
+              default: "[STRING] matched by regex /[REGEX]/[FLAGS] as array",
```
SharkPool-SP: name should be changed to `matches of STRING using regex ...`

SharkPool-SP (changes requested):

extensions/text.js
ScratchFakemon: Done

SharkPool-SP (approved):

extensions/text.js
```diff
@@ -139,6 +139,21 @@
...
+              STRING: {
+                type: Scratch.ArgumentType.STRING,
+                defaultValue: "apple",
+              },
+              SPLIT: {
+                type: Scratch.ArgumentType.STRING,
+                defaultValue: "p",
+              },
+            },
+          },
```
SharkPool-SP: on second thought this block is useless, as it requires a json extension to be used, and the one that is listed in the gallery as a block that does this:
image

SharkPool-SP: this is just a comment so I can change the approval. Remove this block and I will merge

PR #2310, opened 2025-10-26
Enable the Iframe extension to use multiple iframes.

izum00:
Enable the Iframe extension to manage multiple iframes. Keep compatibility with existing projects loading the extension from `https://extensions.turbowarp.org/iframe.js` by minimizing changes and adding new arguments instead of removing existing ones.

### DIFF
--- a/extensions/iframe.js
+++ b/extensions/iframe.js
@@ -7,625 +7,603 @@
 // Edited by izum00
 
 (function (Scratch) {
-    "use strict";
+  "use strict";
 
-    /** @type {Object.<string, {iframe: HTMLIFrameElement, overlay: any, x: number, y: number, width: number, height: number, interactive: boolean, resizeBehavior: string, latestMessage: string|number|boolean}>} */
-    const iframes = {};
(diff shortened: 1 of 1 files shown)


CubesterYT:
!format

extensions/iframe.js
```diff
@@ -148,47 +196,77 @@
           {
             opcode: "display",
             blockType: Scratch.BlockType.COMMAND,
-            text: Scratch.translate("show website [URL]"),
+            text: Scratch.translate("show website [URL] as [ID]"),
```
CST1229, line 199: i think adding arguments to existing blocks counts as breaking compatiblity? (or at the very least, it would probably leave blank un-typable inputs in their place in old projects)

CST1229: actually, yeah https://docs.turbowarp.org/development/extensions/compatibility#arguments-must-never-be-added-to-existing-blocks

izum00: I handled those blank inputs by using "iframe" as the default ID when the field is empty. It still works properly even if the input is left blank.

izum00: After checking again, I found that it becomes like [https://studio.penguinmod.com/#1781399388](https://studio.penguinmod.com/#1781399388)
. So I’m planning to release this as a separate extension named “Iframe+”.

PR #2311, opened 2025-10-27
A new extension for creating and managing multiple iframes.

izum00:
A new extension for creating and managing multiple iframes.
From: [https://github.com/TurboWarp/extensions/pull/2310](https://github.com/TurboWarp/extensions/pull/2310)

### DIFF
--- a/extensions/iframe.js
+++ b/extensions/iframe.js
@@ -4,12 +4,16 @@
 // Context: "iframe" is an HTML element that lets websites embed other websites.
 // License: MIT AND MPL-2.0
 
+// Edited by izum00
+
 (function (Scratch) {
   "use strict";
 
-  /** @type {HTMLIFrameElement|null} */
-  let iframe = null;
-  let overlay = null;
+  /** @type {Object.<string, {iframe: HTMLIFrameElement, overlay: any, x: number, y: number, width: number, height: number, interactive: boolean, resizeBehavior: string, latestMessage: string|number|boolean}>} */
+  const iframes = {};
+
+  // ブロックモードを管理する変数 (デフォルトは古いブロック)
+  let useNewBlocks = true;
 
   const featurePolicy = {
     accelerometer: "'none'",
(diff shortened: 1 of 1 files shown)


CubesterYT:
Hello, please stop closing and recreating PRs


Also please refrain from adding translations yourself in the PR. I see other edits but I can't check them out from my phone.

izum00:
> Hello, please stop closing and recreating PRs
> 
> Also please refrain from adding translations yourself in the PR. I see other edits but I can't check them out from my phone.

I'm sorry. These are my first pull request and I'm new to using Github.
Also, what do you mean by "I see other edits but I can't check them out from my phone"?

izum00:
> Imo it would be better to just add these features to the main iframe extension. I'm not a reviewer and these aren't official change requests, but I think you should at least look into how plausible it would be for the extension to support multiple iframes.

I also submitted a pull request to add an ID assignment feature for the iframe in the existing Iframe extension. My assumption was that projects already using the existing Iframe extension would leave the argument blank or use the extension’s default value, so I designed it so that if the ID argument was empty, the iframe would use a default ID of “iframe.” However, in practice, there is a high possibility that the argument fields of the blocks may not display properly - such as in "[https://studio.penguinmod.com/#1781399388](https://studio.penguinmod.com/#1781399388.%E2%80%9D)".
 Therefore, adding only an argument to an existing extension is risky.
That pull request was submitted at https://github.com/TurboWarp/extensions/pull/2310
 and has since been closed.
This demonstrates that existing extensions cannot be easily replaced, and by releasing it as a new extension, I aimed to avoid such bugs.

izum00:
> > > Imo it would be better to just add these features to the main iframe extension. I'm not a reviewer and these aren't official change requests, but I think you should at least look into how plausible it would be for the extension to support multiple iframes.
> > 
> > 
> > I also submitted a pull request to add an ID assignment feature for the iframe in the existing Iframe extension. My assumption was that projects already using the existing Iframe extension would leave the argument blank or use the extension’s default value, so I designed it so that if the ID argument was empty, the iframe would use a default ID of “iframe.” However, in practice, there is a high possibility that the argument fields of the blocks may not display properly - such as in "[https://studio.penguinmod.com/#1781399388](https://studio.penguinmod.com/#1781399388.%E2%80%9D)". Therefore, adding only an argument to an existing extension is risky. That pull request was submitted at #2310 and has since been closed. This demonstrates that existing extensions cannot be easily replaced, and by releasing it as a new extension, I aimed to avoid such bugs.
> 
> have two separate blocks: one with the id and one without. as stated by the documentation, [arguments must never be added to existing blocks.](https://docs.turbowarp.org/development/extensions/compatibility#arguments-must-never-be-added-to-existing-blocks)

That's a good idea!!
I will fix it.

izum00:
...

PR #2319, opened 2025-11-08
Add DollyPro Extension

hwxqyworld:
adapted from [Gandi-IDE](https://github.com/Gandi-IDE/custom-extension)


Brackets-Coder:
@CubesterYT I'm going to close this for the following reasons:

1. The image is blatantly generated entirely with AI and contains text (even if licensing it isn't a problem, it still contains text which can't be translated)
2. Speaking of translation, this image contains comments (I think) in Mandarin which make the reviewing process extremely difficult
3. The entire extension structure is incorrect and does not follow Turbowarp's best practices:
    - There seems to be an "assets" folder and an "icon" file, both of which are out of place.
    - The extension is not inside an IIFE and completely breaks lint
    - The naming conventions in the files are inconsistent and so will not be parsed by the builder correctly
    
TLDR basically the entire PR breaks the contributing guidelines

Issue #2328, opened 2025-11-17
AD notation for "Format Numbers" extension is inaccurate.


rb-f:
**Hello, I'm ten_6044 from Scratch.**

In the "format numbers" extension, when converting to AD standard, numbers less than 10^36 work as intended. However, once it reaches 10^36, it sounds incorrect.

### Expected Result
If you format number 10^33 to AD standard, it returns `1Dc`, meaning 1 decillion.
If you format number 10^36 to AD standard, it should return `1UDc`, meaning 1 undecillion.

### Actual Result
If you format number 10^33 to AD standard, it returns `1Dc`, meaning 1 decillion.
If you format number 10^36 to AD standard, it instead returns `1 DcD` (1 duodecillion), which is supposed to be `1UDc`.

So it turns out that the "-illion" value is off by one, and the order of the words are not in order.
After 1 decillion (1Dc) is 1 undecillion (1UDc), not `1 DcD`.

Duodecillion is actually `DDc`, not `DcD`.

Additionally, quindecillion (should be 10^48, but yielded at 10^45) is `DcQt`, which is unconventional because "quindecillion" does not contain the letter "T." It is supposed to say "QnDc".

### Proof

<img width="541" height="88" alt="Image" src="https://github.com/user-attachments/assets/0cecb61e-05d4-415a-bf58-68df15edf190" />

DogeisCut:
Will get to it as soon as I can!

PPPDUD:
Can reproduce.

PR #2331, opened 2025-11-18
DogeisCut/FormatNumbers: Resolve AD notation inconsistencies

DogeisCut:
Resolves: https://github.com/TurboWarp/extensions/issues/2328

Results:
10^33: 1Dc
10^36: 1 UDc
10^45: 1 QaDc
10^48: 1 QnDc

See https://pastebin.com/JpUFnLtp for full results, please let me know if you see anything wrong as I can not locate the resource I was previously using to verify this.

### DIFF
--- a/extensions/DogeisCut/FormatNumbers.js
+++ b/extensions/DogeisCut/FormatNumbers.js
@@ -56,13 +56,13 @@
         "No",
         "Dc",
       ];
-      const unitPrefixes = ["U", "D", "T", "Qa", "Qt", "Sx", "Sp", "O", "N"];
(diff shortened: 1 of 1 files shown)


extensions/DogeisCut/FormatNumbers.js
Brackets-Coder, line 52: Was this also meant to be a Qn instead of Qt?

rb-f: Well no, you can still use `Qt` here, the change to `Qn` only applies to quindecillion, quinvigintillion, but not quintillion by itself.

GarboMuffin (approved):

CubesterYT (approved):

PR #2339, opened 2025-11-22
obviousAlexC/newgroundsIO: use minified JS instead

GarboMuffin:
Saves some space. It should be easy to verify that the JS file was generated at the same commit as the unminified JS so you don't need to do crazy testing.

### DIFF
--- a/extension-dependencies.json
+++ b/extension-dependencies.json
@@ -36,8 +36,8 @@
       "sha256": "95f4d1cbf099f57161b664bc048426ec3df92637801a4c79116e83315aa787e7",
       "contentType": "application/javascript; charset=utf-8"
     },
-    "https://raw.githubusercontent.com/PsychoGoldfishNG/NewgroundsIO-JS/b0383337ef72d42dac12b7a4d7a24dfbc4105eb3/dist/NewgroundsIO.js": {
-      "sha256": "b6f68cf85f59b17d6489d1dde4b499827b0cc29f58472832643abaa8b9352ae8",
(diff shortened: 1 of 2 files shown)


PPPDUD:
Does this count as a logic change under `REVIEW_PROCESS.MD`?

GarboMuffin:
yes you would want to test that it still works because i actually didn't do that

PPPDUD:
> if you can look at the repository where the JS is coming from, look at the commits that added them and convince yourself that the minified and non-minified JS files were generated at the same time and should work identically, then you can get away with not a lot of testing and still be probably confident that nothing was broken

I'll just leave this to someone else, I think.

David-Orangemoon:
> * [ ]  I tested all modified functionality and found no bugs or unintuitive behavior.
> 
>     * [x]  I believe a well-formed project using an older version of this extension will not be broken by this new version.
> 
>     * [ ]  I have no additional concerns.
> 
> 
> I'm not sure if a game has to be published for the API to work, but it seems all the blocks in the `Save data` section (except for the hat) return empty strings. It's a similar situation with the medals, and scoreboard, as despite providing valid ids into both, and successfully connecting to the game and logging in, they didn't seem to do anything. Results are inconclusive and would need larger testing from a more Newgrounds savvy reviewer.
> 
> Despite this, I believe due to the small change, that this is likely just how this extension has always worked, and is unrelated to the change to a minified version. However, I will not be approving until I can get some confirmation on this behavior.

You just need to be connected with a login prompt verified. Then all blocks should work.

PR #2340, opened 2025-11-22
Add CODEOWNERS

GarboMuffin:
This file makes GitHub automatically request review from the people or teams listed in the file when a pull request is marked ready for review. For now, we won't enforce requiring review from the person/team listed there. There's another setting in GitHub settings where I can add separate rules for that. Hoping to not need to use that.


PPPDUD:
Are you sure that this is a good idea? Auto-requesting review for every PR will lead to a dozen emails every time that @dependabot finds a new change to make.

PPPDUD:
> the proposed CODEOWNERS file says that changes to package.json package-lock.json will only request review from me

I see now, thanks.

Brackets-Coder:
May I politely inquire as to why this is still a draft?

PR #2359, opened 2025-12-14
SharkPool/CameraV2 -- Fix hidden sprites not receiving camera updates

SharkPool-SP:


CubesterYT (approved):

Issue #2370, opened 2025-12-29
Extension: Camera V2 fails to set sprite zoom, position, and rotation correctly when a "shown" block is used.


DogeisCut:
https://github.com/user-attachments/assets/9d604786-88c4-4b96-9a62-67e38183d063

The video demonstrates the issue pretty clearly, I think.
For the record, this is not what the previous behavior was.

I believe this to be caused by this commit: https://github.com/TurboWarp/extensions/pull/2359#event-21770524129
Frankly I should've help test then but I was busy.

@SharkPool-SP

PR #2372, opened 2025-12-29
SharkPool/Camera V2 -- Fix hidden sprites not receiving camera updates

SharkPool-SP:
This also fixes https://github.com/TurboWarp/extensions/issues/2370

Its basically the same code as before the revert, except I added the important line that makes it *not* refresh the camera if the visibility change is the same value is before...

this was tested under the same conditions and project @DogeisCut provided


DogeisCut:

DogeisCut (approved):

PR #2375, opened 2025-12-31
Create ntfyscratch.js

xmeroriginals:
Send notifications from your Scratch project to your phone, tablet, or computer using ntfy.sh

### DIFF
--- a/extensions/XmerOriginals/ntfyscratch.js
+++ b/extensions/XmerOriginals/ntfyscratch.js
@@ -1,4 +1,3 @@
-// !format
 // Name: ntfyScratch
 // ID: ntfyScratch
 // Description: Send notifications from your Scratch project to your phone, tablet, or computer using ntfy.sh. Instantly receive messages, alerts, and status updates from your project in real time.
--- a/extensions/XmerOriginals/ntfyscratch.js
+++ b/extensions/XmerOriginals/ntfyscratch.js
@@ -89,7 +89,7 @@
       const topic = args.TOPIC;
       const message = args.MESSAGE;
 
-      fetch(`https://ntfy.sh/${topic}`, {
+      Scratch.fetch(`https://ntfy.sh/${topic}`, {
         method: 'POST',
         body: message
       }).catch((err) => {
@@ -106,7 +106,7 @@
         tags: args.TAGS.split(',').map((tag) => tag.trim())
       };
 
-      fetch('https://ntfy.sh/', {
+      Scratch.fetch('https://ntfy.sh/', {
         method: 'POST',
         body: JSON.stringify(bodyData),
         headers: {
@@ -120,7 +120,7 @@
     _openUrl(url) {
       Scratch.canOpenWindow(url).then((allowed) => {
         if (allowed) {
-          window.open(url, '_blank');
+          Scratch.openWindow(url, '_blank');
         }
       });
     }
--- a/extensions/XmerOriginals/ntfyscratch.js
+++ b/extensions/XmerOriginals/ntfyscratch.js
@@ -5,83 +5,85 @@
 // License: MPL-2.0
 
 (function (Scratch) {
-  'use strict';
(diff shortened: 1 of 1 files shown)


Seigh-sword:
lint, problem must be caused by, transilation issues, so... you can fix it by making each block's name to, 
```code 
Scratch.translate("Block name"), 
```
***

xmeroriginals:
bro is this bot kidding me

Seigh-sword:
and also, my extension is still under review so, ya

extensions/XmerOriginals/ntfyscratch.js
```diff
@@ -0,0 +1,139 @@
+// !format
```
Brackets-Coder: This line should not be here
```suggestion
```

Brackets-Coder (changes requested):

Brackets-Coder:
!format

extensions/XmerOriginals/ntfyscratch.js
```diff
@@ -0,0 +1,204 @@
+// Name: ntfyScratch
```
ampelectrecuted, line 1: scratch is a trademark; i suggest you change this to something like turbontfy or

extensions/XmerOriginals/ntfyscratch.js
```diff
@@ -0,0 +1,204 @@
...
+
+  class ntfyScratch {
+    getInfo() {
+      return {
+        id: "ntfyScratch",
+        name: Scratch.translate("ntfyScratch"),
+        color1: "#55bba6",
+        color2: "#358876",
+        blocks: [
+          {
```
ampelectrecuted, line 18: maybe use docsURI here

extensions/XmerOriginals/ntfyscratch.js
```diff
@@ -0,0 +1,204 @@
...
+                defaultValue: 3,
+                menu: "priorityMenu",
+              },
+              TAGS: {
+                type: Scratch.ArgumentType.STRING,
+                defaultValue: "clock,calendar",
+              },
+              CLICK: {
+                type: Scratch.ArgumentType.STRING,
+                defaultValue: "https://xelabs.xmeroriginals.com",
```
ampelectrecuted, line 112: a bit self-promotional even though iirc another extension on the gallery did this (lily's toolbox, shovelutils, and rixxyx are literally named after the users who made them)

ampelectrecuted (commented):
just some suggestions because i'm not a formal reviewer

SharkPool-SP:
We have this, and it doesnt require an outside app for functionality:
image

PPPDUD:
> We have this, and it doesnt require an outside app for functionality: image

@SharkPool-SP I think that ntfy.sh lets you notify other devices too.

SharkPool-SP:
Confirm that and come back to me about it. Not exactly how useful that would be to the gallery.

PPPDUD:
> Confirm that and come back to me about it. Not exactly how useful that would be to the gallery.

image

> _It allows you to send notifications to your phone or desktop via scripts from any computer..._

[(source)](https://ntfy.sh)

Brackets-Coder:
I'm still confused about what this is _supposed_ to do

Brackets-Coder:
@xmeroriginals 
@TurboWarp/extension-reviewers 

I think I'm going to close this PR. I like the idea of push notifications, but the only advantages it seems to propose over the existing notifications extension is saving notifications to public topics on a third party host which may or may not be accessible depending on internet access. Additionally, it also seems to require the user to have a Nfty account or app, which means this is probably only applicable to single-use projects and not something that can be easily distributed.

Let me know if I have any misunderstandings and feel free to re-open if you think this might be beneficial

GarboMuffin:
i think it's fine to close: in theory the HTTP extension(s) can do this one (not tested) with some effort

Issue #2271, opened 2025-10-07
Consolidation of duplicate PRs


Brackets-Coder:
I'm going to try to consolidate some of the duplicate extension pull requests. This is a non-conclusive list and is subject to change.
Some of these extensions will be harder to consolidate than others as they are either well-developed or both have benefits over the other.
Some of these we may not even want to consolidate yet, and others are part of a bigger issue (like sensing+ or "secure" storage).

- [ ] #1354 and #1066
- [ ] #2113, #1861, #2531, and #2424
- [ ] #1617, #1432, and #1608
- [x] #2201 and #2373
- [ ] #1183 and #1176
- [x] #1040 and #1060
- [x] #2104 and #1122 
- [x] #2389 and #2423 — somewhat similar, but different context/nuance
- [x] #2158 and #2431
- [ ] #1015 and #1143

CubesterYT:
I think a very big issue with "consolidating duplicate extensions" is that each dev has their own ideas and want to have their stake in the extension gallery. By taking down duplicates, you fuel the fire of "one extension is better than the other, so the lesser ones must be closed". If there's duplicate extensions, and none have been merged yet, I rather wait to see which one ACTUALLY gets merged before we close duplicates. It's nice to give people chances considering this repository is slow and having a chance against devs who are already on the gallery are tight.

CubesterYT:
There was fights over this over a year ago with this same issue.

SharkPool-SP:
That's not really fair. Older ones came first can should have a higher priority.

Newer ones show up higher in the list. People might not realize there's a duplicate lower.

SharkPool-SP:
If the older ones were merged, we wouldn't be having duplicates in the first place

Brackets-Coder:
It's clear that this issue is a necessary one to address, but should be done so carefully. 
I'm going to attempt to proceed with caution with the assistance of anyone who's willing to provide feedback or help

SharkPool-SP:
I think it's fine to have collaboration. But in the case neither developer wants to collaborate, the better-written/older pr should take priority

SharkPool-SP:
> > > I think it's fine to have collaboration. But in the case neither developer wants to collaborate, the better-written/older pr should take priority
> > 
> > That was my intention all along. For example, see [#1040](https://github.com/TurboWarp/extensions/pull/1040). Although it was older and technically had more priority, your extension had the same features and more. See the note I added upon closing for extra context if necessary
> 
[...]

PR #2389, opened 2026-01-17
Locale: An Extension for Manual Translations

ScratchFakemon:
Locale is an extension for registering manual translations. This is useful if you want real, human-created translations instead of simple machine ones that the Translate extension provides. This isn't the most feature-filled extension and it didn't take too long to make, so there may be weird bugs or problems.


ScratchFakemon:
!format

ScratchFakemon:
Ignore the random commits about temp files and changing the deployment branch, the former were for creating the folder on GitHub web and the latter were so I could preview the gallery with these changes (which didn't end up working). These were reverted.

ScratchFakemon:
I'm working on a block for getting the user's language code as opposed to the Translate extension's language _name_ feature, just need to actually figure out how to get the TurboWarp language

ScratchFakemon:
!format

ScratchFakemon:
Eslint warnings are because of the format changing comment positions, 1 sec

ScratchFakemon:
grr

ScratchFakemon:
# **_FINALLY_**

ScratchFakemon:
Before I forget, I'll update image attributions

ScratchFakemon:
wow im so glad my markdown didnt trigger a type warning

ScratchFakemon:
# Things to eventually get done
###### Last updated: January 19, 2026
###### These are in chronological order of when I wanted them done.

- [x] Add blocks for converting between language codes and names (likely with a large lookup table): `(name of language code [es v])` and `(code of language named [Spanish v])`
- [x] Add a proper list of languages for the above blocks to call
- [x] Allow any translation of a language's name to be used in `codeFromName`, similarly to the Translate extension
- [x] Add documentation
- [x] Update documentation

ScratchFakemon:
Most thorough lint check ever

ScratchFakemon:
!format

ScratchFakemon:
Now I just need documentation, which shouldn't be difficult

ScratchFakemon:
!format

ScratchFakemon:
no clue why i'm using this comment section as a devlog when that's not what this is for but it seems ready to mark as ready for review

SharkPool-SP:
this is just an JSON extension for a specific use. You can use the general-purpose JSON extension and replicate this...

ScratchFakemon:
> this is just an JSON extension for a specific use. You can use the general-purpose JSON extension and replicate this...

image

SharkPool-SP:
the actual translation part

ScratchFakemon:
> the actual translation part

What do you suggest I change?

SharkPool-SP:
Remove that portion

ScratchFakemon:
> Remove that portion

###### I know this comment might sound rude at first, but it's _argumentative_. I'm proving my point and why I believe that the blocks can stay in the extension.

TurboHook, despite being almost fully replaceable by the HTTP and JSON extensions, still exists. Utilities and Fetch share a fetch block. Asset Manager and ShovelUtils can import sprites, costumes, sounds, extensions, and even whole projects. So many extensions have similar, near-identical blocks, Additionally, CONTRIBUTING.md only states that "Extensions that are very similar to existing ones" are discouraged. This extension has a unique purpose: to easily allow for manual translations _in the same place_. While the documentation for the extension says that the JSON extension can be useful for helping with some blocks, it's not required. **The blocks mitigate the need for storing JSON in variables, instead putting everything in one place.** Additionally, removing the blocks that could technically be replaced with JSON removes 9 out of 15 blocks—60% of the extension—in favor of a combination of JSON and variables. This includes the merge blocks, which cannot be replicated as easily with the JSON extension (it doesn't have a dedicated merge block), which would have to be removed as they use the data set by other blocks suggested for removal.

ScratchFakemon:
I finally added a button to hide blocks that can be recreated with other extensions and a backup of the lookup table for offline users

ScratchFakemon:
I tested and fixed the backup lookup table, that'll be the next commit

Brackets-Coder:
I know this isn't exactly a duplicate, but for organization purposes I'll be putting this and #2421 together in #2271

ScratchFakemon:
> I know this isn't exactly a duplicate, but for organization purposes I'll be putting this and #2421 together in #2271

alr

ScratchFakemon:
!format

ScratchFakemon:
> I know this isn't exactly a duplicate, but for organization purposes I'll be putting this and #2421 together in #2271

Have you decided what you'll do to Locale and Translate+? It looks like not a lot has happened in any of these PRs for a while.

Brackets-Coder:
Since Translate+ has been closed, the "duplicate" issue has been resolved. We can proceed as normal with this pull request. I haven't been able to review due to an abnormally full schedule, but I'll see what I can do in the upcoming weeks.

ScratchFakemon:
!format

ScratchFakemon:
lol some of my default values were literally broken

ScratchFakemon:
thanks github desktop :/

ScratchFakemon:
@Brackets-Coder when you have a moment, could you review this PR and give me your thoughts? Thanks!

ScratchFakemon:
!format

extensions/Fakemon/Locale.js
```diff
@@ -0,0 +1,785 @@
...
+        )
+      ) {
+        return await getLanguageNameAndCodeLookupTableGLOBALIZED(); // Retry on user request only
+      } else {
+        fetchResult = null;
+      }
+    }
+
+    if (fetchResult) {
+      return JSON.parse(await fetchResult.text());
```
pid-j: `JSON.parse` can fail, I'm pretty sure you can do `fetchResult.json()` instead but that would still throw an error

extensions/Fakemon/Locale.js
```diff
@@ -0,0 +1,785 @@
...
+       * @type {{ name: any; code: any; }[]}
+       */
+      let arrayThusFar = [];
+      // @ts-ignore
+      try {
+        if (languageNameAndCodeLookupTableGLOBALIZED.menuMap) {
+          languageNameAndCodeLookupTableGLOBALIZED.menuMap[
+            this._matchLanguages(
+              Object.keys(languageNameAndCodeLookupTableGLOBALIZED.menuMap),
+              JSON.parse(this.getLanguageArray())
```
pid-j: **Please** make a function to just get the raw value

extensions/Fakemon/Locale.js
```diff
@@ -0,0 +1,785 @@
...
+            ],
+          },
+        },
+      };
+    }
+    toggleRecreatableBlocks() {
+      function doToggle() {
+        showRecreatableBlocks = !showRecreatableBlocks;
+        hasToggledRecreatableBlocks = true;
+        Scratch.vm.extensionManager.refreshBlocks();
```
pid-j: This refreshes the entire palette if `refreshBlocks` has no argument, you could probably just add `fakemonLocale` as an argument

ScratchFakemon: oh yeah i probably *should* do that

ScratchFakemon:
i made the same commit twice since i thought it didnt work but i guess it did

extensions/Fakemon/Locale.js
```diff
@@ -0,0 +1,814 @@
...
+   */
+  let languageNameAndCodeLookupTable;
+
+  // @SharkPool-SP mentioned that many blocks in Locale can be recreated by other extensions (most notably JSON)
+  let showRecreatableBlocks = true; // Whether or not to show blocks that can be recreated with other extensions
+  let hasToggledRecreatableBlocks = false; // Whether or not the previous variable has been toggled so the explanation only shows once
+
+  let localeObject =
+    // @ts-ignore
+    Scratch.vm.runtime.extensionStorage["fakemonLocale"]?.localeObject || {};
```
pid-j, line 83: You should use `Object.create(null)` instead of `{}` as it doesn't allow accessing dangerous values (such as `__proto__` and `constructor`)

extensions/Fakemon/Locale.js
```diff
@@ -0,0 +1,814 @@
...
+        doToggle();
+      }
+    }
+    /**
+     * @param {{ JSON: string; }} args
+     */
+    setFullLocaleJSON(args) {
+      let backupLocaleObject = localeObject; // Save a backup in case something breaks
+      try {
+        localeObject = JSON.parse(args.JSON);
```
pid-j, line 473: You should check if the value being parsed is actually an `Object` (~~`localeObject instanceof Object`~~ `Object.getPrototypeOf(localeObject) === Object.prototype` should work) because `JSON.parse` can return non-dictionary types (e.g. `JSON.parse("1")`)

extensions/Fakemon/Locale.js
```diff
@@ -0,0 +1,814 @@
...
+        let translation = localeObject[args.LANG][args.TEXT];
+        if (!translation) {
+          translation = args.TEXT;
+
+          allFailedTranslations.push({
+            ...translationLogInfo,
+            error: `The ${args.LANG} translation object does not contain an entry for ${args.TEXT}`,
+          });
+        }
+        return translation || args.TEXT;
```
pid-j, line 575: This is very dangerous! (JavaScript sucks, it's not your fault)

You should check if the value being returned is a string (`translation instanceof String`) because this can return `Object.prototype` (`localeObject["constructor"]["prototype"]`) and `Function.constructor` (`localeObject["constructor"]["constructor"]`) with the latter basically equaling `eval` (`Function` creates a new wrapper around a function that is created with the provided text)

extensions/Fakemon/Locale.js
pid-j: You can also nest objects and arrays

extensions/Fakemon/Locale.js
```diff
@@ -0,0 +1,814 @@
...
+      // @ts-ignore
+      this._updateLocaleInfo();
+    }
+    /**
+     * @param {{ JSON: string; }} args
+     */
+    mergeFullLocaleJSON(args) {
+      let backupLocaleObject = localeObject; // Save a backup in case something breaks
+      try {
+        localeObject = { ...localeObject, ...JSON.parse(args.JSON) };
```
pid-j, line 486: I would have said to check if this is an Object but the spread operator seems to do nothing if it isn't. Still nice to have if something updates or breaks

GarboMuffin:
Sorry for letting you drag on for months on this. I don't think this is mergeable...

the biggest upside of making an extension in this space should be to give an easy way to manage the strings across languages. such as, perhaps, a nice UI to do the translating by hand, import/export buttons that work with the formats that real translation problems use, but as is, everyone using this is still practically forced to write JSON. am i missing something?

you can, of course, continue to use and develop this yourself though

GarboMuffin:
other things that would be useful to consider but are not present

 - a story for placeholder replacement
 - a story for advanced placeholder replacement with support for the pluralization formats used by real translation platforms and ways to safely do multi-placeholder substituion with arbitrary text content not being able to break other placeholders
 - the UI should warn when it detects a language does not have matching placeholders
   -  example: https://github.com/TurboWarp/extensions/blob/master/development/download-translations.js#L123
 - try to auto-generate a first-pass translation via normal translate extension (actually, various reasons why this shouldn't be done, at least not via trampoline, but it's an idea of the kind of UI enhancement that would make a large difference compared to the status quo)

PR #2391, opened 2026-01-20
box2d: Correct mixup of the x and y coordinates

### LINKED ISSUE #935: Box2D go to x y relative does not work?
If I'm understanding correctly, this block basically acts like the change x/y blocks. What's happening is I'm trying to move the sprite to the right 100 steps but it's doing very strange things. I think what's happening is it's trying to add to the x position when setting both the x and y, so if x is -100 the sprite will end up trying to move relative to x: -100 and y: -100.

Try this code to see what I mean:
```
go to x: (-100) y: (100) [in world v] :: box2d
wait (2) seconds
go to x: (0) y: (0) [relative v] :: box2d
wait (2) seconds
repeat (4) {
  go to x: (50) y: (0) [relative v] :: box2d
  wait (1) seconds
}
```

I could just use change x/y but it fences the sprite.

Steve0Greatness:
The relative space type of the set position block in Box2D was set incorrectly such that it set the y position relative to the current x position rather than the y position.

This means that setting inputs of -2, 20 would result in the sprite going negative in both directions, with the y coordinate always being +20 > the x coordinate. (see below video)

------------------------------------------------------------------------------------------------------

The following video shows the incorrect behavior (current) and the corrected behavior (this patch) using the same script:

Screenshot 2026-01-19 191552

https://github.com/user-attachments/assets/fe0ba0cd-c4cf-4379-8e03-de75ef6a31f1

------------------------------------------------------------------------------------------------------

Didn't look before making this change but I'm pretty sure this patch resolves this issue:

Fixes #935

Also, I think this bug is almost 7 years old, since I found the original patch in Griffpatch's VM fork, and it's from February 2019.

### DIFF
--- a/extensions/box2d.js
+++ b/extensions/box2d.js
@@ -13720,7 +13720,7 @@
           }
           break;
         case SPACE_TYPE_OPTIONS.RELATIVE: {
-          _setXY(util.target, util.target.x + x, util.target.x + y);
+          _setXY(util.target, util.target.x + x, util.target.y + y);
           if (body) {
             const pos = body.GetPosition();
             const pos2 = new b2Vec2(pos.x + x / zoom, pos.y + y / zoom);


Brackets-Coder:
I'll investigate this

Brackets-Coder (approved):
Can confirm this fixes a typo that breaks expected behavior. Nice catch!

 - [ ] I believe a well-formed project using an older version of this extension will not be broken by this new version. - `some projects may use workarounds to fix this manually, so fixing it in the extension may break their workarounds. I can't confirm this but it's such a trivial change that I'm going to approve`

Brackets-Coder (approved):
Can confirm this fixes a typo that breaks expected behavior. Nice catch!

 - [x] I believe a well-formed project using an older version of this extension will not be broken by this new version. - `some projects may use workarounds to fix this manually, so fixing it in the extension may break their workarounds. I can't confirm this but it's such a trivial change that I'm going to approve`

Brackets-Coder:
oops accidentally double-commented

@SharkPool-SP or @PPPDUD could you review and approve so we can get this merged?

SharkPool-SP:
This will break compatibility

Brackets-Coder:
Yes, it will, but is it a sacrifice we're willing to make? It's likely that we're going to have users create issues about it not working if this is merged but it's also not working as intended right now

SharkPool-SP:
Eh good point

SharkPool-SP (approved):

Brackets-Coder:
Alright I'ma go ahead and merge

PR #2415, opened 2026-02-18
TurboBytes Extension

OrangeHub10:
Hi!, Today Thanks To Google Ai And Me, I Made This Extension Called, TurboBytes, I Hope It Can Be In The File Revision!

Backup: If The File Has Been Corrupted Or Useless, Here Is A Backup Code!

(function(Scratch) {   'use strict';    class DataConverter {     getInfo() {       return {         id: 'dataconverter',         name: 'TurboBytes',         color1: '#ff4c4c',         blocks: [           {             opcode: 'getConvertedSize',             blockType: Scratch.BlockType.REPORTER,             text: 'Bytes of [TEXTO] in [UNIT]',             arguments: {               TEXTO: {                 type: Scratch.ArgumentType.STRING,                 defaultValue: 'Bytes'               },               UNIT: {                 type: Scratch.ArgumentType.STRING,                 menu: 'unitsMenu',                 defaultValue: 'Bytes'               }             }           }         ],         menus: {           unitsMenu: {             acceptReporters: true,             items: ['Bytes', 'Kilobytes', 'Megabytes', 'Gigabytes',  'Terabytes',  'Petabytes',  'Exabytes',  'Zettabytes',  'Yottabytes',  'Bits',  'Nibbles',  'Kibibytes']           }         }       };     }      getConvertedSize(args) {       const bytes = new Blob([args.TEXTO]).size;       const unit = args.UNIT;        switch (unit) {         case 'Kilobytes':           return bytes / 1000;         case 'Megabytes':           return bytes / Math.pow(1024, 2);         case 'Gigabytes':            return bytes / Math.pow(1024, 3);           case 'Terabytes':            return bytes / Math.pow(1024, 4);         case 'Petabytes':          return bytes / Math.pow(1024, 5);        case 'Exabytes':         return bytes / Math.pow(1024, 6);        case  'Zettabytes':         return bytes / Math.pow(1024, 7);        case  'Yottabytes':         return bytes / Math.pow(1024, 8);        case  'Bytes':         return bytes;         case  'Bits' :         return bytes * 8;         case 'Nibbles' :         return bytes * 2;         case 'Kibibytes' :         return bytes / 1024;         default: return 0;       }     }   }    Scratch.extensions.register(new DataConverter()); })(Scratch);


Brackets-Coder:
!format

OrangeHub10:
Uptade Note:This MUST Run Sandboxed

CST1229:
> Uptade Note:This MUST Run Sandboxed

i'd assume extensions here CANNOT be required to be run _sandboxed_ since extensions.turbowarp.org is unsandboxed by default

OrangeHub10:
@CST1229 Thanks for viewing this pull request!, Also, Then Were do you Do Request An Sandboxed Extension?

OrangeHub10:
@CST1229 You Said That Its For Default, How Do You Change It?

CST1229:
> @CST1229 Thanks for viewing this pull request!, Also, Then Were do you Do Request An Sandboxed Extension?

you can't in here, you basically just have to put it on your own website or something. or somehow fix it so it works unsandboxed, it's literally just math so it should work (but that might also be a reason for rejection, as it's just one single block that can be replicated with a simple division or added to an existing extension)

OrangeHub10:
@CST1229  Oh Ok

Brackets-Coder:
@CST1229 correct me if I'm wrong, but the functionality of this extension can already be replicated with Bitwise by converting a string to binary and counting bytes (a group of 8 ones or zeros)?

e.g., length of (string to binary / 8 / unit)

GarboMuffin:
This is better as a custom block

OrangeHub10:
@Brackets-Coder But Bitwise Does NOT use "New Blob" So I Think You Are Parciatly Correct!, But NEW BLOB is esential for detecting emoji bytes and special characters (e.g:, ˜,Ò, ect...) + You Could simplify all of that into one single block, or like garbomuffin said, better like A "custom block"(garbo has a point) please tell me if i'm correct, or if im losing the point, ect...

Brackets-Coder:
As of now, bitwise only converts numbers to binary, so you would have to do some kind of unicode or ASCII encoding to get it to work with text...

OrangeHub10:
@Brackets-Coder If Using Unicode for bitwise, thats great but then using  length of (string to binary / 8 / unit) and remembering its hard, + the code itself uses math powers (again garbo has another point because you cant upload mathed extensions sorry first time pull requesting reason i created github) Please tell me if im incorrect or if i'm losing the point ect...

PR #2036, opened 2025-03-16
Add FormatNumbers Extension

DogeisCut:
This is a simple extension that lets you format large numbers into AD Standard, Fixed Decimal, Comma Separated, or Scientific Notation.

It adds but one block, yet a very useful one imo, especially for incremental games!

msedge_2ZeCztrapG
msedge_b8scWM1ITL
msedge_easzgrLqwB
msedge_bvo6IzMFhC
msedge_d6Rc1Kz7ia
msedge_i5PasinNgF
msedge_SZVAPgfXeq

This is the banner, drawn by Dillon (dillonr on Discord)!
FormatNumbers

PR #2424, opened 2026-02-21
BreakEternity Extension

gum-gummii:


gum-gummii:
!format

gum-gummii:
!format

gum-gummii:
!format

gum-gummii:
!format

Brackets-Coder (changes requested):
Including an unminified library (even a small one) directly in the source is not recommended. Use `Scratch.external` (importModule or evalAndReturn) instead to load external dependencies. 

Additionally, it seems like this extension aims to provide better (or more precise) representation of especially large or small numbers. While I may be misunderstanding the goal of this extension, I believe it can already be replicated with #2036 and existing complex numbers pull requests. For organization purposes, I will be categorizing it in #2271 with the complex number extensions, but note that it is not an exact duplicate of those pull requests. 

This is not an in-depth review

gum-gummii:
> Including an unminified library (even a small one) directly in the source is not recommended. Use `Scratch.external` (importModule or evalAndReturn) instead to load external dependencies.
> 
> Additionally, it seems like this extension aims to provide better (or more precise) representation of especially large or small numbers. While I may be misunderstanding the goal of this extension, I believe it can already be replicated with #2036 and existing complex numbers pull requests. For organization purposes, I will be categorizing it in #2271 with the complex number extensions, but note that it is not an exact duplicate of those pull requests.
> 
> This is not an in-depth review

the extension's goal is not to provide better representation of numbers, but rather allow bigger numbers to be used past the limit, along with a multitude of mathematical operations like logs, roots and tetration. and sure i will add the Scratch.external to it

SharkPool-SP:
There has been no messages of collaboration between this PR and its duplicates. This PR has gone stale with no requested changes. Other PRs are more polished. I am closing this.

As per https://github.com/TurboWarp/extensions/issues/2271

PR #2440, opened 2026-03-18
CST1229/zip: don't use regex for data validation

### LINKED ISSUE #2439: Zip Binary Limits
I recently made a sort of parser/converter for an obscure file type and used plain binary to work with it. When trying to save a medium-large file (~4 mb of utf-8 text in binary) in the zip extension, it will not save and says this in the console:

Image

Of course I could convert the file to text, but the format doesn't *always* store text, and because of unsafe characters, if I get a file that wasn't plain text it would completely break it. The solution I decided to do was to convert everything that is binary to base64, then save. The problem with that is it's quite slow. Of course on a better PC it might be faster, but I don't feel like that's much of an excuse. 

Afaik there's no extension (besides zip) that can just immediately convert binary into base64. I may be able to get something working with hex, but I fear that base64 and hex probably all reach the same limit eventually. 

I completely understand if this is outside the scope of the extension, but I can't say I wouldn't be disappointed. 

edit: optimized code a bit by to make it convert to base64 if the file doesn't save properly from binary, so it's at least significantly faster on that front. The repeated readbacks can't be that great speedwise, though

LegoBrainBiker:
this should resolve #2439

### DIFF
--- a/extensions/CST1229/zip.js
+++ b/extensions/CST1229/zip.js
@@ -667,14 +667,14 @@
     }
     isStringValid(string, partLength, validchars) {
       if (string.length % partLength !== 0) {
-        return false
+        return false;
       }
       for (let i = 0; i < string.length; i++) {
         if (!validchars.includes(string[i])) {
           return false;
         }
       }
-      return true
+      return true;
     }
     // get a file/folder by path
     getObj(path, zip = this.zip) {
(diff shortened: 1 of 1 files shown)


LegoBrainBiker:
!format

CST1229 (approved):

CST1229:
right the review template exists

CubesterYT (approved):

snowboyz0825:
Woah, I'm shocked at how quickly this actually got fixed, that's incredible. Appreciate everything y'all do

Brackets-Coder:
> Woah, I'm shocked at how quickly this actually got fixed, that's incredible. Appreciate everything y'all do

*cough cough* it's not normally this fast, you should be shocked

PR #2442, opened 2026-03-19
DogeisCut/FormatNumbers: don't trim off zeros when decimals = 0

### LINKED ISSUE #2436: [Format Numbers] Converting to AD Standard with 0 decimal places causes issues
Hello, I'm **ten_6044** from Scratch!

If you convert a number to AD standard with 0 decimal places, numbers with trailing zeroes will have their trailing zeroes trimmed.

### Steps to Reproduce
1. Open a new proejct, and load the "Format Numbers" extension.
2. Drag the "format number" block, and enter the following arguments:
    - **Number**: `10000`
    - **Notation**: `AD standard`
    - **Number of decimal places**: `0`
3. See result.

### Expected Result
The block would return `10K`.

### Actual Result
The block returns `1K` instead.
Image

GarboMuffin:
I suspect we can get rid of this whole .replace(...) string operation stuff with just numbers but decided to keep it safe for now

This was investigated with some help from Claude. Given just the prompt "investigate issue 2436" it was able to identify the problem and propose a fix (not a great one, but still it would've worked)

Closes #2436

### DIFF
--- a/extensions/DogeisCut/FormatNumbers.js
+++ b/extensions/DogeisCut/FormatNumbers.js
@@ -29,6 +29,11 @@
       this.decimalPlaces = 2;
     }
 
+    toFixedTrimmed(number, decimalPlaces) {
+      const formatted = number.toFixed(decimalPlaces);
+      return decimalPlaces > 0 ? formatted.replace(/\.?0+$/, "") : formatted;
+    }
+
     convertToADStandard(number, decimalPlaces = 2) {
       if (typeof number !== "number" || isNaN(number)) {
         return "Invalid Input";
@@ -86,7 +91,7 @@
       }
 
       if (Math.abs(number) < 1000) {
-        return number.toFixed(decimalPlaces).replace(/\.?0+$/, "");
+        return this.toFixedTrimmed(number, decimalPlaces);
       }
 
       const tier = Math.max(0, Math.floor(Math.log10(Math.abs(number)) / 3));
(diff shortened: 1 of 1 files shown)


Brackets-Coder:
> This was investigated with some help from Claude

What the agentic stuff can do is crazy

PPPDUD:
@GarboMuffin What do you mean by "investigated with Claude"? Do you mean that you had it help you debug or research, or did it actually write this code?

GarboMuffin:
Even Claude Haiku 4.5 (the smallest of their models) is able to identify the bug with just the GitHub issue text and nothing else - https://paste.sr.ht/~muffin/a8490b8f3e15626ef752e3c00767dcf5ea4bfd6f

(you could also just give it the issue number and it'll get the text on its own, but in this case it would see that the issue is already closed which would let it realize there is already a solution, so I didn't do that today)

Now, the change proposed there by Claude Haiku 4.5 is wrong. Claude Opus 4.6 and Claude Sonnet 4.6 both provide more valid fixes but neither of those are what I committed here.

GarboMuffin:
For a small bug fix to a niche extension that's doing pretty standard things that shows up plenty of times in training data, I don't see any reason to be against letting a language model write things if there's still a knowledgeable human in the loop.

PPPDUD:
> For a small bug fix to a niche extension that's doing pretty standard things that shows up plenty of times in training data, I don't see any reason to be against letting a language model write things if there's still a knowledgeable human in the loop.

Grumble grumble.

GarboMuffin:
You're welcome to elaborate

PPPDUD:
> You're welcome to elaborate

As you are probably aware, I have very strong opinions about LLMs, but I'm trying to ignore this case and get on with my life.

Brackets-Coder:
I don't want to add fuel to the fire, but kinda agree with @GarboMuffin here. AI can be helpful when used properly. Most of the time it's not used properly and most of the time when it is used properly it generates code I don't like. But it can be helpful (rubber duck debugging)

PR #2457, opened 2026-04-09
Skyhight173/json: add cache to improve performance

LegoBrainBiker:
the json extension can be slow because it needs to parse and stringify each time a block is called. this pr adds a cache so that if the json was recently created by the project, it can be looked up without using JSON.parse.


LegoBrainBiker:
!format

LegoBrainBiker:
!format

extensions/Skyhigh173/json.js
```diff
@@ -13,6 +13,25 @@
    */
 
   const vm = Scratch.vm;
+  const objectCache = {};
+  function parse(string) {
+    return (
+      objectCache[string] ??
```
GarboMuffin: never write code like this. clever code is bad and unreadable

extensions/Skyhigh173/json.js
```diff
@@ -13,6 +13,25 @@
...
+      (setTimeout(() => {
+        delete objectCache[string];
+      }, 100),
+      (objectCache[string] = JSON.parse(string)))
+    );
+  }
+  function stringify(object) {
+    const string = JSON.stringify(object);
+    objectCache[string] = object;
+    setTimeout(() => {
```
GarboMuffin: find a way to not be making a new timeout for every single string

LegoBrainBiker:
!format

extensions/Skyhigh173/json.js
LegoBrainBiker: is it acceptable without the timeout?

GarboMuffin: i was referring to `objectCache[string] ?? (objectCache[string] = JSON.parse(string));` which is very indicipherable to normal people

extensions/Skyhigh173/json.js
```diff
@@ -878,69 +888,69 @@
...
+        return stringify(json.concat(json2));
       } catch {
         return "";
       }
     }
 
     json_array_push({ item, json }) {
       try {
-        json = JSON.parse(json);
+        json = parse(json);
```
GarboMuffin: Here's another set of issue. parse() stores a reference to the object to reuse later. But in many of the blocks, we're taking these possibly-cached references and mutating them in place. So later blocks that change the same JSON string would actually change the same object internally, doing the wrong thing

So, we would need a way for that caching to not reuse objects incorrectly in these instances

LegoBrainBiker:
!format

Xeltalliv:
I don't think this should be merged.

Caching isn't always good, and in some cases people may want to avoid it. If people do want caching:
- Skyhight173 (the same person who made Skyhight173/json) made his official solution as a separate extension, which is awaiting to be merged: https://github.com/TurboWarp/extensions/pull/1066
- SharkPool also made his own ["Swift JSON (JSON and Array)" extension](https://sharkpools-extensions.vercel.app/extension-code/JSON-Array.js).

This extension is currently unique precisely because it does not do caching.

The main concern with caching is that it increases RAM usage. Especially with implementation like this, if project does a lot of operations synchronously (within "run without screen refresh" custom block), it may quite easy to run out of RAM and crash the browser tab on projects that previously ran fine.

SharkPool-SP:
I gotta agree with Xeltalliv here, this should be closed in favor of the other JSON extension PRs

LegoBrainBiker:
> I don't think this should be merged.
> 
> Caching isn't always good, and in some cases people may want to avoid it. If people do want caching:
> 
> * Skyhight173 (the same person who made Skyhight173/json) made his official solution as a separate extension, which is awaiting to be merged: [Object extension #1066](https://github.com/TurboWarp/extensions/pull/1066)
> * SharkPool also made his own ["Swift JSON (JSON and Array)" extension](https://sharkpools-extensions.vercel.app/extension-code/JSON-Array.js).
> 
> This extension is currently unique precisely because it does not do caching.
> 
> The main concern with caching is that it increases RAM usage. Especially with implementation like this, if project does a lot of operations synchronously (within "run without screen refresh" custom block), it may quite easy to run out of RAM and crash the browser tab on projects that previously ran fine.

anything that would use a significant amount of memory already wouldn't run "fine" the current implementation runs very slowly with 
[turbo-refrence.js](https://github.com/user-attachments/files/29781931/turbo-refrence.js)
repeated operations. the caches are also cleared often, so the memory impact aren't that insignificant. 

I personally don't use the JSON extension, instead using [my own object extension](https://github.com/user-attachments/files/29781931/turbo-refrence.js). I agree that the json extension is not useful in most projects, and could be replaced, but that's no reason not to make the existing json extension the best it can be.

SharkPool-SP:
The only problem with changing the existing extension is that it would *have* to implement breaking changes

LegoBrainBiker:
> The only problem with changing the existing extension is that it would _have_ to implement breaking changes

this pull request has no breaking changes, to a user, it would work exactly the same but faster.

SharkPool-SP:
> > The only problem with changing the existing extension is that it would _have_ to implement breaking changes
> 
> 
> 
> this pull request has no breaking changes, to a user, it would work exactly the same but faster.

The problem is it would only work faster for *preexisting* objects that are the exact same. Additionally it would hog RAM space like previously mentioned

SharkPool-SP:
So there's not much speed gain for regular usage.

LegoBrainBiker:
> > > The only problem with changing the existing extension is that it would _have_ to implement breaking changes
> > 
> > 
> > this pull request has no breaking changes, to a user, it would work exactly the same but faster.
> 
> The problem is it would only work faster for _preexisting_ objects that are the exact same. Additionally it would hog RAM space like previously mentioned

preexisting objects normally need to be reparsed whenever multiple items of an array/object are read. additionally, if an existing array or object is modified or indexed, my changes also cache those, so if for example if you want to apply multiple operations to an  object or array, it would only have to parse the initial object or array, not each time an item is modified. I cannot think of a single scenario which this would not speed up.

SharkPool-SP:
Yes but that quickly builds up in ram. And in those cases it's typically a one-time-use

LegoBrainBiker:
> Yes but that quickly builds up in ram. And in those cases it's typically a one-time-use

it only lasts for a short time.

GarboMuffin:
usually yes but not necessarily always

for example, a run w/o screen refresh would never be clearing its cache with the extension as-written

there's some ways you could address that. a max size cap in some way (probably try to account for big JSON objects and such) might make it tolerable w/o making things slower even in the worst case

i'd be chill with that kind of idea, done well

LegoBrainBiker:
that is true, but it still likely wouldn't use that much memory, and before these changes it would be much slower. as for implementing a size cap, I'm not sure I could create an effective way to store the order that they were used in to remove the last one.

SharkPool-SP:
I think the core issue is that we're caching objects that *have* to be used more than once or twice. Otherwise there's no utility to it. We're just temporarily storing things for no reason

SharkPool-SP:
Take this example:
image

We store 4 different objects where 3 of which end up being unused and taking unnecessary space.

SharkPool-SP:
IMO the smarter choice would be to check the block stack. 

I.e., if the json block currently running is in an input of another json block, return the raw json without stringifying or anything. Otherwise for all other blocks return stringified json

LegoBrainBiker:
I could make it so that blocks that "modify" like in your example remove their inputs from the cache quite simply. I don't think your method would be especially useful because that kind of thing would be more likely to be done in a loop.

SharkPool-SP:
Not really I've done it before in extensions. It just checks in the block function it's surrounding blocks and decides wether to return raw or stringified json

GarboMuffin:
a well-done string cache seems reasonable but 🤷‍♂️

SharkPool-SP:
See:

https://github.com/SharkPool-SP/SharkPools-Extensions/blob/main/extension-code/Rigidbodies.js#L815-L819

https://github.com/SharkPool-SP/SharkPools-Extensions/blob/main/extension-code/Rigidbodies.js#L870-L879

SharkPool-SP:
> a well-done string cache seems reasonable but 🤷‍♂️ 

Again, it will be pretty useless for one time use objects. Like ones procedurally generated in a loop

LegoBrainBiker:
> > a well-done string cache seems reasonable but 🤷‍♂️
> 
> Again, it will be pretty useless for one time use objects. Like ones procedurally generated in a loop

It will prevent those from needing to be parsed, and may 

is there any chance of this being merged eventually? I can stop working on it if that is the case.

SharkPool-SP:
Make a new or

GarboMuffin:
I think the concept is fine...

 - some sort of FIFO or LRU cache (either can probably work)
 - evict first (FIFO) or last used (LRU) entry/entries when data contained within the map exceeds some known size
 - that size is determined by the sum of the length of the stringification of the JSON - that way project processing large JSON won't use infinite memory
 - set that limit at like 10MB or something large - most projects will never hit it during their runtime
 - the cache updating and storing has minimal to no performance impact
 - No one has to rewrite their projects to use some other extension/different blocks/etc.

PR #2460, opened 2026-04-16
Change attribution from ClaytonTDM to clay.rip

clay-rip:
I've updated my online username in most places to "Clay" or "clay.rip" — this PR updates it in the TurboWarp extension gallery.


CubesterYT (approved):
- [X] This is a trivial change that doesn't require testing.

GarboMuffin:
realized that clay.rip is literally a URL. we generally want to avoid linking to non-scratch websites within that gallery

would you be okay with me changing it to "clay-rip" instead? (matching github) it might be less obviously a website in that case (kind of a silly thing to worry about, i know)

clay-rip:
yeah, of course

PR #2482, opened 2026-05-15
Add dectalk banner

Dogo6647:


CubesterYT (approved):
Looks good to me

Brackets-Coder (approved):
I realized after merging the extension that I forgot to mention it lacked a banner
I thought we weren't supposed to include text in the image but it's not english so I guess it's not a translation problem

PR #2484, opened 2026-05-16
NishiOwO/dectalk: a couple fixes

GarboMuffin:
- rename, elaborate in description
- instruct translators to not do anything funny with "DECtalk"
- first BSD3c so we should add the license txt


NishiOwO:
will update on my repository too

PR #2489, opened 2026-05-16
TheShovel/CanvasEffects: don't scale render size block by DPI

CST1229:
It's probably more expected behavior for the block to keep a fixed resolution than for it to scale with browser zoom and DPI.
(e.g it's probably often used for pixelated games that keep a resolution)

Before:
https://github.com/user-attachments/assets/fc01c2cf-c3c6-43d3-99e4-e1dd5f4376ca

After:
https://github.com/user-attachments/assets/d45b25a9-5e7e-4d77-953c-666866434446

### DIFF
--- a/extensions/TheShovel/CanvasEffects.js
+++ b/extensions/TheShovel/CanvasEffects.js
@@ -505,7 +505,7 @@
     renderscale({ X, Y }) {
       // The function normally expects a stage size and therefore scales by DPI.
       // However, this block is meant for a fixed pixel size
-      // (usually used in cojunction with the pixelated resize rendering mode).
+      // (usually used in conjunction with the pixelated resize rendering mode).
       // Therefore, scale it back according to the devicePixelRatio.
       const pixelRatio = window.devicePixelRatio || 1;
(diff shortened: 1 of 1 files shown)


Brackets-Coder (approved):
looks fine to me

The stage size is reset when entering fullscreen, changing the runtime size in settings, etc. I think this is intended though

CST1229:
> The stage size is reset when entering fullscreen, changing the runtime size in settings, etc. I think this is intended though

it might not be, but it would be more complicated from a design+implementation standpoint (e.g adding a "reset canvas render size" block, and finding some way to listen to or override the resize() resolution changes)

SharkPool-SP:
Do we think a toggle for backwards compat is needed?

CST1229:
> Do we think a toggle for backwards compat is needed?

do any projects even rely on this old behavior (instead of actually being *fixed* by the new one?)

SharkPool-SP:
Probably not I guess. If we ever get an issue about it, we can always just come back and add a block for it.

SharkPool-SP (approved):

SharkPool-SP:
I believe this also closes a couple issues

PR #2490, opened 2026-05-17
box2d: polygon mode is back

### LINKED ISSUE #42: Problem with [Polygon] option in "Griffpatch's 2D Box Physics"
The block, "Enable this [Polygon]" doesn't seem to really work. Instead, it puts the start flag into a loop, essentially meaning that the project freezes. However, when changing it to anything but [polygon] then the project will work again, whether or not the last box is [precision] or [normal]

other than that, the extension seems to be working quite well! I hope this helps with development

GarboMuffin:
How it works: https://muffin.ink/slop/polygon.html

tl;dr scan costume pixels to generate polygon instead of relying on actual SVG path elements as the original one from griffpatch

Demonstration of the difference:

[Screencast_20260516_223905.webm](https://github.com/user-attachments/assets/d14ff70d-bc0e-4053-9d4c-4ec92c02c9ea)


Brackets-Coder:
Interesting that Claude could create a prototype. The code looks great, but of course I'd be interested in more optimized/prettier/human-like versions of this that are more cohesive with the rest of the extension

On a side note, I'd be willing to work on getting Scratch.external into this thing. As for an external kripken made an emscripten port of the original C-based Box2d https://github.com/kripken/box2d.js. Same person that ported Ammo physics. I don't know if it's in parity with the version here (probably not) but we should likely just find a way to modularize and minify the library in the extension to host it on this repo

GarboMuffin:
[Screencast_20260517_121458.webm](https://github.com/user-attachments/assets/1ff7247b-7731-4b9c-9386-946a1da2c922)

Brackets-Coder:
Most impressive

GarboMuffin:
the state of this one is that it seems to work about as good as one would expect tracing image pixels to work

SharkPool-SP:
seems fine to me. I havent checked the full box2d codebase, only the given changes. Does the sillouette to polygon thing cache? If not it would be a good thing to add before merging.

extensions/box2d.js
```diff
@@ -13456,25 +14206,30 @@
...
+        } catch (e) {
+          traced = null;
+        }
+
+        if (traced && traced.length > 0) {
+          _definePolyFromHull(traced[0]);
+          allHulls = traced;
+        } else {
+          // Tracing produced nothing usable (empty or degenerate costume):
+          // fall back to the convex hull, same as "this costume" mode.
```
GarboMuffin, line 14132: thread to add caching

extensions/box2d.js
```diff
@@ -12240,6 +12240,753 @@
...
+    const coord = [0, 0];
+    let opaqueCount = 0;
+
+    // Sample each interior cell: map it to a [0, 1] texture coordinate, and
+    // store the result offset by (1, 1) so it lands inside the border.
+    for (let j = 0; j < gyI; j++) {
+      coord[1] = j / (gyI - 1);
+      for (let i = 0; i < gxI; i++) {
+        coord[0] = i / (gxI - 1);
+        if (skin.isTouchingLinear(coord)) {
```
GarboMuffin: thread to add effect transforms

extensions/box2d.js
SharkPool-SP: What do you mean? Like it exists or no?

GarboMuffin: means it doesn't right now

GarboMuffin: just not going to for now. it's not super clear what we would want to cache by

extensions/box2d.js
GarboMuffin: was added

GarboMuffin:
!format

extensions/box2d.js
SharkPool-SP: I'm saying cache the polygon so it doesn't re-make it every time

GarboMuffin: I understood that. I'm saying it's not clear what one would want to use as the key for that cache for what events for invalidating that cache. You're welcome to add caching or otherwise clean up some of the slop in here

SharkPool-SP: We could use a current costume asset ID joined with mutable effects?

PR #2495, opened 2026-05-19
style: update Comment Blocks extension color for continuity

ghost:
Replaces the jarring high-brightness color of the Comment Blocks extension with a muted blue color for continuity.

**From color #e4db8c to color #7683b0.**

block_5_19_2026-4_22_57 PM

### DIFF
--- a/extensions/Lily/CommentBlocks.js
+++ b/extensions/Lily/CommentBlocks.js
@@ -16,9 +16,7 @@
       return {
         id: "lmscomments",
         name: Scratch.translate("Comment Blocks"),
-        color1: "#e4db8c",
-        color2: "#c6be79",
-        color3: "#a8a167",
+        color1: "#7683b0",
         blocks: [
           /* eslint-disable extension/should-translate */
           {


Brackets-Coder:
Screenshot 2026-05-19 at 5 53 09 PM

I'm not opposed to this change. However, if I remember correctly, the extension colors were chosen to mimic the existing comments, so I don't think the color was intended to be changed. If it's desired, one can get a less jarring color with Settings > Block Colors > Dark (Beta)

LilyMakesThings:
I don't agree with this change.

Brackets-Coder:
> I don't agree with this change.

Same here—not opposed, but don't agree

ghost:
> I'm not opposed to this change. However, if I remember correctly, the extension colors were chosen to mimic the existing comments, so I don't think the color was intended to be changed.

Scratch's native comment system is designed to float alongside scripts, not to wrap directly into the block stack itself. When you adapt that high-brightness color into a physical block that integrates directly into the sprite stack, it creates a harsh visual break from the rest of the more colorful, less overwhelmingly bright blocks. Out of human instinct to filter out high-contrast, jarring elements, this can actually cause users to visually skip past the comments entirely, including me. 

> If it's desired, one can get a less jarring color with Settings > Block Colors > Dark (Beta)

Requiring users to globally alter their entire block color theme using a Beta feature just to fix a single extension's default readability issue isn't a great out-of-the-box UX. The goal here was to make the default experience cohesive. 

I'm definitely not the only one who finds the current default high-contrast color difficult to work with inline. If the team is completely set on keeping the current color, I can close this out, but I think addressing the default continuity is worth considering.

ghost:
Maybe I can compromise and switch it to a darker version of the current color while still having it fit in with the rest of the standard blocks.

Brackets-Coder:
@TurboWarp/extension-reviewers ?

CubesterYT:
I'm not a big fan, the current color is great

PR #2498, opened 2026-05-20
Add File Processor and Converter extension

as-designer:
**_File Processor and Converter_** is an _unsandboxed_ extension that enables users to convert popular **image formats** (such as PNG, WebP and JPG), **document formats** (such as TXT, JSON, MarkDown, HTML), **audio formats** (such as MP3 and WAV) and **data formats** (such as JSON and CSV) directly within **_TurboWarp_**. It also features a **live microphone utility**, allowing creators to record audio from their hardware and download it instantly. Because it requires _access to browser file APIs_ and _hardware recording permissions_, it is explicitly configured to run in _unsandboxed_ mode.


as-designer:
!format

SharkPool-SP:
This does not actually convert file formats, just the naming. + There is a separate PR for a microphone recording extension (which should be a separate extension)

Also, this extension does nothing more than the existing Files extension

GarboMuffin:
it definitely did *some* conversions somewhere

PR #2517, opened 2026-05-30
Add decaying sine wave maker extension

greenkidneybeans:
This block generates a Damped Sine Wave based on five parameters: Amplitude(Wave size relative to frequency), Frequency(how fast the wave oscillates), Decay multiplier(for example, the normal constant would be 1 or -1 if you do -0.5, the decay is slower, if you do for example 500, it would go flat in a near instant.), Speed(the amount of steps forward on the x axis it takes each iteration), StartX(the starting X), And StartY(the starting Y). This uses a formula that creates a sine wave instantly and is expected to be visually mapped using Pen.

### DIFF
--- a/sinewave.js
+++ b/sinewave.js
@@ -1,0 +1,55 @@
+(function() {
+    class SoundWaveDecay {
+        getInfo() {
+            return {
+                id: 'soundwavedecay',
+                name: 'Sine sound wave',
+                color1: '#417505',
+                color2: '#ffffff',
+                blocks: [
+                    {
+                        opcode: 'followSineWave',
+                        blockType: Scratch.BlockType.COMMAND,
+                        arguments: {
+                            Amplitude: { type: Scratch.ArgumentType.NUMBER },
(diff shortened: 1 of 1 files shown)


SharkPool-SP:
1) This extension is easily re-createable via operator math or the Tween extension
2) Theres only 1 block in this extension, and it doesnt even seem to work either. The sprite just instantly shoots to the corner

Sorry to close this...

greenkidneybeans:
> 1. This extension is easily re-createable via operator math or the Tween extension
> 2. Theres only 1 block in this extension, and it doesnt even seem to work either. The sprite just instantly shoots to the corner
> 
> Sorry to close this...

The sprite shoots to the corner because it is made to process instantly without you knowing. It is meant to be visually tracked if needed by having pen down with the pen extension.

GarboMuffin:
This is better handled by a script inside the project which can apply any and all customizations it wants, as this block as-is is just not particularly useful

PR #2382, opened 2026-01-09
New extension: Mesh

ampelectrecuted:
Per the docs:
> Mesh lets you communicate with other projects running in the same browser or in TurboWarp Desktop, using special versions of variables and broadcasts. This allows you to for example, create a game that talks to another project running at the same time.

This was [actually a real feature in Scratch 1.4](https://en.scratch-wiki.info/wiki/Mesh).

TODO: ws integration (though that can be added at a later time).

PR #2518, opened 2026-05-30
New extension: Cross-Project

ampelc:
Per the docs:
> Mesh lets you communicate with other projects running in the same browser or in TurboWarp Desktop, using special versions of variables and broadcasts. This allows you to for example, create a game that talks to another project running at the same time.

This was [actually a real feature in Scratch 1.4](https://en.scratch-wiki.info/wiki/Mesh).

TODO: ws integration (though that can be added at a later time).

This is a copy of #2382 which is from an account I am now locked out of.


SharkPool-SP:
I dont think 'Mesh' is the greatest name for this

GarboMuffin:
Is this meaningfully compatible or equivalent with the Scratch 1.4 feature enough to justify giving it the same name?

ampelc:
> Is this meaningfully compatible or equivalent with the Scratch 1.4 feature enough to justify giving it the same name?

actually not really

since mesh might be confusing to some anyways i've renamed this to cross-project

Brackets-Coder:
!format

extensions/8to16/mesh.js
```diff
@@ -0,0 +1,352 @@
...
+  let connectedUsers = 0;
+  let finishedWaits = {};
+
+  if (!vm.runtime.extensionStorage["8to16mesh"])
+    vm.runtime.extensionStorage["8to16mesh"] = { messages: [], variables: {} };
+  vm.runtime.on("RUNTIME_DISPOSED", () => {
+    vm.runtime.extensionStorage["8to16mesh"] = { messages: [], variables: {} };
+  });
+  vm.runtime.on("PROJECT_LOADED", () => {
+    vm.runtime.extensionManager.refreshBlocks();
```
SharkPool-SP, line 26: `vm.extensionManager.refreshBlocks()` takes in arguments to refresh a specific extension palette. For optimization, use your extension id as the first argument

extensions/8to16/mesh.js
```diff
@@ -0,0 +1,352 @@
...
+  const getMeshages = () =>
+    vm.runtime.extensionStorage["8to16mesh"].messages ?? [];
+  const addMeshage = (name) => {
+    if (getMeshages().includes(name)) return;
+    if (name === "") return;
+    vm.runtime.extensionStorage["8to16mesh"].messages = [
+      ...getMeshages(),
+      name,
+    ];
+    vm.extensionManager.refreshBlocks();
```
SharkPool-SP, line 39: same here

extensions/8to16/mesh.js
```diff
@@ -0,0 +1,352 @@
...
+  };
+  const delMeshage = (name) => {
+    if (!getMeshages().includes(name)) return;
+    vm.runtime.extensionStorage["8to16mesh"].messages = getMeshages().filter(
+      (n) => n !== name
+    );
+    vm.extensionManager.refreshBlocks();
```
SharkPool-SP, line 46: same here

extensions/8to16/mesh.js
```diff
@@ -0,0 +1,352 @@
...
+  const getMeshVars = () =>
+    vm.runtime.extensionStorage["8to16mesh"].variables ?? {};
+  const addMeshVar = (name) => {
+    if (Object.keys(getMeshVars()).includes(name)) return;
+    if (name === "") return;
+    vm.runtime.extensionStorage["8to16mesh"].variables = {
+      ...getMeshVars(),
+      [name]: "",
+    };
+    vm.extensionManager.refreshBlocks();
```
SharkPool-SP, line 59: same here

extensions/8to16/mesh.js
```diff
@@ -0,0 +1,352 @@
...
+  };
+  const delMeshVar = (name) => {
+    if (Object.keys(getMeshVars()).includes(name)) return;
+    vm.runtime.extensionStorage["8to16mesh"].variables = Object.fromEntries(
+      Object.entries(getMeshVars()).filter((v) => v[0] !== name)
+    );
+    vm.extensionManager.refreshBlocks();
```
SharkPool-SP, line 66: same here

extensions/8to16/mesh.js
```diff
@@ -0,0 +1,352 @@
...
+
+  // Message utilities
+  const getMeshages = () =>
+    vm.runtime.extensionStorage["8to16mesh"].messages ?? [];
+  const addMeshage = (name) => {
+    if (getMeshages().includes(name)) return;
+    if (name === "") return;
+    vm.runtime.extensionStorage["8to16mesh"].messages = [
+      ...getMeshages(),
+      name,
```
SharkPool-SP, line 37: bad practice, just add the message via

`messages.push(name)`

Also fix spelling for `message`, not `meshages`

extensions/8to16/mesh.js
```diff
@@ -0,0 +1,352 @@
...
+    ];
+    vm.extensionManager.refreshBlocks();
+  };
+  const delMeshage = (name) => {
+    if (!getMeshages().includes(name)) return;
+    vm.runtime.extensionStorage["8to16mesh"].messages = getMeshages().filter(
```
SharkPool-SP, line 43: bad practice, remove the message via

`messages.splice(name_index, 1)`

Also fix spelling for `message`, not `meshages`

extensions/8to16/mesh.js
```diff
@@ -0,0 +1,352 @@
...
+      (n) => n !== name
+    );
+    vm.extensionManager.refreshBlocks();
+  };
+
+  // Var utilities
+  const getMeshVars = () =>
+    vm.runtime.extensionStorage["8to16mesh"].variables ?? {};
+  const addMeshVar = (name) => {
+    if (Object.keys(getMeshVars()).includes(name)) return;
```
SharkPool-SP, line 53: you can just check if the key returns undefined

extensions/8to16/mesh.js
```diff
@@ -0,0 +1,352 @@
...
+    if (name === "") return;
+    vm.runtime.extensionStorage["8to16mesh"].variables = {
```
SharkPool-SP, line 55: bad practice, just add the variable via

`variables[name] = ""`

extensions/8to16/mesh.js
```diff
@@ -0,0 +1,352 @@
...
+      ...getMeshVars(),
+      [name]: "",
+    };
+    vm.extensionManager.refreshBlocks();
+  };
+  const delMeshVar = (name) => {
+    if (Object.keys(getMeshVars()).includes(name)) return;
+    vm.runtime.extensionStorage["8to16mesh"].variables = Object.fromEntries(
```
SharkPool-SP, line 63: bad practice, just delete the variable via

`delete variables[name]`

extensions/8to16/mesh.js
```diff
@@ -0,0 +1,352 @@
...
+        // broadcast and wait handling
+        if (data.willWait) {
+          await new Promise((resolve) => {
+            const poll = () => {
+              if (
+                hats.filter((thread) => vm.runtime.threads.includes(thread))
+                  .length === 0
+              )
+                resolve();
+              else setTimeout(poll, 5);
```
SharkPool-SP, line 85: instead of setTimeouts you can use `vm.runtime.on("AFTER_EXECUTE", function)`, which runs after every tick execution. Then when all threads finish use `vm.runtime.off("AFTER_EXECUTE", function)

extensions/8to16/mesh.js
```diff
@@ -0,0 +1,352 @@
...
+      ScratchBlocks.prompt(
+        Scratch.translate("Remove cross-project message named:"),
+        "",
+        delMeshage,
+        Scratch.translate("Remove Message"),
+        "broadcast_msg"
+      );
+    }
+    newVar() {
+      // taken from SharkPool/Camera
```
SharkPool-SP, line 147: We dont need to repeat this comment, just once is fine

extensions/8to16/mesh.js
```diff
@@ -0,0 +1,352 @@
...
+        Scratch.translate("Remove Variable"),
+        "broadcast_msg"
+      );
+    }
+    getInfo() {
+      return {
+        id: "eightxtwoMesh",
+        name: Scratch.translate("Cross-Project"),
+        docsURI: "https://extensions.turbowarp.org/8to16/mesh",
+        color1: "#4cdab2",
```
SharkPool-SP, line 175: I would make the block color a bit darker so you can see the block text better

extensions/8to16/mesh.js
```diff
@@ -0,0 +1,352 @@
...
+      vm.runtime.startHats("eightxtwoMesh_when", { BROADCAST }); // TODO: add wait here
+      bc.postMessage({
+        type: "broadcast",
+        name: BROADCAST,
+        willWait: true,
+      });
+      await new Promise((resolve) => {
+        const poll = () => {
+          if (finishedWaits[BROADCAST] >= connectedUsers) resolve();
+          else setTimeout(poll, 25);
```
SharkPool-SP, line 322: same thing with runtime.on("AFTER_EXECUTE"), alternatively use `util.yield()`, which I can give examples for.

extensions/8to16/mesh.js
```diff
@@ -0,0 +1,352 @@
...
+
+  if (!Scratch.extensions.unsandboxed) {
+    throw new Error("Cross-Project Broadcast extension must run unsandboxed");
+  }
+  const vm = Scratch.vm;
+
+  // Init the broadcastchannel
+  const bc = new BroadcastChannel("extensions.turbowarp.org/8to16/mesh");
+  let connectedUsers = 0;
+  let finishedWaits = {};
```
SharkPool-SP, line 18: change `{}` to `Object.create(null)` since keys are set by users, which means they can set dangerous keys like `toString` or `prototype`

extensions/8to16/mesh.js
```diff
@@ -0,0 +1,352 @@
...
+
+  if (!vm.runtime.extensionStorage["8to16mesh"])
+    vm.runtime.extensionStorage["8to16mesh"] = { messages: [], variables: {} };
```
SharkPool-SP, line 21: change `{}` to `Object.create(null)` since keys are set by users, which means they can set dangerous keys like `toString` or `prototype`

extensions/8to16/mesh.js
```diff
@@ -0,0 +1,352 @@
...
+  vm.runtime.on("RUNTIME_DISPOSED", () => {
+    vm.runtime.extensionStorage["8to16mesh"] = { messages: [], variables: {} };
```
SharkPool-SP, line 23: change `{}` to `Object.create(null)` since keys are set by users, which means they can set dangerous keys like `toString` or `prototype`

SharkPool-SP (changes requested):
I really like this extension, but do these first. I can also make a better banner for you if you want

extensions/8to16/mesh.js
ampelc: meshages was originally meant to be a bit of  a pun when this was called mesh

SharkPool-SP: Seemed more like a spelling mistake to me but whatever. You might wanna change it to message and keep the original as a comment.


You don't have to though.

PR #2537, opened 2026-06-14
Add missing credits and require credits for all extensions moving forward

GarboMuffin:
To avoid certain types of questions

### DIFF
--- a/development/builder.js
+++ b/development/builder.js
@@ -264,13 +264,12 @@ class ExtensionFile extends BuildFile {
       );
     }
 
-    const allCredits = [
-      ...metadata.by,
-      ...metadata.original
-    ];
+    const allCredits = [...metadata.by, ...metadata.original];
 
     if (allCredits.length === 0) {
-      throw new Error("Missing any credits. Add a // By: or // Original: comment to the header.");
+      throw new Error(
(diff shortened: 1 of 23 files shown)


LilyMakesThings:
Don't forget lab/text

development/builder.js
```diff
@@ -264,7 +264,16 @@ class ExtensionFile extends BuildFile {
...
     }
 
-    for (const person of [...metadata.by, ...metadata.original]) {
+    const allCredits = [
+      ...metadata.by,
+      ...metadata.original
+    ];
+
+    if (allCredits.length === 0) {
+      throw new Error("Missing any credits. Add a // By: or // Original: comment to the header.");
```
Brackets-Coder: Looks alright, I would just change the note here to something like
```suggestion
      throw new Error("Missing credits. Add a // By: or // Original: comment to the header.");
```

or

```suggestion
      throw new Error("Has no credits. Add a // By: or // Original: comment to the header.");
```

Brackets-Coder (changes requested):

development/builder.js
GarboMuffin: removed "any "

PR #2540, opened 2026-06-18
Gemini

deevbog-ui:
Allows you to add Google Gemini integration to your project


GarboMuffin:
hi, appreciate your interest but we do not think this will be merged because

 - anything that requires people to provide their own API key is going to be a bad experience and will 100% result in people accidentally sharing keys and racking up absurd expenses which then will be blamed on us
 - your extension appears to be written by an AI without appropriate disclosure or signs of human review
 - your extension does not follow several requirements outlined in CONTRIBUTING.md and elsewhere such as being written in English, supporting translations, or being in the right folder

PR #2558, opened 2026-07-03
utilities: Fix boolean casting in string to boolean block

### LINKED ISSUE #1986: Utilities.js doesn't use the Scratch.Cast API
Pertains to https://github.com/TurboWarp/extensions/issues/1810

As pointed out in https://github.com/TurboWarp/extensions/issues/1985#issuecomment-2688247581, [the Utilities extension](https://github.com/TurboWarp/extensions/blob/master/extensions/utilities.js) has several instances that use `.toString` over `Scratch.Cast.toString`.

Furthermore, the following block has possibly unintended functionality due to not casting the input to a boolean before returning:

https://github.com/TurboWarp/extensions/blob/f9a53719e4de123c2a2cae43db1dcb264bc8f7bb/extensions/utilities.js#L384-L386

SharkPool-SP:
Resolves: https://github.com/TurboWarp/extensions/issues/1986

### DIFF
--- a/extensions/utilities.js
+++ b/extensions/utilities.js
@@ -392,7 +392,7 @@
     }
 
     stringToBoolean({ STRING }) {
-      return STRING;
+      return Scratch.Cast.toBoolean(STRING);
     }
 
     regexReplace({ STRING, REGEX, NEWSTRING }) {


Brackets-Coder (approved):
This should've been fixed like 18 months ago, nice one-liner

Brackets-Coder:
The only concerns i might have are the rare occasions where a project might rely on this not returning a boolean, but I significantly doubt that since the block is a boolean anyway and that's what most users would expect to be returned

SharkPool-SP:
Relying on a bug for a coupler block that goes into Boolean slots.

If someone is using this for another behaviour (which is incorrect) than that's on them.

GarboMuffin:
Ok, sure

GarboMuffin (approved):

PR #2563, opened 2026-07-05
devcontainer: Add configs for Dev Containers

ghost:
Adds configuration files for Dev Containers, a VS Code feature that uses Docker to create disposable containers for development (as the name suggests). **It also configures GitHub Codespaces, to a degree** (which uses Dev Containers in the backend).

I used `mcr.microsoft.com/devcontainers/javascript-node:4-22-trixie` as the base image, which is Node.js v22. It also automatically exposes the development server port `8000`, and runs `npm ci` after creation.


GarboMuffin:
You can use that yourself and put it into .git/info/exclude to avoid committing it

PR #2564, opened 2026-07-06
SamuelLouf/Geolocation: mark as not supported in desktop app

GarboMuffin:
In Electron, the geolocation APIs don't work unless we provide a GOOGLE_API_KEY with billing enabled for Google to charge per request. This key has to be embedded in plain text in every copy of the desktop app too, so it's public and anyone could start putting a lot of usage on it overnight. Don't really want to fund that, so it's just not going to work there. That'll include the desktop app and packaged Electron projects.

For the desktop app, making this work on macOS would also probably require the location services entitlement. I don't really want to try explaining to App Store Review why the app needs location services without having some good explanation of what people are doing with this thing.

### DIFF
--- a/extensions/SamuelLouf/Geolocation.js
+++ b/extensions/SamuelLouf/Geolocation.js
@@ -95,7 +95,7 @@
    * @returns {boolean}
    */
   function isElectron() {
-    return navigator.userAgent.includes('Electron');
+    return navigator.userAgent.includes("Electron");
   }
 
   /**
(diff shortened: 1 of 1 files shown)


GarboMuffin:
!format

GarboMuffin:
It was already disabled in the desktop app via security manager: https://github.com/TurboWarp/desktop/blob/f0b8f21854222e889c8d7931b5e024d032464f37/src-renderer-webpack/editor/gui/desktop-hoc.jsx#L73

And the desktop app's permission checks deny all unneeded permissions, such as location.

This change makes it properly report as not supported at all (not just not allowed)

PR #2568, opened 2026-07-06
Resolve all border disputes and achieve world peace

GarboMuffin:
It took just over 24 hours for me to receive a long essay from someone complaining about which combination of China and Taiwan got highlighted. No matter what combination we chose, someone is inevitably going to be upset. By removing the map, we ensure that everyone ends up equally unhappy.


GarboMuffin:
If you think the solid green background looks bad, please make a better one just don't try to use a map of the world

SharkPool-SP:
In all honesty who gives a shit 😭

LordCat0:
World peace

GarboMuffin:
The people (multiple) who have complained to me about this certainly do seem to care.

SharkPool-SP:
Doesn't seem like something that should be your problem to solve though imo

CubesterYT:
Why didn't we just leave the map unhighlighted? with no locations marked

GarboMuffin:
> Doesn't seem like something that should be your problem to solve though imo

I don't want to be getting emails from the many (many) people using TurboWarp Desktop in China complaining about this

> Why didn't we just leave the map unhighlighted? with no locations marked

It leads to questions like why Taiwan and other places are just not even on the map at all

I don't care to deal with that

qpwq1:
世界和平！
但我怎么没看到有什么地图

PR #2587, opened 2026-07-12
A lot of fixes

### LINKED ISSUE #483: Sensing + speech recognition isn't detecting if recording is on even though it is

GarboMuffin:
yeah


SharkPool-SP:
Does this close any issues?

extensions/Lily/Assets.js
GarboMuffin, line 1: todo: figure out what happened here for the entire file to get marked as changed

GarboMuffin: the line endings in the original one were bad. they got fixed here

PR #2589, opened 2026-07-13
Remove extension-reviewer codeowners

GarboMuffin:
Apparently GitHub just sent out 147 emails because of this thing, so we're going to stop using it now


Brackets-Coder:
I didn't mean to make a problem of something that wasn't a problem, there was a quick solution and I kind of like getting emails about what's going on

qpwq1:
?

Issue #1358, opened 2024-03-15
[Suggestion] Can anyone make local storage extension able to get all keys not set to 0 or blank stored in local storage?


ASTHEBEST:
Can anyone make local storage extension able to get all keys not set to 0 or blank stored in local storage? So then you would know which keys have a value stored in them.

Drago-Cuven:
bump.

GarboMuffin:
resubmit with more details about intended use

SharkPool-SP:
I'll be handling this

PR #2598, opened 2026-07-20
local-storage.js -- Add ability to 'get all keys' in local storage

SharkPool-SP:
Resolves:
#1358

### DIFF
--- a/docs/local-storage.md
+++ b/docs/local-storage.md
@@ -44,6 +44,12 @@ Or delete everything stored in the namespace:
 delete storage :: #0FBD8C
 ```
 
+If you ever loose track of what variables you previously set in storage, you can use the below block. It will return an array of all variables (or 'keys') in storage.
+
+```scratch
+(all keys from storage :: #0FBD8C)
+```
+
 ## Performance
 
 The local storage extension is inevitably slower than regular variables.
--- a/docs/local-storage.md
+++ b/docs/local-storage.md
@@ -44,11 +44,19 @@ Or delete everything stored in the namespace:
 delete storage :: #0FBD8C
 ```
 
-If you ever loose track of what variables you previously set in storage, you can use the below block. It will return an array of all variables (or 'keys') in storage.
+These next blocks are used to inspect and iterate through all namespace items.
 
 ```scratch
-(all keys from storage :: #0FBD8C)
+(number of keys in storage :: #0FBD8C)
 ```
+
+This block reports the total number of stored variables.
+
+```scratch
+(key # [1] in storage :: #0FBD8C)
+```
+
+This block reports the name of a key at a specific index position.
 
 ## Performance
 
--- a/extensions/local-storage.js
+++ b/extensions/local-storage.js
@@ -200,6 +200,12 @@
             },
           },
           {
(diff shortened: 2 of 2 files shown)


SharkPool-SP:
!format

GarboMuffin:
i think i've shared my thoughts on adding new blocks that returns lists as JSON ...

SharkPool-SP:
> i think i've shared my thoughts on adding new blocks that returns lists as JSON ...

And several blocks in extensions in your gallery do this, there isn't a proper alternative.


I don't know what you expect it should do.

Brackets-Coder:
Tend to agree with Garbo here. JSON is hard to work with without the JSON extension. Native Scratch is usually space-delimited for similar blocks. Of course, when doing something like this, you could run into parsing issues when the key or item contains the delimiter in its name, e.g., `item 1 item 2 item 3`, where instead of `item 1` being parsed it may come out as `item` `1`.

Screenshot 2026-07-30 at 9 13 41 AM

SharkPool-SP:
Then what's the better alternative?


The scratch method has its flaws with commas or spaces.


And for this block to actually be useable they'd have to split the output to an array or to different parts using text extension (which is an extra dependency regardless)

Brackets-Coder:
I don't know. With a bit of extra work you can split by spaces pretty easily in native Scratch, but it doesn't really work when the values contain the delimiter, which is why I think JSON is better. Unfortunatley, it's another extension users would have to load and potentially is more confusing for some users

SharkPool-SP:
> I don't know. With a bit of extra work you can split by spaces pretty easily in native Scratch, but it doesn't really work when the values contain the delimiter, which is why I think JSON is better. Unfortunatley, it's another extension users would have to load and potentially is more confusing for some users

And like I said, if you split by spaces, users would have to add the text extension to use the block to split text by a delimiter.

That is unless the user wants to make their own parser, which is bulky and in the long run a waste of their time if they could use the alternative.

SharkPool-SP:
Let's not forget some extensions in the gallery already do this... plus it's always reality that if the user doesn't understand the block, they won't use it which is fine.

Brackets-Coder:
Okay

ScratchFakemon:
I agree with SharkPool here; plenty of extensions have blocks that return information as JSON, and there's a JSON extension for a reason. The list reporters aren't particularly good at conveying information anyway, so a space-separated list of keys won't be much different.
image image
image  image

GarboMuffin:
i'm not proposing a space separator

GarboMuffin:
the scratch approach to lists is not big json

the scratch approach to lists is blocks like "length of list" and "item 1 of list"

no reason you can't have "number of things in storage" and "key of item 1 in storage" equivalent

or even a "store keys in [list]" block

CST1229:
> i'm not proposing a space separator

well thats what we all assumed

~~should have specified earlier~~

SharkPool-SP:
Exactly

GarboMuffin:
Ok

SharkPool-SP:
!format

SharkPool-SP:
@GarboMuffin @Brackets-Coder fixed

PR #2599, opened 2026-07-20
Lily/SoundExpanded -- Add 'stop all sounds in sprite' block

### LINKED ISSUE #1650: Sound Expanded: Stop All Sound In Certain Sprites
A Stop All Sound in certain selected sprites. 
There are some certain chases we're we made a Looped audio environment like Forest Sounds, Fireplace Sound, Waterfall sounds and we only need to stop some sprites sound/ambience.

Ofc there was A Stop Sound [] Block
But...
image
I don't think this is efficient

SharkPool-SP:
image

Resolves:
#1650

### DIFF
--- a/extensions/Lily/SoundExpanded.js
+++ b/extensions/Lily/SoundExpanded.js
@@ -180,6 +180,18 @@
             extensions: ["colours_sounds"],
           },
           {
+            opcode: "stopAllSpriteSounds",
+            blockType: Scratch.BlockType.COMMAND,
+            text: Scratch.translate("stop all sounds in [TARGET]"),
+            arguments: {
+              TARGET: {
+                type: Scratch.ArgumentType.STRING,
+                menu: "targets",
+              },
+            },
+            extensions: ["colours_sounds"],
+          },
+          {
             opcode: "pauseSounds",
             blockType: Scratch.BlockType.COMMAND,
             text: Scratch.translate("pause all sounds"),
@@ -494,6 +506,23 @@
       const soundId = sprite.sounds[index].soundId;
       const soundBank = sprite.soundBank;
       soundBank.stop(target, soundId);
+    }
+
+    stopAllSpriteSounds(args, util) {
+      let target = Scratch.vm.runtime.getSpriteTargetByName(args.TARGET);
+      if (args.TARGET === "_myself_") target = util.target;
+      if (args.TARGET === "_stage_") target = runtime.getTargetForStage();
+      if (!target) return;
+
+      const sprite = target.sprite;
+      const soundBank = sprite.soundBank;
+
+      const allSounds = Object.values(soundBank.soundPlayers);
+      for (const sound of allSounds) {
+        if (sound.isPlaying) {
+          soundBank.stop(target, sound.id);
+        }
+      }
     }
 
     pauseSounds(args, util) {
--- a/extensions/Lily/SoundExpanded.js
+++ b/extensions/Lily/SoundExpanded.js
@@ -515,17 +515,17 @@
     stopAllSpriteSounds(args, util) {
       const targets = [];
       switch (args.TARGET) {
-        case '_myself_':
+        case "_myself_":
           targets.push(util.target);
           break;
-        case '_stage_':
+        case "_stage_":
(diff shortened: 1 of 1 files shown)


SharkPool-SP:
!format

GarboMuffin:
This block offers no way to stop sounds being played by clones

SharkPool-SP:
Is it not:

'stop all sounds in (myself)'

GarboMuffin:
That's better than nothing, at least

SharkPool-SP:
> That's better than nothing, at least

That's an option in the block what do you mean:
image

GarboMuffin:
I know. The question was more about whether "stop all sounds in sprite1" should stop sounds from the clones too.

SharkPool-SP:
Ah, I could add that.

Also you should just start yolo merging my fix PRs... the other contributors seem inactive

GarboMuffin:
I might if no one else does

GarboMuffin:
I've also thought about just lowering it to 1 required review for changes to existing extensions

SharkPool-SP:
image

SharkPool-SP:
!format

Brackets-Coder:
> Ah, I could add that.
> 
> Also you should just start yolo merging my fix PRs... the other contributors seem inactive

I'm still alive but we definitely need more review activity around here to keep things running smoothly

extensions/Lily/SoundExpanded.js
```diff
@@ -496,6 +512,32 @@
...
 
+    stopAllSpriteSounds(args, util) {
+      const targets = [];
+      if (args.TARGET === "_myself_") targets.push(util.target);
+      else if (args.TARGET === "_stage_")
+        targets.push(runtime.getTargetForStage());
+      else if (args.TARGET === "_myfamily_")
+        targets.push(...util.target.sprite.clones);
+      else {
+        targets.push(Scratch.vm.runtime.getSpriteTargetByName(args.TARGET));
```
Brackets-Coder: Would this be clearer as a switch? These repeated conditionals are kinda hard to read for me

Brackets-Coder (commented):
Looks okay, but I'm in a quiet environment at the moment and don't want to make a bunch of noise testing so I'm going to probably test this in a few hours

extensions/Lily/SoundExpanded.js
SharkPool-SP: It would be twice the size if it were a switch.

Brackets-Coder: I see

SharkPool-SP: This is more of a personal gripe but switch case should only be implemented if you have more than three cases, excluding a default handler

The format for switch case is pretty bulky for a low number of cases

Brackets-Coder: Generally I tend to agree, but really it's a matter of preference. I just feel like there's a much easier way to refactor it, seeing the `args.TARGET ===` and `targets.push` on every line just makes it harder to read for me.

SharkPool-SP: Ight ill change it

Brackets-Coder (approved):
Works great

extensions/Lily/SoundExpanded.js
Brackets-Coder: Doesn't have to be a switch, it was just a suggestion

Brackets-Coder: Claude suggests a lookup map but that just feels wrong in this scenario

SharkPool-SP: L too late lol

SharkPool-SP:
!format

Brackets-Coder:
> the other contributors seem inactive

At this point it's just down to you and me, eh? 😅
Others are still around but less frequently

SharkPool-SP:
> > the other contributors seem inactive
> 
> At this point it's just down to you and me, eh? 😅

Preeeeetty muchg

SharkPool-SP:
@GarboMuffin

CubesterYT (approved):
Seems ok

Issue #1682, opened 2024-09-13
When getting a file via data URL from the zip extension, the "prefix" is always "data:application/octet-stream;base64,"


snowboyz0825:
This doesn't seem like it should be a problem, but most extensions that handle dataURL require it to be the proper "prefix." You can work around it by splitting it off and joining it with the proper prefix, but it's annoying, clunky, and counter-intuitive.

PR #2600, opened 2026-07-20
CST1229/zip.js -- Use correct MIME type when reading file Data.URIs

SharkPool-SP:
Resolves:
#1682

Before:
image

After:
image

Not exactly a behavior-changing commit. It doesnt break compatibility since extensions such as Skins would reject the old output whereas now it will accept it.


SharkPool-SP:
!format

extensions/CST1229/zip.js
```diff
@@ -4,12 +4,41 @@
...
+    THE SOFTWARE IS PROVIDED 'AS IS', WITHOUT WARRANTY OF ANY KIND,
+    EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+    MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+    IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+    CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+    TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+    SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+  */
+  const MimeTypes = await Scratch.external.importModule(
+    "https://cdn.jsdelivr.net/npm/mime-types@2.1.35/+esm"
```
GarboMuffin: Several issues here:

 - you didn't run the build server to update the dependency metadata (not optional)
 - +esm includes import within the file; the inliner does not handle those so this wouldn't work offline
 - the comment "Do NOT use SRI with dynamically generated files!" within the linked file indicates that its SHA checksum is not constant, but the dependency metadata requires a constant checksum

SharkPool-SP: Fixed

SharkPool-SP:
!format

Drago-Cuven:
I'd say this is a good change. Does it understand ALL file types? Or no. 

(I'm hoping it understands code files like js, lua, rb, py, etc etc)

SharkPool-SP:
> I'd say this is a good change. Does it understand ALL file types? Or no.
> 
> (I'm hoping it understands code files like js, lua, rb, py, etc etc)

whichever the database supports, which includes js and lua...

GarboMuffin (approved):

SharkPool-SP:
@Brackets-Coder

extensions/CST1229/zip.js
```diff
@@ -4,12 +4,56 @@
...
+      if (!dbResponse.ok) return [];
+
+      // We need to convert this table to a lookup array with
+      // mime types with applicable file type extensions.
+      const mimeTable = await dbResponse.json();
+      return Object.entries(mimeTable)
+        .filter((m) => m[1].extensions !== undefined)
+        .map((m) => [m[0], m[1].extensions]);
+    } catch {
+      console.warn("Couldnt not retrieve data from Mime DB!");
```
Brackets-Coder:
```suggestion
      console.warn("Could not retrieve data from Mime DB!");
```

Brackets-Coder (changes requested):

PR #2604, opened 2026-07-22
DT/Camera.js -- Include '(Buggy)' in the name

SharkPool-SP:
Labeling the two extensions as V1 & V2 insinuates that V2 is an improvement over V1, and that both actually work...

In reality, V1 is buggy.

### DIFF
--- a/extensions/DT/cameracontrols.js
+++ b/extensions/DT/cameracontrols.js
@@ -1,4 +1,4 @@
-// Name: Camera V1
+// Name: Camera V1 (Buggy)
 // ID: DTcameracontrols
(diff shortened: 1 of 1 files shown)


GarboMuffin (approved):

Brackets-Coder (approved):

Brackets-Coder:
I remember this one being called "Buggy" previously, why did we remove it?

SharkPool-SP:
> I remember this one being called "Buggy" previously, why did we remove it?

More accidental. It was renamed to v1 because it got hidden when v2 was merged and people wanted it back

PR #2615, opened 2026-08-01
Add Speech to Text extension

MinhNhat1409212:
Hi TurboWarp team!
I would like to submit a new extension: Speech to Text

- Description: Uses the Web Speech API to convert voice input into text in real-time

- Features: Supports multiple languages (English, Vietnamese, Japanese, Korean, Chinese, etc.) and returns an empty string when no result is available

- Note: Requires unsandboxed mode to access browser speech recognition APIs

### DIFF
--- a/extensions/Speech to text.js
+++ b/extensions/Speech to text.js
@@ -1,0 +1,128 @@
+(function(Scratch) {
+  'use strict';
+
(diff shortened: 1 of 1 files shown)


GarboMuffin:
Sensing+ has this already, and the underlying browser API is sufficiently unreliable and unusable that we don't think it makes sense to have a dedicated extension for it (it would make it seem more reliable than it is )

PR #2635, opened 2026-08-16
New blocks in More Motion

tom35011:
add blocks to control other sprites to More Motion extensions


R3d5t0n3GUY:
Commenting `!format` (in this discussion, at least) should fix the formatting issues preventing your PR from passing the required checks.

GarboMuffin:
what real world scenario do you expect someone to use these blocks in and how much effort does it save compared to doing the same thing with normal blocks?

tom35011:
> what real world scenario do you expect someone to use these blocks in and how much effort does it save compared to doing the same thing with normal blocks?

I think, personally that sometimes, it's faster to do it in one sprite and it can be helpful for clarity sometimes, for example, I once used those for a one-script setup instead of one script for each sprite.

tom35011:
@GarboMuffin so is it going to be added or not?

PR #2640, opened 2026-08-20
Add BoolVariables extension

utakeuchigames:
This extension adds the BoolVariables extension, allowing users to create, manage, and use boolean-specific variables smoothly.

AI was used to assist with coding, English translation, and structure. The implementation, debugging, and testing were reviewed and performed by me.


penta-quark-neutro:
Looking everything over and testing out this extension...
What is the actual point of this extension?
It can be done perfectly well with standard blocks, and they'll run faster.

utakeuchigames:
The purpose is to simplify the storage and management of true and false values in TurboWarp, so users don't have to go through the trouble of writing <(variable) = (true)> or <(variable) = (1)> every time.

(Because it is translated from Japanese, the grammar may be wrong.)

utakeuchigames:
Do I need a comment for debugging?

I will add it if necessary.

penta-quark-neutro:
I recommend simplifying the code and using `extensionStorage`.
I think the code is overly complicated in some parts compared to what is actually necessary.

GarboMuffin:
What problem are you solving?

utakeuchigames:
> What problem are you solving?

The main purpose of this extension is to simplify the management of true/false values, so users don't have to repeatedly write <[variable] = (true)> or manage numeric flags. This streamlines condition checking and improves the overall development experience in TurboWarp.
Additionally, this extension provides features that standard blocks lack, such as built-in logical operators and event hat blocks that trigger when a boolean variable changes.
(I am using translation, so I apologize if my English is a bit unnatural.)

utakeuchigames:
There was something urgent that had to be fixed, so this PR will be in a draft state for a while.

SharkPool-SP:
Closing for these reasons:
image

These solutions are technically less tedious than making a new bool variable... or just using a `< (x) = (y) >` block

PR #2660, opened 2026-09-04
Add Text Input

HarmonyDock:
A draggable text input box that appears per sprite. When shown, it replaces the sprite's display and allows direct typing. Supports customizable width, height, font size, border color, and background color. Can be used to get user input without needing the ask/answer blocks.


HarmonyDock:
Add Text Input extention

R3d5t0n3GUY:
This is really neat! One thing I would like added is the ability for multi-line input (maybe via Shift+Enter) because
image
might get a little irritating after a while. Maybe also the ability to check if the textbox is focused? (i.e. selected and waiting for keystrokes)
scratchblocks

HarmonyDock:
@SharkPool-SP
I’ve searched through the existing extensions in the /extensions folder and all open PRs, and I couldn’t find any extension with similar functionality to mine.

Could you please clarify what this PR duplicates with? If it’s about code style or internal API usage, I’m happy to refactor. But if it’s a false positive, I’d appreciate it if the duplicate label could be removed.

For reference, my extension primarily implements put textinput box on the stage, which I believe is not currently available in TurboWarp.

Thanks!

SharkPool-SP:
Popup Pheonix and modals

